PR #59, opened 2014-08-02
Fix running without -r option

cbagwell:
If you run without any command line options, gcovr crashes without
this change since options.root never gets defined.  Using
root_dir in this case.


whart222:
This issue is now fixed in the master branch.

PR #65, opened 2014-09-07
If given, options.output is relative to start directory.

rickardlindberg:
If we enter a relative path with --output, we want it to be relative to
the directory we stand in when we run gcovr.

Because we call os.chdir in the code, the easiest way to guarantee this
is to make the path absolute before we start calling os.chdir.

### DIFF
--- a/gcovr/args.py
+++ b/gcovr/args.py
@@ -9,8 +9,9 @@
 #  For more information, see the README.md file.
 #  _________________________________________________________________________
 
-from os import environ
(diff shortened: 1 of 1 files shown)


latk:
Thank you for this PR. However, gcovr's "dev" branch is no longer under development. This issue has been fixed on the master branch in commit 22d215b7c5b5ed619544c2b25c2e1a49291767bd, and was first released as part of gcovr 3.3.

Issue #64, opened 2014-09-06
gcovr does not work with CMake out-of-source builds


sakra:
gcovr is not able to locate source files when doing an out-of-source build with cmake whose build directory is not in the source tree, i.e.:

```
mkdir /home/user/build
cd /home/user/build
cmake /home/user/source/project
make
ctest
```

This will generate the *.gcno and *gcda in target-specific subdirectories of the directory `/home/user/build/CMakeFiles`. Running 

```
gcovr -v -r /home/user/source/project
```

does not locate any coverage data files.

```
gcovr -v --object-directory=/home/user/build/CMakeFiles -r /home/user/source/project
```

does not locate any coverage data files, either.

nickbroon:
Possibly related to #61

latk:
Does this work when you invoke gcovr from within your build directory? Does using the `--verbose` option produce interesting output, e.g. explaining why any files were filtered out? Gcovr should support this use case. But without clear steps to reproduce, I cannot determine the problem.

I am closing this issue for now. If the problem persists, please add further details so it can be re-opened. If someone else is experiencing a similar issue, please create a new issue.

PR #72, opened 2014-11-11
Use --object-directory to scan for gcov data files

kblaschke:
Fixes an issue with gcovr being run on projects with out-of-source
builds, e.g. CMake-based builds. If --object-directory is given, this
path will be scanned in addition to --root to find gcda/gcno files.

Possible fix for issue #64

### DIFF
--- a/scripts/gcovr
+++ b/scripts/gcovr
@@ -2163,12 +2163,14 @@ else:
 #
 if len(args) == 1:
     if options.root is None:
-        datafiles = get_datafiles(["."], options)
+        search_paths = ["."]
     else:
-        if options.objdir is not None:
-            datafiles = get_datafiles([options.objdir, options.root], options)
-        else:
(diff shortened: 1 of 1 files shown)


DaMouse404:
Any feedback on why this isn't in yet? I've also come across problems and have patched my version to do this.

theold190:
What if --root is not specified, then the --object-directory is still ignored for searching for data files.

kblaschke:
@theold190 valid point, I amended my changes to implement your proposal. Sorry for the delay.

@DaMouse404 The last commit to the upstream repository has been made over a year ago, so it seems that the maintainer is no longer paying attention to the project.

whart222:
This commit merged these changes and resolved associated conflicts.  However, I'm going to add a ticket, since we don't have tests to verify this new behavior.

theold190:
Nice!

PR #73, opened 2014-11-20
Use os.sep instead of '/' to fix gcovr on Windows

toelke:


kankri:
I didn't notice your fix when I made a similar change on the `dev` branch in December. In my case I need to be able to handle absolute paths with the drive letter and colon. Windows supports both `\` and `/` as path separators and in my flow I get paths with both of those so I need to support that as well.

I just made a pull request of my changes in #79.

whart222:
This is resolved with commit b30aa30

Issue #74, opened 2014-11-20
Display branches in HTML


mikkleini:
LCov displays taken, not-taken and unreachable branches in HTML report with [-], [+] and [#].
Could GCovr also be updated with something like that? This data is already inside it and ends up in XML report, so maybe add it HTML also?

PR #75, opened 2014-11-24
Added branches taken/not taken info into HTML.

mikkleini:
Solves issue #74 
Now there's a branch column between line and execution count. Green ticks show taken branch, red crosses show not taken branch.


whart222:
Fixes #74

PR #79, opened 2015-01-15
Handle Windows paths correctly when generating HTML file names

kankri:
Currently `gcovr` translates slash to underscore when it generates HTML file names from `.gcov` path names.

Windows path names can contain slash, backslash and colon, which are not allowed as characters in the file name part. Thus all of these 3 characters should be mapped to underscores to make sure the generated file names are valid.


latk:
Unfortunately, the "dev" branch has not seen active development in a while. I'm therefore closing this PR.

But the PR is very good. If you want to re-submit the same change relative to the master branch, I would be delighted! The corresponding code on master is a bit more complicated, but currently has the same problem.

PR #87, opened 2015-02-19
Added ability to exclude directories from the search path of gcov files....

trapzero:
...  This helps alleviate cases where a symlink points to a higher level directory and searches take exceptionally long (162 minutes, vs 5 seconds w/ the exception).


whart222:
Although I've merged this PR, I'd like to see a test case exercise this logic. It would be nice to illustrate the performance that you reference in the PR.

PR #89, opened 2015-03-28
Fix XML report generation with python 3

eunikolsky:
This patch replaces `xrange` function (removed in python 3) with `range`
in XML report generation (when using the `--xml-pretty` option).


whart222:
This is really a broader issue related to Python 3 support.  I've separately committed a py3k fix (see commit 5f78efacec262370cabc3b336ad7a86e9dba4c81).

PR #103, opened 2015-06-08
fix common prefix detection for files

bbreinbauer:
if all files names start with the same string this string
will be removed by commonprefix. Use dirs as input for
commonprefix.


whart222:
I've merged this PR, but I would like to see a test added that demonstrates this functionality.

PR #106, opened 2015-07-21
Now handles source files that are not next to the object files better

attie:
We have the following directory structure for our projects:

    ./${SOURCE_FILES}
    ./obj/${OBJECT_FILES}
    ./test/${TEST_HARNESS_SOURCE_FILES}
    ./test/obj/${TEST_HARNESS_OBJECT_FILES}

Previously, our hierarchy was not playing nicely with `gcovr`. Typically, it would look for `${TEST_HARNESS_SOURCE_FILE}` in `./` which didn't work out too well for obvious reasons.

After the patch, this now works correctly - it will search the directory and sub directories for `${TEST_HARNESS_SOURCE_FILE}`.

One issue that I can foresee is where there are multiple filenames that match with what the `*.gcov` files are calling for... in this case the functionality is not improved. I am considering implementing a 'best match' approach, though have no particular reason to at the moment. I wonder if the `*.gcno`, `*.gcga` or `*.gcov` files have any further information that we could use (e.g: source file timestamp, total line count, etc...) to perform better matching where multiple files with a given name exist.

Attie

### DIFF
--- a/scripts/gcovr
+++ b/scripts/gcovr
@@ -71,6 +71,13 @@ c_style_comment_pattern = re.compile('/\*.*?\*/')
 cpp_style_comment_pattern = re.compile('//.*?$')
 
 
+def find_file(path, name):
+    result = []
+    for root, dirs, files in os.walk(path):
+        if name in files:
+            result.append(os.path.join(root, name))
+    return result
+
 def version_str():
     ans = __version__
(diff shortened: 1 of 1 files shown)


attie:
Placed this pull request on it's own branch... I want to do more fixups...

PR #109, opened 2015-07-23
llvm-cov: Check that code exists on lines before marking as uncovered

thatguystone:
llvm-cov sometimes outputs that non-code lines are uncovered (and
which lines it reports changes from version to version, and even
from -std=gnu++11 to -std=gnu++14). While llvm-cov should obviously
fix its own bug, this is a far quicker fix.


whart222:
Added issue #134, since we should be testing coverage with llvm

PR #115, opened 2015-09-09
replaced option parser with more elegant plac (improves feedback for …

krichter722:
…missing commands, help message and minimizes maintainance efforts); removed usage of options dictionary because exposing function arguments provides more API stability and improves code quality; added incomplete (for installation, distribution and pypi upload) setup.py to facilitate the installation of plac

PR #116, opened 2015-09-09
added coverage_min argument 

krichter722:
which makes the process return with a value != 0 if a minimal coverage isn't reached for all files

based on #115


whart222:
Since this is based on #115, I'm deferring this.

Issue #17, opened 2013-09-19
Using gcovr html details on windows


mretsh91:
When using the command  (gcovr --html --html-details) on cmd windows it don't work.
I think because of the paths that are interpreted in a way that windows doesn't understand it. The error is:
IOError: [Errno 2] No such file or directory: 'E:user\PGA_Source_Code\PGA_prg.c'

I think the "//" is not interpreted correctly in windows.

Please I need help with this urgently

mrx23dot:
Please use the following for all inputs:

from os.path import normpath
abs_filename = os.path.abspath(normpath(filename))

This will unify all paths for given OS.

itavero:
Has anyone submitted a PR for this?

latk:
This may have been fixed by pull request #157.

If the problem persists, please add further details so it can be re-opened. If someone else is experiencing a similar issue, please create a new issue.

PR #122, opened 2016-01-24
Bugfix issue #17: Using gcovr html details on windows

goriy:
gcovr could not create detailed output html files on Windows OS
due to characters '\' and ':' in relative and absolute source file paths.

These characters should be replaced to underscores like '/' chars
was replaced for unix-like systems.

OS-independent way to replace '\' on Windows is to use
os.sep instead of hard-coded '/'.

But it's still necessary to replace also a colon characters.
It's done by second stage replace.

Similar (but not full - without colons replacement) patch was
introduced on master branch by kreuzberger at 19:09:05 10/09/2014
It's commit hash was 6a4b40900c0506e6fb97b91abe5979ac25fd2d3e.

### DIFF
--- a/gcovr/prints/html.py
+++ b/gcovr/prints/html.py
@@ -540,11 +540,12 @@ def print_html_report(covdata, options):
         if len(ttmp) > 1:
             cdata._sourcefile = \
                 '.'.join(ttmp[:-1]) + \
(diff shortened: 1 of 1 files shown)


goriy:
Similar ones #79 and #73 are already there.

PR #124, opened 2016-01-26
allow users (esp. Clang users) to supply args to the gcov command

jefftrull:
Analyzing coverage reports generated by Clang/LLVM requires the use of a substitute for `gcov`.  gcovr supports doing this via an environment variable **GCOV** or with a command line option.  Unfortunately Clang users must run not just a single command, but a command with an option: `llvm-cov gcov`.  This results in an error when gcovr tries to run the gcov subprocess.

This PR addresses the problem by interpreting gcov commands with embedded spaces as a command with supplied arguments and prefixes its own set of arguments with those supplied by the user.  The result is that Clang users can run gcovr with something like:

```
GCOV='llvm-cov gcov' gcovr --root=. other-arguments...
```
and get good results.

### DIFF
--- a/scripts/gcovr
+++ b/scripts/gcovr
@@ -805,6 +805,12 @@ def process_datafile(filename, covdata, options):
             sys.stdout.write(
                 "Running gcov: '%s' in '%s'\n" % (' '.join(cmd), os.getcwd())
             )
+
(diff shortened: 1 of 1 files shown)


whart222:
I added #133 to ensure that we test this sort of logic.

Issue #88, opened 2015-03-27
IOError: [Errno 63] File name too long


runv:
gcovr -r fails when processing objective c file with a name, which is longer then 63 letters:

Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/2.7/bin/gcovr", line 1961, in <module>
    process_datafile(file_,covdata,options)
  File "/Library/Frameworks/Python.framework/Versions/2.7/bin/gcovr", line 749, in process_datafile
    process_gcov_data(fname, covdata, options)
  File "/Library/Frameworks/Python.framework/Versions/2.7/bin/gcovr", line 412, in process_gcov_data
    INPUT = open(data_fname,"r")
IOError: [Errno 63] File name too long: u'#var#lib#slave#workspace#com.abcde.app.aaaa#modules#client#abc-abc-abc-abc#src#a#items#process1#subprocess1#subprocess2#ff#ABCEfgHijklMnopQRstuvSomeClassName.m.gcov'

gcovr --version
gcovr 3.2

theold190:
Similar problem when I run my scripts. File names was 123 characters, while full path - 217.

whart222:
Can you add a test to confirm whether #128 resolves this issue?

PaulTaykalo:
The bug is still reproduceable.
It started when switched cocopods to use frameworks instead of static typing.

tchernobog:
This is probably a problem in gcov instead of gcovr. Can you tell gcovr not to remove intermediate files, and paste here the full path on disk to one of the .gcov files triggering the error?

PaulTaykalo:
IOError: [Errno 63] File name too long: u'#Users#jenkins-ci#jenkins-slave#workspace#xxxxxxxxx_IOS_develop-tests_publish#cintegration#output#Build#Intermediates#ArchiveIntermediates#xxxxxxxxxTests#BuildProductsPath#Coverage-iphonesimulator#CedarAsync#CedarAsync.framework#Headers#AsyncActualValue.h.gcov'

PaulTaykalo:
Hm.. and for some reason files like that were actually created inside of the root of the proejct

PaulTaykalo:
This is kind'a fixes it...
```
def process_gcov_data(data_fname, covdata, source_fname, options):
    try:
        INPUT = open(data_fname, "r")
    except Exception as e:
        sys.stdout.write("  Cannot open gcov file %s\n" % e.strerror)
        return
```

[...]

PR #128, opened 2016-06-22
Shorten HTML output filename if exceeds fs limits

tchernobog:
Should partially fix: https://github.com/gcovr/gcovr/issues/88

Note that the original issue speaks of the length of gcov file names,
rather than the HTML or XML output produced by gcovr. As such, the
issue with *.gcov filenames should probably be resolved outside of
gcovr.


whart222:
I merged this commit a3b7c9f , though somehow I screwed up the merge information.

PR #135, opened 2016-08-05
Fix nested makefiles

whart222:


whart222:
This merge is complete.

PR #142, opened 2016-08-29
ensure that path is not using / on windows

bbreinbauer:
I hit a problem that the file path for html report would contain "/" characters and windows doesn't like that. So gcovr would bail out with an IO Error. Normalizing the path ensures that all "/" are converted to "\" on windows.

Add: (I don't know from where the slashes are coming from, maybe mingw?)

### DIFF
--- a/scripts/gcovr
+++ b/scripts/gcovr
@@ -1577,7 +1577,7 @@ def print_html_report(covdata, details):
     )
     for f in keys:
         cdata = covdata[f]
-        filtered_fname = options.root_filter.sub('', f)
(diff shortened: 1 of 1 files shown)


latk:
Thank you for suggesting this fix. Can you verify whether it is still needed with gcovr 3.4? The gcovr test suites seems to run fine on Windows without this fix. Does the test suite perhaps have to be extended?

latk:
This PR has not seen any progress in a while, so I am closing it for now. If the problem persists, please add further details so it can be re-opened.

I think HTML output on Windows has been fixed in the meanwhile, at least our tests are passing.

PR #144, opened 2016-09-01
Fix verbose output when using existing gcov files

### LINKED ISSUE #143: --verbose & --use-gcov-files can't be used at the same time
Fails with:

```
    ....
    print('  source_fname '+source_fname)
TypeError: cannot concatenate 'str' and 'NoneType' objects
```

Tested with gcovr@16c81ccbdbcc43844b991c68a6941ddfcbd9a6a0

mayeut:
When gcovr is invoked with `-v` and `-g` switches, it fails with error:

```
    ....
    print('  source_fname '+source_fname)
TypeError: cannot concatenate 'str' and 'NoneType' objects
```

This patch disables printing of `source_fname` when `source_fname is None`

Fixes #143


gcovr/tests/use-existing/main.cpp
```diff
@@ -0,0 +1,26 @@
...
+     return 0; //std::cout << "param is null." << std::endl;
+  }
+}
+
+
+int main(int argc, char* argv[]) {
+  foo(
+     0
+     )
+     ;
```
latk: This fancy code layout is not relevant to the gcovr behaviour being tested: the `-v` option.

gcovr/tests/use-existing/README
```diff
@@ -0,0 +1,5 @@
+A test that verifies coverage when using existing *.gcov coverage files with verbose output.
+
+The second conditional in main() includes lines with 
+(, ) and ; characters.  These need to be treated as 
+covered lines.
```
latk: This paragraph is misleading, because it is unrelated to the actual test. And the conditional doesn't contain any commas.

scripts/gcovr
```diff
@@ -526,7 +526,8 @@ def process_gcov_data(data_fname, covdata, source_fname, options):
         print('  currdir      '+currdir)
         print('  gcov_fname   '+data_fname)
         print('               '+str(segments))
-        print('  source_fname '+source_fname)
+        if not source_fname is None:
+            print('  source_fname '+source_fname)
```
latk: PEP-8 [suggests](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) to write the conditional like `if source_fname is not None`. Let's use that for consistency.

Also, do you think `print('  source_fname '+(source_fname if source_fname is not None else '(None)'))` might be better? This makes it easier to see that the output is aligned, and makes it clearer that this field is explicitly empty. Or if we simply concatenate `str(source_fname)` rather than `source_fname`, we don't even need any conditionals.

gcovr/tests/use-existing/README
mayeut: README file updated to reflect actual content of this folder

gcovr/tests/use-existing/main.cpp
mayeut: c.f. updated README file => same cpp file as simple1 test

scripts/gcovr
mayeut: I went with `str(source_fname)` as suggested.

PR #156, opened 2016-11-22
Fix unicode exceptions

### LINKED ISSUE #148: gcovr crashes with files which are not utf-8 encoded
Unfortunately we have some submodules which are not utf-8 encoded.
If we run gcovr on our project, we got a backtrace:

```
$ gcovr -r .
  File "/usr/lib/python-exec/python3.4/gcovr", line 2312, in <module>
    process_datafile(file_, covdata, options)
  File "/usr/lib/python-exec/python3.4/gcovr", line 891, in process_datafile
    process_gcov_data(fname, covdata, abs_filename, options)
  File "/usr/lib/python-exec/python3.4/gcovr", line 489, in process_gcov_data
    line = INPUT.readline()
  File "/usr/lib64/python3.4/codecs.py", line 319, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xfc in position 1750: invalid start byte
```

In projects where all files are utf-8 encoded everything works fine.

We are using gcovr-3.3.

jkloetzke:
Source files may not be properly encoded. Make the handling of such
files more tolerant.

Fixes #148.

### DIFF
--- a/scripts/gcovr
+++ b/scripts/gcovr
@@ -88,6 +88,13 @@ def version_str():
     return ans
 
 
+# fake open() encoding errors handling on Python 2
+if sys.version_info[0] == 2:
+    __open = open
+    def open(file, mode='r', buffering=-1, errors=None):
+        return __open(file, mode, buffering)
+
+
 #
 # Container object for coverage statistics
 #
@@ -482,7 +489,7 @@ def is_non_code(code):
 # Process a single gcov datafile
 #
 def process_gcov_data(data_fname, covdata, source_fname, options):
-    INPUT = open(data_fname, "r")
+    INPUT = open(data_fname, "r", errors="surrogateescape")
     #
     # Get the filename
     #
@@ -1696,7 +1703,7 @@ def print_html_report(covdata, details):
         data['ROWS'] = []
         currdir = os.getcwd()
         os.chdir(root_dir)
-        INPUT = open(data['FILENAME'], 'r')
+        INPUT = open(data['FILENAME'], 'r', errors="surrogateescape")
         ctr = 1
         for line in INPUT:
             data['ROWS'].append(


strahlc:
can you please merge this PR?

strahlc:
please merge

latk:
Thank you for these changes. I think that overriding the open() function is an elegant way to handle the Python 2 vs 3 differences.

But this solution only hides any errors and doesn't actually support other encodings. Resilience against encoding errors is good, but I am deferring this PR until a general strategy for dealing with source encodings can be developed. See also my comment at https://github.com/gcovr/gcovr/issues/148#issuecomment-364786620 for more context.

jkloetzke:
I think they are two parts that can be considered almost independently. The one thing is to override the default source encoding (inferred by Python from the locale) and the other is how to treat errors. This PR is solely about to handle encoding errors. Note that even if it would be possible to specify a source encoding it would be not sufficient for our use case. We have quite a bit of legacy and 3rd party code that are all compiled together. Because of that there are unfortunately multiple different encodings in the code base.

This PR proposes to use the `surrogateescape` error handler. Quoting the Python documentation:

> 'surrogateescape' will represent any incorrect bytes as code points in the Unicode Private Use Area ranging from U+DC80 to U+DCFF. These private code points will then be turned back into the same bytes when the surrogateescape error handler is used when writing data. This is useful for processing files in an unknown encoding.

I think this is a good trade-off. This way gcovr can read incorrectly encoded source files, process the correctly encoded parts and write back the result with the same byte representation. The other options are basically only to either replace the offending bytes or to bail out. IMHO neither of them look attractive.

So I think it is still valuable to have this kind of error handling even if the source encoding can be properly specified.

latk:
An alternative approach has been implemented in #256, so I'm closing this PR. Still, many thanks for putting this prototype forward as it helped the discussion! If you feel that the currently implemented solution does not support a specific use case, please open a new issue.

PR #157, opened 2016-11-23
Fix gcov '/'s breaking --html-details on Windows

jessicalevine:
The gcov reports (from which gcovr output is generated) use '/' as the
path separator in filenames, regardless of the path separator used by
the operating system. This breaks attempts to read files located in
subdirectories while generating --html-details reports on Windows.

This commit fixes the issue by replacing '/'s present in the gcov
filename with Python's OS-sensitive os.sep character.


Flockstud:
This works well, I had to manually add your changes in, but works perfectly! This needs to  be merged to the master!

PR #158, opened 2016-11-23
Fix exclude filters on Windows

jessicalevine:
When building exclusion filters from CLI input on Windows, the backslash
path separators from os.path.realpath aren't properly escaped before
being built into a regex.

This refactors the process of building the filter into a function, and
includes alternate logic for constructing the filter on Windows in order
to escape the path separators without escaping the CLI input.

### DIFF
--- a/scripts/gcovr
+++ b/scripts/gcovr
@@ -2287,23 +2287,33 @@ else:
 #
 # Setup filters
 #
+def build_filter(regex):
+    if os.name == 'nt':
+        # Windows path separators must be escaped before being parsed into a
+        # regex, but we do not want to escape the regex itself, so instead of
+        # using realpath, we escape the path and join it manually (realpath
+        # doesn't resolve symlinks on Windows anyway)
+        return re.compile(re.escape(os.getcwd() + "\\") + regex)
+    else:
+        return re.compile(os.path.realpath(regex))
+
 for i in range(0, len(options.exclude)):
-    options.exclude[i] = re.compile(os.path.realpath(options.exclude[i]))
+    options.exclude[i] = build_filter(options.exclude[i])
 
 if options.exclude_dirs is not None:
     for i in range(0, len(options.exclude_dirs)):
-        options.exclude_dirs[i] = re.compile(os.path.realpath(options.exclude_dirs[i]))
+        options.exclude_dirs[i] = build_filter(options.exclude_dirs[i])
 
 options.root_filter = re.compile(re.escape(root_dir + os.sep))
 for i in range(0, len(options.filter)):
-    options.filter[i] = re.compile(os.path.realpath(options.filter[i]))
+    options.filter[i] = build_filter(options.filter[i])
 if len(options.filter) == 0:
     options.filter.append(options.root_filter)
 
 for i in range(0, len(options.gcov_exclude)):
-    options.gcov_exclude[i] = re.compile(os.path.realpath(options.gcov_exclude[i]))
+    options.gcov_exclude[i] = build_filter(options.gcov_exclude[i])
 if options.gcov_filter is not None:
-    options.gcov_filter = re.compile(os.path.realpath(options.gcov_filter))
+    options.gcov_filter = build_filter(options.gcov_filter)
 else:
     options.gcov_filter = re.compile('')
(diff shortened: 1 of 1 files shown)


scripts/gcovr
```diff
@@ -2287,23 +2287,33 @@ else:
 #
 # Setup filters
 #
+def build_filter(regex):
+    if os.name == 'nt':
+        # Windows path separators must be escaped before being parsed into a
+        # regex, but we do not want to escape the regex itself, so instead of
+        # using realpath, we escape the path and join it manually (realpath
+        # doesn't resolve symlinks on Windows anyway)
+        return re.compile(re.escape(os.getcwd() + "\\") + regex)
```
latk: This looks like a reasonable solution to fix the Windows filter problem, but do we really need a separate code path just for Windows? Wouldn't something like this also work?:

``` python
return re.compile(re.escape(os.getcwd() + os.path.sep) + regex)
```

Or do you see any drawbacks with that approach, e.g. around symlinks?

goriy:
I have been using gcovr with this patch for several months on Windows (python 3.5). It works fine!

After upgrading python to version 3.6.4, gcovr began to crash in one of my scripts.

```
  File "C:\Python36\Scripts\gcovr", line 2338, in <module>
    options.exclude[i] = build_filter(options.exclude[i])
  File "C:\Python36\Scripts\gcovr", line 2333, in build_filter
    return re.compile(re.escape(os.getcwd() + "\\") + regex)
  File "C:\Python36\lib\re.py", line 233, in compile
    return _compile(pattern, flags)
  File "C:\Python36\lib\re.py", line 301, in _compile
    p = sre_compile.compile(pattern, flags)
  File "C:\Python36\lib\sre_compile.py", line 562, in compile
    p = sre_parse.parse(p, flags)
  File "C:\Python36\lib\sre_parse.py", line 855, in parse
    p = _parse_sub(source, pattern, flags & SRE_FLAG_VERBOSE, 0)
  File "C:\Python36\lib\sre_parse.py", line 416, in _parse_sub
    not nested and not items))
  File "C:\Python36\lib\sre_parse.py", line 502, in _parse
    code = _escape(source, this, state)
  File "C:\Python36\lib\sre_parse.py", line 401, in _escape
    raise source.error("bad escape %s" % escape, len(escape))
sre_constants.error: bad escape \p at position 35
```

When I changed builtin package `re` to third-party `regexp` (`import regex as re` instead of `import re`) - everything began to work again. There are many such advices on different sites, but I don't like this solution in this particular case.

One of gcovr command line parameters in my script looks like `-e ".*\ports\.*"`.
I changed it to `-e ".*\\ports\\.*"` and error was gone! 
So I think it's correct to escape regexps on command line when necessary (problem is in my script, not in patched gcovr), but it looks strange that it worked fine with python's versions 3.5.x.

latk:
@goriy Thank you for testing this.

Note that the `-e` argument is a regex, so if you want a literal backslash you must escape it in the regex.

The [Python 3.6 release notes](https://docs.python.org/3/whatsnew/3.6.html?highlight=new#api-and-feature-removals) state:

> Unknown escapes consisting of `\` and an ASCII letter in regular expressions will now cause an error.

So that's why the `\p` escape stopped working. This check is within the `re` module. The `regex` module does not warn on unknown escapes.

goriy:
@latk Thank you for clarification.

scripts/gcovr
barthoukes: At line 505 in gcovr, "Return if the filename does not match the filter" it will always return in Windows. For me this is a serious issue... by removing 10 lines of code the gcovr tool is really working great for me!!!! wasted 2 days on this grgrrgr...

latk: @barthoukes Thank you for investigating this, and I'm sorry that you found gcovr difficult to use. But please note that this discussion should be limited to the changes _in this pull request_. 

You seem to have found a related but different issue. Please consider opening a new issue for your problem (see our [contributing guide](https://github.com/gcovr/gcovr/blob/master/CONTRIBUTING.rst)). If this is a solution to [your Stack Overflow question](https://stackoverflow.com/q/49506856), consider answering there.

Issue #112, opened 2015-08-20
gcovr doesn't create detailed report when source file are located in other directories


milexy86:
Hi,

if for example my source files are located within source/myfile.c and the gcda and gcno files also within obj/ , than gcovr creates me the detailed html report  -  everything OK. 

Problem is when I have to take the source files for example from:
../../../Project/Common/Source/myfile.c , than the main html report with the summary is created, but not the detailed one of my source file "myfile.c" . 

The gcovr command is:

C:\Python34\Scripts\gcovr -v -b -r C:\Project\UnitTests\myModule\module1 --object-directory      C:\Project\UTests\myModule\module1\test-obj\Common\Source --html --html-details -o report\report.html

For better understanding, I have attached a screenshot. You see, the main html is generated, butt not the one for "myfile.c" (When i click on myfile.c to see the detailed report, I'm not relinked to it, because it doesn't exist). 

Any idea?
![gcovr_result](https://cloud.githubusercontent.com/assets/13888058/9385879/130476f4-4759-11e5-8802-d00690fe4111.PNG)

AndersHoglund:
I have a similar problem, and I have found some workarounds that might help.
1. Use the latest gcovr version on the dev branch.
2. Replace all relative parallell paths in buildscripts/makefiles with symlinks in the filesystem. 
3. Run gcovr twice. First time with "-k" option, second time with "-k -g" options. I also add "-x" for xml output in the last stage.

For example in my Makefile I do this:

    # Workaround limitations in gcovr. .
    # Replace relative path with a symlink "main -> ../main"
    #USER_DIR := ../main
    USER_DIR := main

And the Jenkins job does this:

    /home/andho/projects/coverage/gcovr/scripts/gcovr -r obj/test/ -k
    /home/andho/projects/coverage/gcovr/scripts/gcovr -r obj/test/ -k -g -x > coverage.xml

Hope it is some help....

AndersHoglund:
Found another way way to run gcovr, without symlinks and just run once.
You can supply your coverage data location as a pure arg (no option) and point the root to your sources.

     /home/andho/projects/coverage/gcovr/scripts/gcovr obj/test --root=src/main -x > coverage.xml

Found this out by reading the sources, supplying an arg like that is not documented in the --help text.

nickbroon:
I think this is perhaps a duplicate of #61

nickbroon:
[...]

Issue #61, opened 2014-08-14
Support out of source tree builds


nickbroon:
This does not work for me:

```
cd /tmp/builds/project1;
/bin/python /home/user1/gcovr/scripts/gcovr -v -x -r /home/user1/project1 --object directory=/tmp/builds/project1 -o /home/user1/coverage.xml
```

I get the following output:

```
Scanning directory /home/user1/project1 for gcda/gcno files...
Found 0 files (and will process 0)
Gathered coveraged data for 0 files
```

I expected `gcovr` to use the path given by the `--object-directory` option to look for gcda/gcna files, while using the `-r` option to specify where the source files are, but it appears just look in the path given by `-r`.

lemonkey:
I too can confirm this is happening with the latest (v3.2).  If I copy my .gcno files into the same root folder as the source files defined by the -r argument, it generates the report just fine.  But if I define the object-directory path to be somewhere else, the report processes 0 files.

briangreenery:
I'm not sure it'll work for you, but this is the solution I found: https://github.com/gcovr/gcovr/issues/31#issuecomment-48809388

Basically, inside the build directory run:

    gcovr --root=. --filter=.*

Maybe https://github.com/gcovr/gcovr/pull/72 fixes this?

wangshan:
it's not very clear what --object-directory does, but it may work if you remove it, just specify the directory where the .gcda, .o and .gcno files are, like this: /bin/python /home/user1/gcovr/scripts/gcovr -v -x -r /home/user1/project1 /tmp/builds/project1 -o /home/user1/coverage.xml

nickbroon:
I wonder if #108 relates to this?

latk:
This issue has not seen any progress in a while, so I am closing it for now. If the problem persists, please add further details so it can be re-opened. If someone else is experiencing a similar issue, please create a new issue.

I suppose this was fixed in 3.3 or 3.4. The `--object-directory` is a bit special and overrides normal path detection. In most cases specifying the build directory as an (unnamed) argument is sufficient.

PR #159, opened 2016-11-30
Out of Source build that reflects proper use case

nickbroon:
The following change to the oos testcase, such that I believe it more
closely reflects how tools like autoconf/automake/cmake perform out of
source builds (what they call vpath builds)

It currently fails, showing that gcovr has issues with oos paths.
See github issues #61 #64 #72 #112

### DIFF
--- a/gcovr/tests/oos/Makefile
+++ b/gcovr/tests/oos/Makefile
@@ -1,10 +1,10 @@
 CFLAGS= -fprofile-arcs -ftest-coverage -fPIC
 
 all:
-	mkdir -p build
-	$(CXX) $(CFLAGS) -c src/file1.cpp -o build/file1.o
-	$(CXX) $(CFLAGS) -c src/main.cpp -o build/main.o
-	$(CXX) $(CFLAGS) build/main.o build/file1.o -o build/testcase
+	mkdir -p build/src
+	cd build; $(CXX) $(CFLAGS) -c ../src/file1.cpp -o src/file1.o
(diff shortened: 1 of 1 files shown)


nickbroon:
The failure is expected.
The test case was updated to reflect how out of source tree builds really work.
gcovr needs fixed to address this.

latk:
We can add this testcase – but not if it fails.

The test case failures are *not expected*. Gcovr should be able to handle this style of compiler invocation. The important part is: **gcovr should be invoked from the same directory where you invoke the compiler**. You have not updated the gcovr invocation to account for the changed compiler invocation. What's missing is a change like this:

``` diff
-	../../../scripts/gcovr $(GCOVR_TEST_OPTIONS) -d -o coverage.txt
+	cd build; ../../../../scripts/gcovr $(GCOVR_TEST_OPTIONS) -r .. -d -o ../coverage.txt
```

and so on for all test variants (txt, xml, html).

Afterwards, you only need to update the directories in the reference files from `.` to `..` to account for the different `--root`.

Also, please don't destroy the existing testcase. Instead, add a separate testcase (e.g. `oos3`) for your improvements.

If you update this PR accordingly, I'll be happy to merge. Please also explain if you don't think my suggested gcovr invocation reflects a realistic use case.

latk:
This PR has not seen any progress in a while, so I am closing it for now. If the above points are addressed (e.g. by fixing this PR, or by describing how gcovr should implement this use case), I'll be happy to re-open.

PR #163, opened 2016-12-29
re-enable HTML reports under Python3

latk:
In commit 1cc987a23, the `commonpath()` function was added, which quite sensibly tries to figure out real path names. Unfortunately, it uses `iterools.izip()` which is not available under Python3. Instead, that
functionality is now provided by the `zip()` builtin.

We therefore define an `izip()` function as an alias for `itertools.izip()` if that is available, or else use `zip()`.

Since `commonpath()` is used in HTML reports, it was previously impossible to create HTML reports with GCOVR under Python3. To prevent such problems in the future, the existing HTML tests should
be updated and enabled. (I've already done that in https://github.com/latk/gcovr/commit/15c10ca5171f5968c8bebda8143ca03921cc2368, but the tests will not succeed under Python3 until this pull request is merged.)

### DIFF
--- a/scripts/gcovr
+++ b/scripts/gcovr
@@ -55,6 +55,12 @@ try:
 except NameError:
     xrange = range
 
+# In Python3, zip is equivalent to izip
+try:
+    izip = itertools.izip
+except AttributeError:
+    izip = zip
+
 medium_coverage = 75.0
 high_coverage = 90.0
 low_color = "LightPink"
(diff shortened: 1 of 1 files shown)


mayeut:
@latk, if you already have tests that were showing the error, would be good to have them in the PR as well.

latk:
@mayeut The current HTML tests already trigger the error, but they were commented out of the default test suite. These tests are therefore out of date and would fail under all Python versions.

I don't want to propose changes that cause the build to fail. I will issue the PR for the updated tests once this PR or the equivalent #168 is merged.

latk:
Duplicate of #168

PR #173, opened 2017-07-22
Add --fail-under-line and --fail-under-branch options to return exit status 2/4/6 under a coverage threshold

mayeut:
This might be useful to fail tests if coverage is not high enough.

### DIFF
--- a/gcovr/tests/simple1/Makefile
+++ b/gcovr/tests/simple1/Makefile
@@ -5,6 +5,10 @@ run: txt xml html
 
 txt:
 	./testcase
+	../../../scripts/gcovr --fail-under-line 80.1; test $$? -eq 2
+	../../../scripts/gcovr --fail-under-branch 50.1; test $$? -eq 4
+	../../../scripts/gcovr --fail-under-line 80.1 --fail-under-branch 50.1; test $$? -eq 6
+	../../../scripts/gcovr --fail-under-line 79.9 --fail-under-branch 49.9
 	../../../scripts/gcovr -d -o coverage.txt
 
 xml:
--- a/scripts/gcovr
+++ b/scripts/gcovr
@@ -1039,9 +1039,9 @@ def print_text_report(covdata):
 
 
 #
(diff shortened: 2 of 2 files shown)


denniswjackson:
Making this `--fail-under-lines` allows an option for branch coverage as well.

latk:
I like this enhancement very much. Only two things are missing:

1. Please split the unrelated whitespace changes into a separate commit. Such improvements are good, but not when intermingled with important changes. To do so:

     - switch to your branch: `git checkout fail-under`
    - soft-reset to the state before this commit: `git reset HEAD^`. Your workdir is now dirty with the changes from this commit.
    - add the whitespace changes interactively: `git add -p --all`. You will be asked for each change whether you want to stage it.
    - verify that all whitespace changes are staged and no other changes were staged by running `git diff` and `git diff --staged`. To restart the staging, run `git reset`.
    - commit the staged whitespace changes: `git commit`
    - commit the actual changes: `git commit --all`
    - force-push the changes to update this PR: `git push --force-with-lease`

 2. Test cases. You can copy any suitable existing test case and adapt the copy to your needs. We'll have to add three new cases:

    - one where the coverage for both criteria is exceeded. This test case needs no changes other than adding the options to the gcovr invocations in the Makefile.
    - one where the branch coverage fails. Relative to the success-case, we need to raise the coverage limit and then test the exit status after completion: `gcovr ... ; test $? = 2` or something like that.
    - one where the line coverage fails. Equivalent changes to the previous case.

Please let me know if you need any help, e.g. with running the test suite.

Related PRs:

 - your #175 (merged) fixes the Travis tests. Consider rebasing your changes onto master so that the automated build for this PR succeeds.
 - my #185 (pending) also affects the tests. I'll have to add the necessary changes for whatever PR is merged second.

mayeut:
1. addressed by rebasing on master after #187 
2. Done in simple1 txt test

Exit code now depends on which test fails:
2 => line coverage failed
4 => branch coverage failed
6 => both failed

PR #43, opened 2014-04-19
Extended html output

lu-zero:
And some refactoring to make it more usable.

PR #180, opened 2017-11-12
Add anchors to line numbers

ligurio:
This patch adds ability to make direct links to the specific lines of source code.
It'll look like this https://screenshots.firefox.com/kstmbLOfmqFixEpl/null

### DIFF
--- a/scripts/gcovr
+++ b/scripts/gcovr
@@ -1716,7 +1716,7 @@ def print_html_report(covdata, details):
 def source_row(lineno, source, cdata):
     rowstr = Template('''
     <tr>
-    <td align="right" class="lineno"><pre>${lineno}</pre></td>
+    <td align="right" class="lineno"><pre><a href="#${lineno}">${lineno}</a></pre></td>
     <td align="right" class="linebranch">${linebranch}</td>
     <td align="right" class="linecount ${covclass}"><pre>${linecount}</pre></td>
     <td align="left" class="src ${covclass}"><pre>${source}</pre></td>


latk:
This is a great improvement, but I think this should be deferred until work on other HTML improvements such as #43 is re-started. At the very least, I want the HTML tests (#185) to be in place first.

Also, your screenshot is no longer available (your image host only stores images for two weeks).

latk:
I'd like to merge this. But this requires

 1. that the change is now re-implemented on the current state of the project. The HTML format is now implemented in the `gcovr.html_generator` module.

 2. that the test cases are updated. This is admittedly rather tedious, but can perhaps be done with search & replace. How to run the tests is described in the [CONTRIBUTING](/gcovr/gcovr/tree/master/CONTRIBUTING.rst) guide.

I'm also thinking about how the anchors should be named. You suggest `#123`. I'd prefer if the anchor started with an alphabetical character, e.g. GitHub uses `#L123`. What do you think? We should be clear about this before any effort is expended on updating all those tests.

If you go through with this, could you also add your name to the AUTHORS.txt file? Thank you :)

latk:
Another thought: the line numbers don't just have to link somewhere with a `href`, we also need a link target. I.e. a `id` or `name` attribute containing that anchor on that line. This attribute should probably be on the `<a>` link itself.

latk:
This issue has not seen any progress in a while, so I am closing it for now. I'll be happy to re-open and merge when the mentioned problems are resolved, because linkable line numbers would really be useful :)

PR #185, opened 2018-01-06
Fix HTML tests

latk:
If we are going to improve the HTML reports, we need some kinds of tests first. HTML tests did exist, but were disabled and somewhat out of date. This PR updates and activates these tests.

----

The HTML tests are naturally rather fragile, since lots of CSS is copied into every file. However, they are quite valuable as they exercise important code paths. For example, this would have found issues #163 or #168. By enabling these tests, we can avoid such problems in the future.

Summary of changes:

**`scripts/gcovr`:**
removed some trailing whitespace in the HTML template.

**`gcovr/tests/test_gcovr.py`:**
normalize HTML files so that they can be compared. In particular, we remove the exact version string, and the report creation date.

**`gcovr/tests/*/reference/coverage.html`:**
update the HTML files to match the current structure. This is mostly removing trailing whitespace, and tracking changes to the CSS. If there was some discrepancy in the content, the TXT and XML references were assumed to be up to date and fresh values taken from there.

**`gcovr/tests/linked/reference/coverage.html`:**
update as with the other tests. However, this exposed significant changes to the path name handling. Commit 1cc987a23 changed `commonprefix()` to `commonpath()` to figure out better path names. This function is used only by the HTML test, which therefore reports physical paths. The HTML reference has been updated to use physical paths. E.g. since the test case is constructed on the fly from the “nested” test, the source root dir is `../nested/subdir`. Note that the XML and TXT reports still use symlinks in paths, but that is another problem.


gcovr/tests/shared_lib/reference/coverage.html
```diff
@@ -252,12 +266,12 @@
...
           <tr>
             <td class="headerName">Date:</td>
-            <td class="headerValue">2016-08-06</td>
+            <td class="headerValue">0000-00-00</td>
             <td></td>
             <td class="headerName">Lines:</td>
             <td class="headerTableEntry">8</td>
             <td class="headerTableEntry">10</td>
-            <td class="headerTableEntry" style="background-color:LightGreen">80.0 %</td>
+            <td class="headerTableEntry" style="background-color:#FFFF55">80.0 %</td>
```
wolf99: If changing to using hex values for colors, should they _all_ be changed to hex? It doesn't affect functionality of course, but its a point to consider for consistency's sake

wolf99:
LGTM

PR #186, opened 2018-01-07
Resolves 164. 

libPhipp:
Adding Cobertura's coverage-04.dtd attributes to the coverage-element and adjusting test references.

### DIFF
--- a/gcovr/tests/simple1/reference/coverage.xml
+++ b/gcovr/tests/simple1/reference/coverage.xml
@@ -1,7 +1,7 @@
-<?xml version="1.0" ?>
+<?xml version="" ?>
 <!DOCTYPE coverage
-  SYSTEM 'http://cobertura.sourceforge.net/xml/coverage-03.dtd'>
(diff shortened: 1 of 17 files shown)


gcovr/tests/simple1/reference/coverage.xml
```diff
@@ -1,7 +1,7 @@
-<?xml version="1.0" ?>
+<?xml version="" ?>
```
wolf99: To double check, is this correct? 
All other files with changes have the XML version added, this one has it _removed_...

wolf99: To double check, is this correct? 
All other files with changes have the XML version added, this one has it _removed_...

latk: @wolf99 The files are sanitized by a few simple regexes, one of which removes the gcovr version. The files are edited on-disk, so the sanitized files ought to be committed.

Unfortunately this regex is a bit overzealous and also mangles the XML declaration. The result is *technically* not a well-formed XML document.

I've addressed those issues in #188, so that this PR can move forward independently.

Issue #164, opened 2017-02-13
Add support to the latest Cobertura format


lmussier:
Hi,

Could you add the support of the latest cobertura format [coverage-04.dtd](https://github.com/cobertura/web/blob/master/htdocs/xml/coverage-04.dtd)  for xml output?

Thanks.

dishmal:
Any idea of if/when this would happen to support TFS usage?

shreyaskambl:
VSTS only accepts cobertura-04.dtd format. It would be great if we can have it implemented

libPhipp:
That would be great indeed. The only difference between Cobertura's [03.dtd](https://github.com/cobertura/web/blob/master/htdocs/xml/coverage-03.dtd) and the new [04.dtd](https://github.com/cobertura/web/blob/master/htdocs/xml/coverage-04.dtd) is in the <coverage> element. Five new required attributes were introduced:

lines-covered
lines-valid
branches-covered
branches-valid
complexity 

Does anyone know what information is expected in those attributes? Do we already have that elsewhere or may be at least the summands? Can we get the information from print_xml_report's parameter "covdata"?

latk:
This issue seems to be quite important.

The meaning of these attributes can probably be derived from the Cobertura implementation itself. Their [XMLReport class](https://github.com/cobertura/cobertura/blob/master/cobertura/src/main/java/net/sourceforge/cobertura/reporting/xml/XMLReport.java) looks like a good starting point. Since they use the same code to generate reports for all DTDs, we can do the same.

Calculating cyclomatic complexity is clearly out of scope for gcovr, but I think the other parameters can be derived from the CoverageData objects.

If someone wants to take a go at this:

 0. Run the XML tests to verify that gcovr currently works: `python gcovr/tests/test_gcovr.py GcovrXml`
 1. Add the new required attributes to the report and use the 04.dtd.
 2. Re-run the XML tests. They should now fail due to the added attributes.
 3. For each failing tests, compare the actual report with the reference XML report. If there are no discrepancies other than the new attributes, replace the reference file with the new report.
 4. The tests should now pass again.
 5. Create a pull request to the master branch and mention this issue in the description.

Perhaps we also need to add a command line flag so that the DTD version can be selected explicitly.

denniswjackson:
For those looking for complexity results, check out https://people.debian.org/~bame/pmccabe/download.html

libPhipp:
[...]

PR #188, opened 2018-01-07
Saner XML tests

latk:
The work on #186 and #164  by @libPhipp highlighted two problems with the XML tests:

 1. The XML files are scrubbed with a simple regex to remove the gcovr version. This also removes the XML version, thus creating an invalid XML document. The version regex was made more picky to only scrub gcovr versions.

 2. The tests did not assert that exactly the expected document was created, but would ignore unexpected data. Now the documents must be an exact match. Formatting and order is still ignored.


wolf99:
LGTM, great work @latk & @libPhipp 👍

PR #190, opened 2018-01-08
remove vendored virtualenv script

latk:
For convenience, the repo used to vendor two versions of virtualenv.
This is unnecessary, as virtualenv can be installed separately and venv
is part of the Python 3.3 standard library.


mayeut:
@latk, I forgot to add one remark: should we now consider removing `--exclude=admin` in `.travis.yml` `flake8` test ?

PR #193, opened 2018-01-09
expand README with user + contribution guides

latk:
The current README is a bit spartan. This PR adds:

 - :link: lots of useful links right at the top
 - :framed_picture: screenshots
 - :building_construction: installation instructions
 - :rocket: quickstart documentation
 - :hammer: contributing guidelines

This cannibalizes a bit on the [guide](http://gcovr.com/guide.html), but I'd rather have the important parts right on the repository start page.

A note on the developer role: I've currently set up the pull requests so that they can be merged by any developer if another developer has approved the PR through GitHub's code review features. (And the Travis build must pass, and there mustn't be any merge conflicts). The goal is that every change that is merged into master will have been seen by at least two people. So the developer role is more about reviewing PRs than about writing code.

Currently I am exercising my admin powers to merge my own PRs without approval. But in the long run I'd like to see other people contributing with reviews and merges, if only so that development can continue if none of the admins is at hand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,9 @@ gcovr
 
 generate GCC code coverage reports
 
-[website] • [documentation] • [bugtracker] • [GitHub][repo] • [![Build Status][travis-ci-badge]][travis-ci]
+[website] • [documentation] • [bugtracker] • [GitHub][repo]
+
+[![Build Status][travis-ci-badge]][travis-ci] [![Install from PyPI][pypi-badge]][pypi]
 
 Gcovr provides a utility for managing the use of the GNU gcov utility
 and generating summarized code coverage results. This command is
@@ -37,7 +39,9 @@ Example HTML details:
   [repo]:       https://github.com/gcovr/gcovr/
   [bugtracker]: https://github.com/gcovr/gcovr/issues
   [travis-ci]: https://travis-ci.org/gcovr/gcovr
-  [travis-ci-badge]: https://api.travis-ci.org/gcovr/gcovr.svg?branch=master
+  [travis-ci-badge]: https://travis-ci.org/gcovr/gcovr.svg?branch=master
+  [pypi]: https://pypi.python.org/pypi/gcovr
+  [pypi-badge]: https://img.shields.io/pypi/v/gcovr.svg
 
 Installation
 ------------
--- a/README.md
+++ b/README.md
@@ -1,22 +1,140 @@
-[![Build Status](https://travis-ci.org/gcovr/gcovr.png?branch=master)](https://travis-ci.org/gcovr/gcovr) 
+gcovr
+=====
 
-Gcovr Overview
---------
+generate GCC code coverage reports
+
+[website] • [documentation] • [bugtracker] • [GitHub][repo] • [![Build Status](https://travis-ci.org/gcovr/gcovr.png?branch=master)][travis-ci]
 
 Gcovr provides a utility for managing the use of the GNU gcov utility
 and generating summarized code coverage results. This command is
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,22 +1,140 @@
-[![Build Status](https://travis-ci.org/gcovr/gcovr.png?branch=master)](https://travis-ci.org/gcovr/gcovr) 
+gcovr
+=====
 
-Gcovr Overview
---------
+generate GCC code coverage reports
+
+[website] • [documentation] • [bugtracker] • [GitHub][repo] • [![Build Status](https://travis-ci.org/gcovr/gcovr.png?branch=master)][travis-ci]
```
mayeut: Would you consider switching to the SVG badge format instead of PNG ? It's much more readable I think.
SVG: SVG https://api.travis-ci.org/gcovr/gcovr.svg?branch=master
vs
PNG: PNG https://api.travis-ci.org/gcovr/gcovr.png?branch=master

latk: Fixed! TBH I can't see a difference unless I zoom in, but that's exactly why it's great to have PRs reviewed by another set of eyes :D

README.md
```diff
@@ -1,22 +1,141 @@
-[![Build Status](https://travis-ci.org/gcovr/gcovr.png?branch=master)](https://travis-ci.org/gcovr/gcovr) 
+gcovr
+=====
 
-Gcovr Overview
---------
+generate GCC code coverage reports
+
+[website] • [documentation] • [bugtracker] • [GitHub][repo] • [![Build Status][travis-ci-badge]][travis-ci]
```
mayeut: maybe a pipy badge could be a good thing too ?
[PyPi](https://pypi.python.org/pypi/gcovr)

mayeut:
@latk, thinking more about this, `README.md`is used as-is in `setup.py`
https://github.com/gcovr/gcovr/blob/232fb11fd077ad686a39e95eedb2179add5f216d/setup.py#L25-L27 and
https://github.com/gcovr/gcovr/blob/232fb11fd077ad686a39e95eedb2179add5f216d/setup.py#L38

PyPi does not support markdown syntax (this can be seen easily https://pypi.python.org/pypi/gcovr, you'll see the markdown syntax for the build status image as-is)
In order for README to be rendered properly by both GitHub and PyPi, it should be in reStructuredText markup as mentioned [there](https://packaging.python.org/specifications/core-metadata/#description-optional) and [there](http://python-packaging.readthedocs.io/en/latest/metadata.html#a-readme-long-description)

There are many samples out there for reStructuredText if you're not familiar with it. [Here](https://github.com/mayeut/pybase64/blob/master/README.rst)'s one in one of my repo.
Worth mentioning, all external content (badges, screenshots) will be referenced in PyPi as well. I just  don't bother with that for all badges in my repo which all refer to latest master whatever the version you're showing on PyPI, that might be another issue with screenshots that might disappear from master altogether and, for now, are referenced by a relative URL.

latk:
@mayeut I also noticed the readme rendering problem, but there is a bit more context:

 - PyPI does support markdown readmes, if configured properly. It is possible to strip the badges from the source before showing the readme on PyPI. In any case, this only becomes relevant once we try to publish the next release, which I'd still like to defer for a bit.

 - The documentation is split over the readme (markdown), the guide (asciidoc) and the change log (asciidoc). Should the documentation be unified, using a more standard Python toolchain (perhaps ReST + Sphinx)? This is not entirely trivial.

I'm still pondering on the scope of these problems, and will open a few help-wanted issues once I've found the right questions to ask.

Issue #98, opened 2015-05-28
Coverage Threshold parameter


mrx23dot:
An optional Threshold parameter could be given
e.g. --threshold=90
and when the merged threshold is >90% the script could return 0, otherwise return 1.

Would be great for CI.

mrx23dot:
eg.
parser.add_option(
    "--threshold",
    help="Coverage % threshold value (range:0-100), below this retval:1, otherwise:0, default 90%",
    action="store",
    dest="cov_threshold",
    default=90
)

at the end:
print("\nSummary:");
cov_global = print_summary(covdata);

# retval based on threshold parameter
if (options.cov_threshold is not None):
  thresh = int(options.cov_threshold)
  if (thresh <= 100) and (thresh > 0):
    if cov_global < thresh:
      print("failed, not enough coverage (got:%d, expected min:%d)" % (cov_global, thresh))
      sys.exit(1); # failed, not enough coverage
  else:
    print("Please give value 0-100 for options.threshold", thresh);
    sys.exit(1); # everything was fine
  print("\nMinimum covarage was:%d%%\n" % thresh);

print("\nResult stored in index.html");
print("\nFinished running");
sys.exit(0); # everything was fine

mrx23dot:
eg.
parser.add_option(
    "--threshold",
    help="Coverage % threshold value (range:0-100), below this retval:1, otherwise:0, default 90%",
    action="store",
    dest="cov_threshold",
    default=90
)

at the end:
print("\nSummary:");
cov_global = print_summary(covdata);

# retval based on threshold parameter
if (options.cov_threshold is not None):
  thresh = int(options.cov_threshold)
  if (thresh <= 100) and (thresh > 0):
    if cov_global < thresh:
      print("failed, not enough coverage (got:%d, expected min:%d)" % (cov_global, thresh))
      sys.exit(1); # failed, not enough coverage
  else:
    print("Please give value 0-100 for options.threshold", thresh);
    sys.exit(1); # everything was fine
  print("\nMinimum covarage was:%d%%\n" % thresh);

print("\nResult stored in index.html");
print("\nFinished running");
sys.exit(0); # everything was fine

mrx23dot:
eg.
parser.add_option(
    "--threshold",
    help="Coverage % threshold value (range:0-100), below this retval:1, otherwise:0, default 90%",
    action="store",
    dest="cov_threshold",
    default=90
)

at the end:
print("\nSummary:");
[...]

PR #194, opened 2018-01-11
Add validation for "--fail-under-line" and "--fail-under-branch" options

mayeut:
Add validation. Threshold must be in range 0.0-100.0

c.f. https://github.com/gcovr/gcovr/issues/98#issuecomment-356763747

### DIFF
--- a/scripts/gcovr
+++ b/scripts/gcovr
@@ -44,7 +44,7 @@ import xml.dom.minidom
 import datetime
 import zlib
 
-from optparse import OptionParser
+from optparse import Option, OptionParser, OptionValueError
 from string import Template
 from os.path import normpath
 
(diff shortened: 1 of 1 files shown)


scripts/gcovr
```diff
@@ -2097,7 +2097,25 @@ def print_xml_report(covdata):
 #
 # Create option parser
 #
-parser = OptionParser()
+
+# helper for percentage actions
+def check_percentage(option, opt, value):
+    try:
+        x = float(value)
+        if x < 0.0 or x > 100.0:
```
latk: I'd typically write `if not (0.0 <= x <= 100.0)` so that:

 - the condition we are validating for is spelled positively, and
 - we use chained comparisons to verify the range.

But this is mostly a personal preference. What do you think?

mayeut: It's much nicer. I'm still new to python so learning new things every time.

mayeut:
@latk, I pushed the modification you proposed.
Regarding developer access, I'd rather let people that can commit to this project in a more regular and/or long-term way than I will be able to.
Nevertheless, I will probably drop by now and then if I need to (or if I have a bit of time eventually). When I do, be certain I will always try to contribute with a Pull Request (if the bug fix/enhancement implementation is in my skill set of course). Thank you for your efforts reviving gcovr, it's a great tool and I'm glad to see it's picking up again.

latk:
Thank you, merged!

Developer access does not come with any expectations and implies no commitment – any contributions are entirely voluntary. The only difference would be that you can also contribute by managing issues and PRs if you so wish, in addition to submitting PRs and participating in discussions which you can already do. If you're sure you'll never want to contribute with management tasks that's perfectly fine, and I obviously won't add you against your will. But please let me know if you change your mind :)

PR #195, opened 2018-01-13
Drop support for Python < 2.6

latk:
The code base needs to be compatible across a wide range of Python versions. But that can only go so far. The user guide mentions these supported versions:

> CPython 2.7, 3.4, and 3.5, and PyPy 2.7 and 3.5.

It is also likely that some users are stuck on 2.6, so we shouldn't intentionally break 2.6 compatibility (at least not without a good reason).

But support of earlier versions has disproportianate cost and little value. This PR removes any code that is unreachable under 2.6 or later.

As an additional change, the unit tests now use decorators to annotate the test classes.


hugovk:
Here's the pip installs for gcovr from PyPI for the last month (via `pypinfo --percent --pip --markdown gcovr pyversion`)

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |   65.3% |          3,716 |
| 3.5            |   16.7% |            949 |
| 3.4            |   14.8% |            842 |
| 2.6            |    2.1% |            118 |
| 3.6            |    1.1% |             65 |

latk:
Thank you @hugovk for these statistics – they demonstrate that this change will not hurt users.

And thank you too for the suggestions re classifiers and python_requires, I'll look into those shortly.

The remaining unneeded version checks are constrained to the vendored virtualenv scripts. This will be resolved by deleting the scripts completely in #190.

PR #196, opened 2018-01-15
In HTML output, use LightGray and '-' output when there are no branches

### LINKED ISSUE #105: Branch coverage is reported wrongly
In some cases you might have 0 branches in all files. In that case branch coverage is reported differently in a HTML report:
* per file it is reported as 100%, 0/0
* per module it is reported as exec:0, total:0, coverage: 0%

It looks like 0% value is selected to avoid the zero division. An extra check could be added to show a proper statistics information.

Tested with official gcovr 3.2 release.

mayeut:
Fixes #105


gcovr/tests/nobranch/Makefile
```diff
@@ -0,0 +1,21 @@
+all:
+	$(CXX) -fprofile-arcs -ftest-coverage -fPIC main.cpp -o testcase
+
+run: txt xml html
+
+txt:
+	./testcase
+	../../../scripts/gcovr -d --fail-under-branch 100.0 -o coverage.txt
```
latk: Would it make sense to use the `--branch` report here, to test that it too can deal sensibly with 0/0 branch coverage?

scripts/gcovr
```diff
@@ -1594,8 +1599,8 @@ def print_html_report(covdata, details):
         branchCovered += covered
     data['BRANCHES_EXEC'] = str(branchCovered)
     data['BRANCHES_TOTAL'] = str(branchTotal)
-    coverage = calculate_coverage(branchCovered, branchTotal)
-    data['BRANCHES_COVERAGE'] = str(coverage)
+    coverage = calculate_coverage(branchCovered, branchTotal, zero_value=None)
+    data['BRANCHES_COVERAGE'] = '-' if coverage is None else str(coverage)
```
latk: can be simplified with `... = str(calculate_coverage(..., zero_value='-'))`

scripts/gcovr
```diff
@@ -1728,8 +1732,8 @@ def print_html_report(covdata, details):
         branchTotal, branchCovered, tmp = cdata.coverage()
         data['BRANCHES_EXEC'] = str(branchCovered)
         data['BRANCHES_TOTAL'] = str(branchTotal)
-        coverage = calculate_coverage(branchCovered, branchTotal)
-        data['BRANCHES_COVERAGE'] = str(coverage)
+        coverage = calculate_coverage(branchCovered, branchTotal, zero_value=None)
+        data['BRANCHES_COVERAGE'] = '-' if coverage is None else str(coverage)
```
latk: can also be simplified with `... zero_value='-'`

gcovr/tests/nobranch/Makefile
mayeut: done. It makes perfect sense.

mayeut:
I replaced code with `calculate_coverage()` everywhere I found `100.0 *`
This helped identify one last place where branch coverage will display `0.0 %`. 
The summary (no test for that if I'm not mistaken):
`./scripts/gcovr -s`
```
                           GCC Code Coverage Report
Directory: .
TOTAL                                          0       0    --%
lines: 0.0% (0 out of 0)
branches: 0.0% (0 out of 0)
```
We might want to check `0 / 0` line coverage as well ?

latk:
0/0 line coverage is going to be quite rare (can only happen for an empty report?) so I don't think it's worth the effort to investigate and fix this. Of course you can still look into this if you want.

In the summary report, replacing `0.0%` with `--%` might look odd? I have no clear preference and leave this to your judgement.

Thank you for your effort so far. I'll try to merge this tomorrow, or whenever the Travis builds have caught up. :snail:

mayeut:
I won't look into 0/0 line coverage. As you said it should not happen except for an empty report (TBC).
I have no preference as well for the summary report.

Issue #202, opened 2018-01-22
Plan for the next release


mayeut:
All is in the title, what's the plan for the next release ?

latk:
My very rough roadmap is:

 - merge the dev branch (see #54, #55)
 - merge the HTML improvements (see #43)
 - perhaps some bug fixes, e.g for #191
 - rework the documentation (see https://github.com/gcovr/gcovr/pull/193#issuecomment-358604083)
 - update docs to reflect the changes since the last release
 - publish a 4.0 release, since there are some potentially breaking changes

Not all of that has to happen for the next release. There's actually a good argument for releasing right now: there are already lots of valuable bugfixes, e.g. for Windows usage. But I'd rather hold off for ~2 more months, so that the two large refactoring items (modularization and HTML templating) can be included.

I'd welcome any discussion on what should be included and what should be deferred. I'd also like to hear opinions (especially from users!) whether a there should be a release with the current bug fixes, before more refactoring is done. I can then create a milestone to track progress towards the release.

mayeut:
IMHO there probably should be a release without major refactoring i.e. single script layout.
I don't think I'm the only one seeing `gcovr` used in a questionable way using the 'single script feature'. Having one official release without major refactoring will allow for users to get one last round of bug-fix while planning for the transition to a more pythonic way of dealing with `gcovr` installation.
An argument could be made that if 'single script feature' is to be used, a user might just pick any version in the repo history.

goriy:
I agree with @mayeut that there should be a release before major refactoring.

Current stable release (3.3) simply doesn't work on Windows without manual patching. Current version from repository works fine (at least for my use cases) and I think that it's ready enough for minor release. 

I think that it would be great to have working stable version of traditional (i.e. single script) gcovr before going further.

latk:
Ok, for a 3.4 release the checklist would be:

 - \* optional: merge any remaining bug fixes that should make it to this release
 - \* update guide/readme with bugfixes/improved features
 - \* update guide/readme with warning that installation is only recommended via pip (i.e., don't just put `scripts/` into the PATH)
 - \* update docs with current contributors
 - \* optional: fiddle around with website generation so that the readme is used as the landing page?
 - \* update the changelog
 - \* update copyright information to account for all the contributions since 2013 and from outside Sandia

Then, on a release branch:

 - \* bump version to 3.4 (see admin/release_checklist)
 - \* regenerate docs, incl. capturing new screenshots
[...]

PR #203, opened 2018-01-27
Update changelog

mayeut:
Please don't merge as-is. I will (or even gcovr maintainers can) update this as needed until it's finally good to go. c.f. #202

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -3,6 +3,27 @@
 === 3.4 ''(TODO)'' ===
  - Fix to support filtering with absolute paths
  - Added --html-encoding command line option (#139)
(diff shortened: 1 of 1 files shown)


latk:
Hmm, you beat me by half an hour :D I was working on the changelog in d6d007aa8, as part of #205. Can you go through both versions of the changelog to see if anything was missed? If you update this PR, I will remove the changelog commit from my #205.

I did not copy the commit messages verbatim since they are sometimes not very informative, or only address internal changes that are not relevant for users. The changelog is supposed to assist users that are going to upgrade, or users that want to know when a specific feature was added. If you agree, you can perhaps rephrase some of the changelog entries?

mayeut:
Closing, superseded by #205

PR #204, opened 2018-01-27
Add warning about using gcovr as a single script

mayeut:
Not sure this is the best way of advertising this but let's open the discussion.
c.f. #202

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ Install development version from GitHub:
 
     pip install git+https://github.com/gcovr/gcovr.git
 
+:warning: Even though gcovr could be used as a single python script file, future enhancements will break this capability.
+
 Quickstart
 ----------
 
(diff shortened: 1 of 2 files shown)


latk:
I think this conveys the deprecation well for the README.

Can you also update doc/guide.txt? That's the file that will be displayed on the [website](http://gcovr.com/guide.html#installation). It currently says:

> Gcovr requires virtually no installation. The gcovr command can be downloaded and used directly without installing additional files.

I think Asciidoc styles a paragraph as a warning if it starts with `WARNING: `, see http://asciidoctor.org/docs/asciidoc-writers-guide/#admonitions

mayeut:
Updated guide

latk:
Thank you @mayeut :)

@sinelaw: The README describes how development versions can be installed from GitHub. If you feel that other deployment scenarios are important please issue a PR, although I'd prefer to discourage that installation method for end users. You're right that at the very least, linking to the pip docs would be an improvement – I remember being confused by various installation procedures when I was new to open source software.

Issue #191, opened 2018-01-09
Exclude flag broken on Windows


wolf99:
Since #152  was merged it seems like gcov is now generating detailed HTML reports on Windows. So the following works as expected - yay! 🙌 

```bash
> gcovr -p -b --html --html-details -r . -o "build/artifacts/gcov/GcovCoverageResults.html"
```

However it appears that the exclude flag is now broken:

```bash
>gcovr -p -b -e "^vendor.*|^build.*|^test.*|^lib.*" --html --html-details -r . -o "build/artifacts/gcov/GcovCoverageResults.html"
Traceback (most recent call last):
  File "C:\Users\Me\AppData\Local\Programs\Python\Python36\Scripts\gcovr.py", line 2326, in <module>
    options.exclude[i] = re.compile(os.path.realpath(options.exclude[i]))
  File "C:\Users\Me\AppData\Local\Programs\Python\Python36\lib\re.py", line 233, in compile
    return _compile(pattern, flags)
  File "C:\Users\Me\AppData\Local\Programs\Python\Python36\lib\re.py", line 301, in _compile
    p = sre_compile.compile(pattern, flags)
  File "C:\Users\Me\AppData\Local\Programs\Python\Python36\lib\sre_compile.py", line 562, in compile
    p = sre_parse.parse(p, flags)
  File "C:\Users\Me\AppData\Local\Programs\Python\Python36\lib\sre_parse.py", line 855, in parse
    p = _parse_sub(source, pattern, flags & SRE_FLAG_VERBOSE, 0)
  File "C:\Users\Me\AppData\Local\Programs\Python\Python36\lib\sre_parse.py", line 416, in _parse_sub
    not nested and not items))
  File "C:\Users\Me\AppData\Local\Programs\Python\Python36\lib\sre_parse.py", line 502, in _parse
    code = _escape(source, this, state)
  File "C:\Users\Me\AppData\Local\Programs\Python\Python36\lib\sre_parse.py", line 368, in _escape
    raise source.error("incomplete escape %s" % escape, len(escape))
sre_constants.error: incomplete escape \U at position 2
```

goriy:
Look at Pull Request #158. This patch solved similar problems for me. 
It introduces another way of processing exclude filters on Windows (proper escape etc) and it really works!

latk:
This will be fixed once #158 is merged. But even then, the filter is slightly incorrect because the current working directory is *prepended*. I.e. you would actually end up with this regex:

[...]

Issue #178, opened 2017-09-27
UPdate documentation about how data are filtered


1ce8erg0H:
By default if sources file are not under the root directory they will be ignored.
This is not an obvious design choice and so it should be documented better.

Thank you

latk:
Thank you for pointing out this problem. As of PR #210, the user guide now contains a section on Using Filters. It says:

> If no --filter is provided, the --root is turned into a default filter. Therefore, files outside of the --root directory are excluded.

I will push this version of the user guide to the website tomorrow.

PR #210, opened 2018-02-07
Great documentation overhaul: rst user guide, release simplifications

latk:
In this PR, a number of changes were made to the documentation system, in particular the User Guide.

The guide was translated from AsciiDoc to ReStructured Text (rst). This allows us to more easily include separate files. E.g. example code can be included without the `include_anchors.py` script. The installation instructions and abstract are taken from the README. The list of contributors is taken directly from the AUTHORS file.

Long term, this change will also allow us to manage the documentation with Sphinx.

The drawback of this approach is that the guide.rst is unreadable on GitHub, because the `.. include::` directive is disabled for security reasons. The guide can still be read on the website.

The guide has gained a new section about using filters. How filters behave is one of the more confusing parts of gcovr (see #178). The guide also includes warnings about how Windows filters are expected to change (see #191).

This PR includes a number of related changes:

 - The examples were renamed and simplified. Instead of `example2.sh` we now have `example_xml.sh`. This makes the examples less confusing, at least for me.

 - The guide is no longer rendered as EPUB and PDF. I'm open to reinstating this for a future release, but right now these extra output formats complicate the release process. This also allows us to throw out the included dbtoepub Ruby script (`doc/epub/*`).

 - As AsciiDoc is no longer used, the icon set (`doc/images/*`) could be removed.

 - Screen shots can (optionally) be made automatically by wkhtmltopdf.

This PR solves the last blockers for the next release (see #202). The next step would be to bump the version and trigger a deployment.


whart222:
Moving away from AsciiDoc to Restructured Text is a great idea.  Is the plan to generate documentation with ReadTheDocs?  I've used that strategy on other projects, and I've been happy with the results.

latk:
Thank you @whart222! I hope it's OK that I'm changing so much stuff around.

ReadTheDocs will definitively be the simplest way to host & build the
documentation. But I have never done that before, so if you can provide any
help that would be great :) Some way along the road, I would need your
assistance to point the gcovr.com domain at RTD.

Right now my priority is getting this release out of the door, then working
on the modularization. Moving to Sphinx and RTD is not at the top of my
list, but it would definitively be an improvement. E.g. there's a Sphinx
plugin to display the command line options in a more readable format.


2018-02-07 20:48 GMT+01:00 William Hart <notifications@github.com>:

whart222:
Yep.  I've been enjoying all of the energy that you've brought to gcovr
development.  :) :)

whart222:
Let me see if I can setup RTD.  I already have an account, so this might be
super easy for me.

mayeut:
Nice job. Moving to Sphinx & RTD would definitively be an improvement. This is a first step toward that.

PR #211, opened 2018-02-11
[Fix #93] --html-detailed triggers html reports by itself

### LINKED ISSUE #93: html-details produces output as print_text_report, not print_html_report
Right now if you specify --html-details flag the output will be done using print_text_report, and not print_html_report. A fix like this should remove the issue:

```
diff --git a/scripts/gcovr b/scripts/gcovr
index 792aece..fd83523 100755
--- a/scripts/gcovr
+++ b/scripts/gcovr
@@ -2176,7 +2176,7 @@ if options.verbose:
 #
 if options.xml or options.prettyxml:
     print_xml_report(covdata)
-elif options.html:
+elif options.html or options.html_details:
     print_html_report(covdata, options.html_details)
 else:
     print_text_report(covdata)
```

art-of-dom:
This fixes #93 with the main fix being a simple change to scripts/gcovr. The other changes were to remove the extra --html flags in the test makefiles to confirm the behaviour of the change.

Eventually, we should have separate tests for html and html-detailed, but that seemed like too big of a change for just this fix.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -11,6 +11,7 @@ The following developers contributed to gcovr (ordered alphabetically):
     Attie Grande,
     Bernhard Breinbauer,
     Carlos Jenkins,
+    Dom Postorivo,
     goriy,
     ja11sop,
     Jessica Levine,
--- a/gcovr/tests/bad++char/Makefile
+++ b/gcovr/tests/bad++char/Makefile
@@ -13,7 +13,7 @@ xml:
 
 html:
 	./testcase
-	../../../scripts/gcovr -d --html --html-details -o coverage.html
+	../../../scripts/gcovr -d --html-details -o coverage.html
 
 clean:
 	rm -f testcase
--- a/gcovr/tests/cmake_oos/Makefile
+++ b/gcovr/tests/cmake_oos/Makefile
@@ -24,7 +24,7 @@ xml:
 
 html:
 	cd build && ./gcovrtest
-	../../../scripts/gcovr -d --html --html-details -o coverage.html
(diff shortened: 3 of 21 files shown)


latk:
This looks fantastic, thank you very much! I will merge this after I'm done with the gcovr-3.4 release, so in a few days.

Could you amend your commit to add your name to the AUTHORS.txt list? It is kept in alphabetical order, and you can use any name or handle you like. It is OK to update this PR by force-pushing to your branch.

As this is your first PR, I'm curious: How did you find out about this issue? Did you have any difficulty setting up a development environment for gcovr? Which documentation did you use? How can we make it easier to contribute? You don't have to answer, but an outside perspective would be very welcome. You can also contact me privately (my email can be found in the git log).

art-of-dom:
Just about to force push the change now. The answers to your questions in order are:

1.  I've done work for the ceedling gcov plugin and instead of trying to home bake our own html report we've used gcovr. I've known about this for a bit, but didn't know that it was a quirk instead of a feature until I saw activity on issue #93.
2. It wasn't too hard, but since this is a python project I would have preferred to have a requirements.txt for easy virtualenv use.
3. I actually used the .travis.yml to figure out how to set up the environment and run the expected tests and linting since travis has to set up from scratch typically.
4. The only thing I had an issue with is knowing what was expected in the PR. Specifically, I didn't know to update AUTHORS.txt  and I didn't know how much testing update I should add to this. I've seen other projects use a template for both issues and pull requests to make it easier to convey to the person submitting each what should be included. That might help in this case.

latk:
Thank you, that was extremely helpful. I've added your suggestions to my todo-list and will try to implement them.

Issue #213, opened 2018-02-13
Appveyor tests aren't using Python 3


latk:
I am currently working on modularization. But the tests fail on Windows! It seems when the test Makefiles invoke the gcovr script, it is using this `sys.path` for our Python 3 tests:

```
./subdir/B/testcase
../../../scripts/gcovr -r subdir -d -x -o coverage.xml
Could not find gcovr modules in:
C:\projects\gcovr\scripts
C:\windows\SYSTEM32\python27.zip
C:\Python27\DLLs
C:\Python27\lib
C:\Python27\lib\plat-win
C:\Python27\lib\lib-tk
C:\Python27
C:\Python27\lib\site-packages
Traceback (most recent call last):
  File "../../../scripts/gcovr", line 44, in <module>
    from gcovr.app import parse_arguments
ImportError: No module named gcovr.app
make: *** [Makefile:16: xml] Error 1
```

So gcovr was installed using Python 3, but we try to run it under Python 2 :/

And indeed, the appveyor.yml selects the Python version through an env variable:

``` yml
environment:
  matrix:
    - PYTHON: C:\Python27\python.exe
    - PYTHON: C:\Python36-x64\python.exe
```

In the Makefile, the script is invoked like this, without using that env variable:

```
../../../scripts/gcovr -d -x -o coverage.xml
```

I assume the tests should be updated to explicitly run the selected Python, e.g.

``` Makefile
PYTHON ?= python

xml:
	$(PYTHON) ../../../scripts/gcovr -d -x -o coverage.xml
```

My questions:

 - does this analysis seem sensible?
 - can someone with a Windows machine verify?
 - would someone like to contribute a fix?

mayeut:
I wanted to get some coverage information for gcovr itself and for that, starting moving things around a bit. I just ran into an issue with python 3.6 on Windows which has no reason to be a result of my work and is probably a pre-existing issue, thus confirming your analysis.
I think before modularization is done, it might be a good thing to know what's the current gcovr status (and therefore, what's likely to break without anyone knowing).
I will open a PR for this work which will most probably open some discussions (and which might give you ideas on the issue you're seeing).

PR #214, opened 2018-02-13
Move script/gcovr to gcovr/__main__.py

mayeut:
This allows to run coverage on the module quite easily.
It will also prepare for future re-architecture of gcovr.

### DIFF
--- a/gcovr/__main__.py
+++ b/gcovr/__main__.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 # -*- coding:utf-8 -*-
 #
 # A report generator for gcov 3.4
@@ -2309,7 +2308,9 @@ def build_filter(regex):
         return re.compile(os.path.realpath(regex))
 
 
-def main(options, args):
+def main():
+    global options
+    options, args = parse_arguments()
 
     if options.version:
         sys.stdout.write(
@@ -2430,5 +2431,4 @@ def main(options, args):
 
 
 if __name__ == '__main__':
-    options, args = parse_arguments()
-    main(options, args)
+    main()
--- a/gcovr/tests/filter-test2/Makefile
+++ b/gcovr/tests/filter-test2/Makefile
@@ -20,15 +20,15 @@ run: txt xml html
 txt:
 	./testcase
 	echo "$(GCOVR_TEST_OPTIONS)"
-	../../../scripts/gcovr $(GCOVR_TEST_OPTIONS) -d -o coverage.txt
+	$(GCOVR) $(GCOVR_TEST_OPTIONS) -d -o coverage.txt
 
 xml:
 	./testcase
-	../../../scripts/gcovr $(GCOVR_TEST_OPTIONS) -d -x -o coverage.xml
+	$(GCOVR) $(GCOVR_TEST_OPTIONS) -d -x -o coverage.xml
 
 html:
 	./testcase
-	../../../scripts/gcovr $(GCOVR_TEST_OPTIONS) -d --html-details -o coverage.html
+	$(GCOVR) $(GCOVR_TEST_OPTIONS) -d --html-details -o coverage.html
 
 clean:
 	rm -f testcase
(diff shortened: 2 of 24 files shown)


mayeut:
To run coverage test: `nosetests -v --with-coverage`

gcovr/__main__.py
```diff
@@ -2309,7 +2308,9 @@ def build_filter(regex):
         return re.compile(os.path.realpath(regex))
 
 
-def main(options, args):
+def main():
+    global options, args
```
latk: Is there a specific reason why options needs to be global? While that's technically a precise translation of the status quo, that isn't the semantics I intended.

gcovr/tests/test_gcovr.py
```diff
@@ -1,15 +1,31 @@
...
+
+
 basedir = os.path.split(os.path.abspath(__file__))[0]
+env = os.environ
+coverage_path = os.path.abspath(os.path.join(basedir, '..', '..'))
+python_interpreter = sys.executable.replace('\\', '/')  # use forward slash on windows as well
+
+if 'coverage' in sys.modules:
+    run_coverage = True
+    env['GCOVR'] = python_interpreter + ' -m coverage run --branch --parallel-mode -m gcovr'
```
latk: Using the GCOVR environment variable like this is clever. I like it!

setup.cfg
```diff
@@ -1,2 +1,10 @@
 [bdist_wheel]
 universal=1
+
+[coverage:run]
+branch = True
+parallel = True
```
latk: Can you add a comment that the same options need to be set when building the command in test_gcovr.py?

setup.py
```diff
@@ -51,5 +48,9 @@ def read(*rnames):
       ],
       packages=['gcovr'],
       keywords=['utility'],
-      scripts=scripts
+      entry_points={
+          'console_scripts': [
+              'gcovr=gcovr.__main__:main',
```
latk: Finally 🎉 :) By moving the whole script to `__main__`, you've managed to introduce entry_points much earlier than I would have expected!

gcovr/tests/filter-test2/Makefile
```diff
@@ -2,8 +2,7 @@ CFLAGS= -fprofile-arcs -ftest-coverage -fPIC
 
 BASE_OS:=$(shell uname | cut -d'-' -f1)
 ifeq ($(BASE_OS),MSYS_NT)
-  #GCOVR_TEST_OPTIONS := -f $(shell cygpath -w "$$PWD")\main.cpp
-  GCOVR_TEST_OPTIONS := -f main.cpp
+  GCOVR_TEST_OPTIONS := -f $(shell cygpath -w "$$PWD")\main.cpp
```
latk: Absolute filters are known to not work on Windows right now. I would rather have this test run with a relative filter on Windows than skipping it entirely.

Can you explain how this change is related to collecting coverage or to extracting `__main__`? If there is a need for this commit, OK. I'd otherwise prefer to not merge it.

mayeut:
example of coverage ran through CI: https://codecov.io/gh/mayeut/gcovr/list/module-coverage/

gcovr/__main__.py
mayeut: There are still some functions that are expecting those variables to be defined at the global scope. Without this, `gcovr` fails.

mayeut:
I will reorganize as 3 distinct commits

gcovr/tests/filter-test2/Makefile
mayeut: the rebase was a bit to extensive... I had no preference between running a relative filter or skipping the test. I will re-enable the test as proposed.

mayeut: PS: the test was not passing the correct options anyway before those commits. It worked with python 2.7 but failed with python 3.6 now that tests are run with the correct interpreter on windows.

mayeut:
Fixes #213

gcovr/__main__.py
latk: OK, we will fix that in a future commit when different modules are extracted.

setup.py
goriy: `pip install` on Windows automatically creates `.exe` wrapper in Scripts directory due to this change! It's great! No more manual `.bat` wrappers necessary for cross-platform `gcovr` calls! :+1:

PR #217, opened 2018-02-15
Add test for CoverageData update function

mayeut:


latk:
Very good test case! Thank you for contributing this.

PR #218, opened 2018-02-15
Use codecov service to get coverage information on gcovr

### LINKED ISSUE #216: Add coverage report for gcovr
IMHO, gcovr is missing a public coverage report on CI builds.

It would allow:
- To track coverage progress when adding tests, before refactoring gcovr
- To make sure PR are well tested and reject PR that have an insufficient diff coverage
- For everyone to know what's tested

There are at least 2 service providers that allow to get a report on each CI build that I know of:
- codecov
- coveralls

Here's a sample output using codecov: https://codecov.io/gh/mayeut/gcovr/branch/module-coverage

mayeut:
Fixes #216

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,13 @@ addons:
 matrix:
   include:
   - python: '2.7'
-    env: CXX=g++-5 CC=gcc-5 GCOV=gcov-5
+    env: CXX=g++-5 CC=gcc-5 GCOV=gcov-5 NOSE_WITH_COVERAGE=1
   - python: pypy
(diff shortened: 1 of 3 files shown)


latk:
OK, the coverage seems to have been uploaded successfully: https://codecov.io/gh/gcovr/gcovr/pull/218

The full report is still missing, apparently because no previous coverage was uploaded from master? Weird, but this is going to fix itself once I merge this.

Next steps: adding a coverage badge, and convincing Codecov to update the PR check status. But those are both not very urgent.

Issue #94, opened 2015-05-28
Consistent printing


mrx23dot:
Please use either sys.stdout.write of print() in GCOVR.
I would suggest using print(). (it is compatible with Python 2 and 3 when using like this:

from **future** import print_function
print("anything")

latk:
Thank you for this suggestion! This has been resolved with #228. As a start, we now have a `Logger` class that manages all output messages and warnings.

PR #222, opened 2018-02-17
Add tests for some command line arguments

mayeut:
Test version and help options are printing expected values on stdout.
Test invalid arguments return the correct exit code and print the
correct error message on stderr.

### DIFF
--- a/gcovr/tests/test_args.py
+++ b/gcovr/tests/test_args.py
@@ -1,0 +1,108 @@
+import sys
+import unittest
+
+from ..__main__ import main, version_str
+
+
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
+
+
+# The CaptureObject class holds the capture method result
+class CaptureObject:
+    def __init__(self, out, err, exception):
+        self.out = out
+        self.err = err
+        self.exception = exception
+
+
+# The StringIOBuffered class is used by the capture method to capture stdout
+# and stderr content even if they are closed by the main method
+class StringIOBuffered(StringIO):
+    def __init__(self):
+        self._data = None
+        StringIO.__init__(self)
+
+    def data(self):
+        self.close()
+        return self._data
+
+    def close(self):
+        try:
+            if self._data is None:
+                self.seek(0)
+                self._data = self.read()
+        finally:
+            StringIO.close(self)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+
+
+# The capture method calls the main method and captures its output/error
+# streams and exit code
+def capture(args):
+    with StringIOBuffered() as newout, StringIOBuffered() as newerr:
+        err, sys.stderr = sys.stderr, newerr
+        out, sys.stdout = sys.stdout, newout
+        try:
+            try:
+                e = None
+                main(args)
+            except SystemExit as exception:
+                e = exception
+            return CaptureObject(newout.data(), newerr.data(), e)
+        finally:
+            sys.stderr = err
+            sys.stdout = out
+
+
+class TestArgs(unittest.TestCase):
+
+    def test_version(self):
+        c = capture(['--version'])
+        self.assertEqual(c.err, '')
+        self.assertTrue(c.out.startswith('gcovr %s' % version_str()))
+        self.assertEqual(c.exception.code, 0)
+
+    def test_help(self):
+        c = capture(['-h'])
+        self.assertEqual(c.err, '')
+        self.assertTrue(c.out.startswith('Usage: gcovr [options]'))
+        self.assertEqual(c.exception.code, 0)
+
+    def test_empty_root(self):
+        c = capture(['-r', ''])
+        self.assertEqual(c.out, '')
+        self.assertTrue(c.err.startswith('(ERROR) empty --root option.'))
+        self.assertEqual(c.exception.code, 1)
+
+    def test_empty_objdir(self):
+        c = capture(['--object-directory', ''])
+        self.assertEqual(c.out, '')
+        self.assertTrue(c.err.startswith(
+            '(ERROR) empty --object-directory option.'))
+        self.assertEqual(c.exception.code, 1)
+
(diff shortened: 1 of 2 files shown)


gcovr/tests/test_args.py
```diff
@@ -0,0 +1,108 @@
...
+        try:
+            if self._data is None:
+                self.seek(0)
+                self._data = self.read()
+        finally:
+            StringIO.close(self)
+
+
+@contextmanager
+def ioscope(file):
```
latk: files already are context managers, so I think this function is a no-op?

gcovr/tests/test_args.py
```diff
@@ -0,0 +1,108 @@
...
+    newout = StringIOBuffered()
+    with ioscope(newout):
+        with ioscope(StringIOBuffered()) as newerr:
+            err, sys.stderr = sys.stderr, newerr
+            out, sys.stdout = sys.stdout, newout
+            try:
+                try:
+                    e = None
+                    main(args)
+                except BaseException as exception:
```
latk: It would usually be preferable to only catch `Exception` subclasses, as this will also catch Ctrl-C (`KeyboardInterrupt`). We do however want to intercept `SystemExit`. We later access the `SystemExit.code` field. So should `capture()` _only_ handle SystemExit, rather than all exceptions? The tests will still fail if we get an unexpected exception.

gcovr/tests/test_args.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class CaptureObject:
+    def __init__(self, out, err, exception):
+        self.out = out
+        self.err = err
+        self.exception = exception
+
+
+class StringIOBuffered(StringIO):
```
latk: Is this class needed? I think a `StringIO.getvalue()` call is sufficient in `capture()`, no need to overrride `StringIO.close()`.

gcovr/tests/test_args.py
```diff
@@ -0,0 +1,108 @@
...
+        with ioscope(StringIOBuffered()) as newerr:
+            err, sys.stderr = sys.stderr, newerr
+            out, sys.stdout = sys.stdout, newout
+            try:
+                try:
+                    e = None
+                    main(args)
+                except BaseException as exception:
+                    e = exception
+                yield CaptureObject(sys.stdout.data(), sys.stderr.data(), e)
```
latk: might be simpler as:
``` python
yield CaptureObject(newout.getvalue(), newerr.getvalue(), e)
```

gcovr/tests/test_args.py
```diff
@@ -0,0 +1,108 @@
...
+        with capture(['--object-directory', '']) as c:
+            self.assertEqual(c.out, '')
+            self.assertTrue(c.err.startswith('(ERROR) empty --object-directory option.'))
+            self.assertEqual(c.exception.code, 1)
+
+    def test_branch_threshold_nan(self):
+        with capture(['--fail-under-branch', 'nan']) as c:
+            self.assertEqual(c.out, '')
+            self.assertTrue('not in range [0.0, 100.0]' in c.err)
+            self.assertNotEqual(c.exception.code, 0)
```
latk: In some places we have `assertNotEqual(code, 0)` and others `assertEqual(code, 1)`. That tripped me up because I overlooked the "not". Can we use the `assertNotEqual(code, 0)` everywhere where a specific exit code is not necessary?

gcovr/tests/test_args.py
```diff
@@ -0,0 +1,108 @@
...
+        yield file
+    finally:
+        file.close()
+
+
+@contextmanager
+def capture(args):
+    newout = StringIOBuffered()
+    with ioscope(newout):
+        with ioscope(StringIOBuffered()) as newerr:
```
latk: Can have multiple context managers in one "with" statement, reduces indentation:
``` python
with StringIO() as newout, StringIO() as newerr:
```

gcovr/tests/test_args.py
```diff
@@ -0,0 +1,108 @@
...
+@contextmanager
+def ioscope(file):
+    try:
+        yield file
+    finally:
+        file.close()
+
+
+@contextmanager
+def capture(args):
```
latk: Does this need to be a context manager? It seems to me like the capture data could simply be returned?

gcovr/tests/test_args.py
```diff
@@ -0,0 +1,108 @@
...
+    newout = StringIOBuffered()
+    with ioscope(newout):
+        with ioscope(StringIOBuffered()) as newerr:
+            err, sys.stderr = sys.stderr, newerr
+            out, sys.stdout = sys.stdout, newout
```
latk: reassigning `sys.stdout` is yucky, but there's currently no alternative :(
We'd need to introduce a better logging mechanism first, compare #94.

gcovr/tests/test_args.py
mayeut: StringIO has no context manager on python 2

gcovr/tests/test_args.py
mayeut: I'll change this.

gcovr/tests/test_args.py
mayeut: `StringIO.getvalue()` will raise an exception if `StringIO.close()` has been called. `StringIOBuffered` offers `StringIOBuffered.data()` which will not raise an exception if `close()` has been called.

gcovr/tests/test_args.py
mayeut: Nice, will update this

gcovr/tests/test_args.py
mayeut: I used `assertEqual(code, 1)` when `gcovr` calls `sys.exit(1)` and `assertNotEqual(code, 0)` when parsing fails (we do not know what optparse will call sys.exit with)

gcovr/tests/test_args.py
mayeut: you're right, will update this

PR #223, opened 2018-02-17
Fix HTML output when writing to stdout

mayeut:
This also adds tests using stdout output


gcovr/__main__.py
```diff
@@ -1596,24 +1596,27 @@ def _alpha(key):
...
-            # we add a hash at the end and attempt to shorten the
-            # filename if it exceeds common filesystem limitations
-            if len(os.path.basename(cdata._sourcefile)) < 256:
-                break
-            longname_hash = "_" + hex(zlib.crc32(longname) & 0xffffffff)[2:]
-            longname = longname[(len(cdata._sourcefile) - len(longname_hash)):]
+        if not details:
+            cdata._sourcefile = None
+        else:
+            ttmp = os.path.abspath(options.output).split('.')
```
latk: What specifically does this if/else fix? Can you explain, ideally by adding a comment to the code? My guess: when `options.output is None`, the detail file name calculation breaks. The result is also unneeded unless `--html-details` are requested. The `cdata._sourcefile` field is only assigned so that it can be read later when calling `html_row()` which does not use that value. Did I get this right?

mayeut: You got all that right.

Issue #215, opened 2018-02-14
How to modularize – structure suggestions


latk:
When splitting the gcovr script into multiple smaller modules, how should we divide them? In this issue, we can discuss design decisions before settling on a particular choice.

The modularization that was done on the dev branch uses this structure:

| Module | Contents |
| --- | --- |
| `gcovr` | namespace |
| `gcovr.args` | supplies `parse_arguments()` function |
| `gcovr.data` | `CoverageData`, and processing of gcov data files |
| `gcovr.driver` | supplies `main()` function, and a `main_()` entrypoint |
| `gcovr.path` | path utilities |
| `gcovr.version` | `__version__` declaration and `version_str()` function |
| `gcovr.prints` | namespace |
| `gcovr.prints.html` | `print_html_report()` and supporting functions |
| `gcovr.prints.xml` | `print_xml_report()` function |
| `gcovr.prints.text` | `print_text_report()` and `print_summary()` functions |

I'd prefer a slightly different design:

- get rid of the `gcovr.prints` namespace
- possibly, separate text and summary output
- put argument parsing, main(), and entrypoint into `gcovr.app`
- (later) add `gcovr.logging` which formats verbose diagnostics and error messages
- put gcov invocation and parsing into `gcovr.gcov`

So I'd probably do:

| Module | Contents |
| --- | --- |
| `gcovr` | namespace |
| `gcovr.app` | `parse_arguments()`, `main()`, entrypoint |
| `gcovr.coverage` | `CoverageData` data model that is generated by `gcovr.gcov` and consumed by the output formats |
| `gcovr.gcov` | invoking gcov and parsing gcov files |
| `gcovr.utils` | utilities, including path handling and filters |
| `gcovr.version` | declares `__version__` and nothing else, can be `exec()`d by `setup.py` |
| `gcovr.html` | HTML report |
| `gcovr.xml` | XML report |
| `gcovr.txt` | text report |
| `gcovr.summary`  | summary report |

One thing that bothers me with both of these approaches is the rather generic module names like `html`. Would `html_output` be a better name?

Are there any alternative suggestions? Any weaknesses with the proposed structures?

Once we've decided on a design, implementing the modularization is going to be fairly easy. Further refactoring can come afterwards.

mayeut:
I'd probably go with a slightly different design

Module | Contents
-- | --
`gcovr` | namespace
`gcovr.cli` | `parse_arguments()`, `main()`, entrypoint
[...]

PR #225, opened 2018-02-18
Modularization

latk:
This PR splits the gcovr script into multiple modules, as discussed in #215.


gcovr/__main__.py
```diff
@@ -29,1024 +29,23 @@
...
-from string import Template
 from os.path import normpath
 
-try:
-    xrange
-except NameError:
-    xrange = range
+from .gcov import get_datafiles, process_existing_gcov_file, process_datafile
+from .utils import get_global_stats, build_filter
+from .version import __version__, version_str  # noqa: F401 unused __version__
```
mayeut: Why import `__version__` that requires `# noqa: F401 unused __version__` ?

latk: 🤔 hmm, good question. I just cargo-culted the existing code here because I'm not 100% sure about the Python conventions in this case. I see two alternatives:

- import the version in the `__init__` so that we have a `gcovr.__version__` attribute.
- leave the version in the `gcovr.version` module until we offer an API that would profit from a version attribute.

What do you think would be best?

latk: On second thought, I'll keep `__version__` but remove the `version_str()` function which no longer adds any value over accessing the `__version__` directly.

Issue #231, opened 2018-03-04
Which versions of gcov are supported?


inorton:
Hi, At the moment I'm having to build with quite an old GCC (4.9.2) and occasionally also 3.0.4!

What versions of gcov does gcovr currently support  as a minimum?

Thanks

latk:
Gcovr is currently only being tested with GCC 5, but is know to work with a variety of GCC and Clang versions. It should also work with earlier GCC versions. I think gcovr was originally developed for GCC 3.4.

There is no clear minimum version. I don't have any personal motivation to support anything pre GCC 4.7, but will gladly accept patches should there be any problems.

Looking at old GCC documentation, it seems that gcov 4.x supports all the required options, but gcov 3.x does not. So I do not think you will be able to use gcovr together with GCC 3.0.4.

PR #233, opened 2018-03-04
Check gcov support re #231

inorton:
Early on check that the gcov command we use supports the required command line options.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ nosetests.xml
 .mr.developer.cfg
 .project
 .pydevproject
+
+# Pycharm
+.idea/
--- a/gcovr/gcov.py
+++ b/gcovr/gcov.py
@@ -56,7 +56,8 @@ class GCovSupportChecker(object):
                         pass
 
                     raise EnvironmentError(
-                        "'" + gcov_path + "' does not support " + option)
+                        "gcov command: '" + gcov_path +
+                        "' does not support " + option)
 
         return self.checked[gcov_path]
 
--- a/gcovr/gcov.py
+++ b/gcovr/gcov.py
@@ -25,6 +25,44 @@ exclude_line_pattern = re.compile('([GL]COVR?)_EXCL_(LINE|START|STOP)')
 c_style_comment_pattern = re.compile('/\*.*?\*/')
 cpp_style_comment_pattern = re.compile('//.*?$')
 
+gcov_required_options = [
+    "--branch-counts",
+    "--branch-probabilities",
+    "--preserve-paths",
+    '--object-directory']
+
+
+class GCovSupportChecker(object):
+    """Check that gcov supports the options we need"""
+    def __init__(self):
+        self.checked = {}
+
+    def check(self, gcov_path):
+        if gcov_path not in self.checked:
+            cmd = gcov_path.split()
+            cmd.append("--help")
+            self.checked[gcov_path] = False
+            try:
+                help = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+            except subprocess.CalledProcessError as cpe:
+                help = cpe.output
+            help = str(help)
+            for option in gcov_required_options:
+                if option not in help:
+                    try:
+                        gcov_path = subprocess.check_output(
+                            ["which", gcov_path]).strip().decode("utf-8")
+                    except subprocess.CalledProcessError:
+                        pass
+
+                    raise EnvironmentError(
+                        "'" + gcov_path + "' does not support " + option)
+
+        return self.checked[gcov_path]
+
+
+gcov_support = GCovSupportChecker()
+
 
 #
 # Get the list of datafiles in the directories specified by the user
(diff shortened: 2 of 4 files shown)


latk:
Thank you for suggesting this check. Please note that #228 refractors the
gcovr.gcov module significantly, so this will have to be rebased at some
point. Don't pour too much effort into this just yet.

I'll take a more thorough look at this PR later today.

inorton:
I'm happy that I can get coverage reports now given the right command line arg for --gcov-executable :)

gcovr/tests/test_args.py
```diff
@@ -29,6 +29,12 @@ def test_version(capsys):
     assert c.exception.code == 0
 
 
+def test_incompatible_gcov(capsys):
+    c = capture(capsys, ['--gcov-executable=/bin/sh'])
+    assert c.exception.code != 0
+    assert c.err.startswith("(ERROR) gcov command: '/bin/sh' does not support ")
```
latk: good test, but /bin/sh doesn't exist on Windows systems. To make this test cross-platform, a phony gcov script should be created in the gcovr/tests directory. All it has to do is output a help message mentioning some but not all of the required flags.

gcovr/gcov.py
```diff
@@ -25,6 +25,44 @@
...
+            help = str(help)
+            for option in gcov_required_options:
+                if option not in help:
+                    try:
+                        gcov_path = subprocess.check_output(
+                            ["which", gcov_path]).strip().decode("utf-8")
+                    except subprocess.CalledProcessError:
+                        pass
+
+                    raise EnvironmentError(
```
latk: Is `EnvironmentError` the correct approach here? An alternative would be to have the function return a list of unsupported options, and let the `main()` manage any error messages.

gcovr/gcov.py
```diff
@@ -25,6 +25,44 @@
...
+            try:
+                help = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+            except subprocess.CalledProcessError as cpe:
+                help = cpe.output
+            help = str(help)
+            for option in gcov_required_options:
+                if option not in help:
+                    try:
+                        gcov_path = subprocess.check_output(
+                            ["which", gcov_path]).strip().decode("utf-8")
```
latk: I think this will fail if `gcov_path` contains any extra options, as may be necessary when using gcovr with Clang (`--gcov-executable='llvm-cov gcov'`). Instead perhaps `gcov_path.split()[0]`.

gcovr/gcov.py
```diff
@@ -25,6 +25,44 @@
...
+                    except subprocess.CalledProcessError:
+                        pass
```
latk: Under which circumstances will a CalledProcessError be raised, and why should it be ignored?

gcovr/gcov.py
latk: Is the `which` utility available on Windows, or should the executable path only be resolved on non-Windows systems? I agree that resolving the path leads to better error messages when possible.

gcovr/gcov.py
```diff
@@ -25,6 +25,44 @@
...
+    "--branch-counts",
+    "--branch-probabilities",
+    "--preserve-paths",
+    '--object-directory']
+
+
+class GCovSupportChecker(object):
+    """Check that gcov supports the options we need"""
+    def __init__(self):
+        self.checked = {}
```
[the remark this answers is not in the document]
inorton: I was thinking a little too far ahead, my first attempt put the check directly before the first use of gcov in this file but then I later moved the call to be from main before any actual data processing. So yes, it has no reason to be a class or have a cache any more.

gcovr/gcov.py
inorton: on windows `type` does the same thing as `which` does. So that could be used instead. (Infact, `type` on linux does a very similar job too)

I did also notice that in the other uses of the `options.gcov_cmd` that if for some reason a user supplied a script with escaped or quoted whitespace then things would go wrong too.

gcovr/gcov.py
inorton: In this case it would be if `which` could not find `gcov_path` program in `$PATH`.

gcovr/tests/test_args.py
inorton: What about:
```
c = capture(capsys, ['--gcov-executable="' + sys.executable +" --help"'])
```

.gitignore
```diff
@@ -33,3 +33,6 @@ nosetests.xml
 .mr.developer.cfg
 .project
 .pydevproject
+
+# Pycharm
+.idea/
```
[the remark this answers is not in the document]
inorton: I believe you can but there are quite a few general use cases when you might want to keep the .idea config files as well as others

gcovr/gcov.py
inorton: Yes, maybe the `which` test to print out what the program really was is overkill.

.gitignore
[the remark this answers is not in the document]
latk: I know this is a silly hill to die on, but I won't add gitignore patterns for every editor under the sun. Unless I'm clearly overruled by community consensus.

The Git documentation has [many alternatives](https://git-scm.com/docs/gitignore) to ignore files. Since this pattern is not relevant to other gcovr developers, consider using `.git/info/exclude` instead.

gcovr/tests/test_args.py
latk: That looks good – except of course that `--help` will be added by the check function, so we don't have to specify it here.

gcovr/gcov.py
latk: OK, then we can select between `which` and `type` depending on the OS. That should also fix the tests on Appveyor.

The `type` utility on Posixish systems is a shell builtin rather than a separate program, and cannot be used reliably to resolve the path of an executable (though `type -P $name` happens to do that in Bash).

latk:
Hello @inorton will you be continuing work on this PR? If so, please note that #239 and #228 have changed the `gcovr.gcov` module a lot in the meanwhile, and your work will have to be rebased.

If not, I'll defer this PR. But I would really like to have gcov feature checks so that we can use `gcov -x` when available, as that may help solve issues such as #88 or #7.

latk:
This PR has not seen any progress in a while, so I am closing it for now. I'll be glad to reopen when development this issue continues. I would also welcome it if someone else opens a new PR with equivalent changes (taking into account the discussion here).

Thank you @inorton for suggesting this enhancement! I hope this gets merged someday, but I don't have the time to do the necessary work myself.

Issue #125, opened 2016-01-26
option to add css for html


dewnull:
This is mostly for Jenkins since its default `Content-Security-Policy` setting doesn't allow for execution of inline style. I end up relaxing this policy by modifying Jenkins configuration file  
`
JENKINS_JAVA_OPTIONS="-Dhudson.model.DirectoryBrowserSupport.CSP=\"sandbox; default-src 'none'; img-src 'self'; style-src 'unsafe-inline' 'self';\""
`

Spacetown:
@latk I think, this can be closed.

latk:
This should be resolved via #348 (removal of inline styles) and  657c294371fe5ed5ebc18ec03cf32a14e9864bbf (--no-html-self-contained).

Issue #25, opened 2013-11-03
Provide css/template overrides for the html output.


lu-zero:
Beside what lcov does, would be nice having proper templating so a twitter bootstrap structure can be leveraged.

tytouf:
@lu-zero What is the status on this work? this looks really nice but it appears it has not been integrated yet.

lu-zero:
It lives in my branch and I think somebody volunteered to integrate it by
reimplementing at least part of it so waited to not step on his feet.

On Wed, Jan 7, 2015 at 11:53 AM, Christophe Augier <notifications@github.com

pranavpsawant:
Did we get a chance to integrate it ... because it looks super awesome....

davidvanwyk:
I'd also love for this to be implemented - something akin to [doxygen-bootstrapped](https://github.com/Velron/doxygen-bootstrapped) would be fantastic.

lu-zero:
The code is still on my tree :) anybody willing to help me get it merged is welcome.

oesteban-vx:
So, this is related (implemented?) by #380?

Spacetown:
@oesteban-vx yes.

PR #234, opened 2018-03-05
Use Jinja2 templates for HTML generator

latk:
Previously, the HTML generator used `string.Template` as a rendering mechanism. Now, Jinja2 is used as a rendering engine. The Jinja2 language is very versatile and should make all of this more maintainable. Jinja is documented at <http://jinja.pocoo.org/>.

This adds Jinja2 as a hard dependency! However, that shouldn't be a problem when installing normally through pip or manually using the `setup.py`. Jinja2 supports all of our supported Python versions.

The templates were moved to the `gcovr/templates` directory, so they no longer obfuscate the Python source code. Added benefit: useful syntax highlighting when editing the templates :heart_eyes:

The row-level templates were folded into their main templates, and are no longer rendered separately.

Jinja does slow gcovr down a bit. When running the test suite on my computer, this went from ~21s to ~28s. Initializing Jinja lazily reduced this to ~23s which is probably an acceptable tradeoff. Note that the test suite is not representative of real-world use as it covers many small projects and probably over-emphasizes startup time.

This PR was based on prior work by @lu-zero in #25 and #43. Thank you Luca for forging a way.

Based on this change, improving the HTML should be easier. E.g. I want to add an option to store the CSS in a separate file and remove the inline CSS which (a) simplifies the HTML tests and (b) solves the Jenkins CSP problem (#125).


fletort:
Any doc for this feature somewhere ?
How select which template is used in the command line ?

latk:
Hi @fletort, the templates are not currently exposed to users but are loaded from the package data that is included in gcovr. Because all of this is internal, there is no documentation.

I'd be open to a PR that makes the templates configurable, but the HTML generation code is a bit of a mess. We'd have to define a clear and stable data model that is made available to the templates. I don't want to do that right now because I'm likely to change this data model in the medium future, e.g. to make the reports more themable via CSS (compare #25, #125, #277).

Related to this, the default templates need some love anyway. E.g. they are still using table-based layout. So far I have shied away from fixing that because a very large number of integration tests would have to be updated. The current templates were written to introduce the minimal amount of test changes. They were not designed to be extensible.

If you'd like to work on user-defined templates, please open an issue to discuss the proposed design of the command line flags, the template structure, and what kind of variables should be made available to the templates.

PR #237, opened 2018-03-06
Move from optparse to argparse

latk:
Argparse is the better optparse. The only reason why we hadn't already switched was that argparse was only introduced in Python 2.7. Luckily there's a backport that can be installed from PyPI, so I added a version-dependent dependency.

The difficult part was enabling the necessary tests on Travis CI first. This surfaced a few compatibility problems that were accumulated in the recent refactorings. Especially `subprocess` and `runpy` lack many convenience functions, and format strings `"foo{bar}"` must not have empty placeholders `{}`. The tests cannot rely on the `python -m gcovr` style invocation, but the `gcovr` entrypoint script is available anyway. Now all tests except for an XML example pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,35 +9,38 @@ addons:
     - gcc-5
     - g++-5
 
+env:
+  global:
+  - CXX=g++-5 CC=gcc-5 GCOV=gcov-5
+  - RUN_LINTER=1
+
 matrix:
   include:
+  - python: '2.6'
+    env: RUN_LINTER=
   - python: '2.7'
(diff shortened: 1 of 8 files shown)


mayeut:
@latk, any clue why the one test fails ?

latk:
The remaining test failure in the example XML just seems to be a whitespace issue:

```
E           ValueError: (<root>.coverage.sources.source) Values differ:
E           baseline:
E           u'.'
E           output:
E           u'   .'
```

possibly because of the --xml-pretty layout. There may be additional differences later in the file. The element in the error message is just the report directory.

It would be good to perhaps investigate this failure at some point, but the other tests pass without a problem so I don't think this failure is very important.

PR #238, opened 2018-03-06
Use CXX environment variable in doc/examples tests

mayeut:
All other tests use CXX environment variable that defaults to g++.
That allows to set CXX to g++-5 when it's not the default C++ compiler.
Also remove the now unsued scripts folder from PATH

### DIFF
--- a/doc/examples/example.sh
+++ b/doc/examples/example.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 
-export PATH=`pwd`/../../scripts:$PATH
+g++() {
+	$(which ${CXX:-g++}) $*
+}
 
 #BEGIN compile
 g++ -fprofile-arcs -ftest-coverage -fPIC -O0 example.cpp -o program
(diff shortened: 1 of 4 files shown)


latk:
This is a good idea (necessary actually to get reliable tests). However, do note that snippets of the example scripts are also included verbatim in the documentation. E.g. the user guide would now read:

> 2.1 Tabular Output of Code Coverage
>
> We compile example1.cpp with the GCC compiler as follows:
>
>     ${CXX:-g++} -fprofile-arcs -ftest-coverage -fPIC -O0 example.cpp -o program

(oops, typo! → `s/example1/example/`)

I think that expanding the environment variable like this is going to be too confusing. How can this be prevented?

I think my preferred solution would be to create a shell alias for `g++` instead, only for use within the script.

mayeut:
@latk, I pushed a new commit with another way to get this working without changing the documentation.

latk:
OK, that's a pragmatic approach to this problem. Thank you for working on this!

Issue #3, opened 2013-09-02
parallelize main loop of gcovr


whart222:
It takes a long time to run gcovr on a large project, especially when used as part of an automated analysis system. It would be nice if this loop were parallel:

``` python
for file in datafiles:
  process_datafile(file,covdata,options)
```

(from gcovr ticket 3956)

mrx23dot:
Max number of parallel jobs should depend on numof CPU cores
eg.:
import joblib
print joblib.cpu_count()

others: http://stackoverflow.com/questions/1006289/how-to-find-out-the-number-of-cpus-using-python

oforpertainingtothesun:
I've created a fork for testing this as we've hit this problem compiling on sparc:

https://github.com/JamesReynolds/gcovr

Its early days, but when I'm satisfied its working and I've got the test suite working I'll submit a PR.

JamesReynolds:
We've hit this issue running tests on a Sparc box so I've created a fork here:

https://github.com/JamesReynolds/gcovr

When I'm satisfied this is working and I've got all the tests working I'll submit a PR. Its necessarily python 2.7 for us as we're rather constrained by our environment.

latk:
Thank you @JamesReynolds for looking into this. A bit of advice:

 - Have you run gcovr under a profiler? Where did you find was most time spent? While calling gcov, or while parsing gcov reports? If it is the parsing code, threads are not going to be helpful because of CPython's Global Interpreter Lock: Python threads don't actually execute in parallel.

 - If you find that you need deeper architectural changes, please open an issue so that they can be discussed beforehand. Such changes can then be implemented independently of the parallelization work. If it helps, I can also create a Gitter chat for this repo so that we can discuss questions without having to use GH issues.

 - It's absolutely no problem that you use Python 2.7. But I recommend that you enable Travis and Appveyor for your fork so that you get test reports across multiple Python versions (incl. pypy and 3.x) whenever you push to GH. Simply sign in on their websites with you GH account, and toggle the "on" switch for your repo.

I can understand your desire for a faster gcovr. When I worked with Solaris/Sparc this was so painful that we only ran gcovr on a separate Linux/x86 box.

JamesReynolds:
Thank you! As far as the changes go:

[...]

Issue #36, opened 2014-02-08
Possible to speed up gcovr on big projects?


holycacao:
hi,

i've written a tool, which uses gcovr to generate coverage reports for a big project. i am selectively instrumenting files based on a diff file users pass in (ie if the user changed file1, file2, file3, then touch these files and do a make with gcov flags enabled). in some cases, especially when a user touches a commonly included header file, hundreds of gcno files are created (i presume wherever these header files are included and possibly recursively?). in these cases, gcovr takes up to 3 hours to complete its analysis. i'm curious if there's a safe (maintains coverage accuracy) way for me to speed things up?

mrx23dot:
GCOV.exe could be easily called parallel, based on number of CPU cores.

itavero:
@tsondergaard Did you ever manage to figure out why the numbers for the conditionals are being messed up?
We tried the same approach and our testing time went from over 9 minutes to approximately 3 minutes, which is a big difference if you ask me.
Unfortunately Gcovr now says our conditional coverage is just about half of what it used to be, even though the system under test and the tests have not been changed.

tsondergaard:
@itavero Unfortunately not.

itavero:
I checked and, in our case, the report claims to have *exactly* twice as much branches as they actual have (what the previous reports show).
However the number of branches reached is not doubled.

What we are doing is running all the different tests for our entire code base and then we run `gcov`, similar to what was mentioned before.
After that `gcovr` is ran, again, similar to what was mentioned before.

If anyone has an idea about what might be causing this, I'd love to hear it.

PR #239, opened 2018-03-07
Parallelize main loop

JamesReynolds:
Fixes issue #3 and #36 

The `workers.py` file creates:

1. a `locked_directory` context manager that ensures that each thread uses a separate working directory or blocks waiting for the directory to become free.
2. a `WorkThread` class that runs a job (function and args) but sets `workdir` for the job to a temporary folder
3. a `Workers` class that manages a pool of workers using a queue to pass jobs to them.

Using this `gcovr.py` has been modified to run `gcov` inside a `locked_directory` context and, when complete, will move all of the resulting gcov files to the thread's working directory. This keeps the locked section short (hopefully).

`main.py` has been modified to use the pool Workers.


latk:
Thank you for this awesome feature!

I've skimmed over the PR, and there are lots of unrelated changes regarding Docker and the README. Can we rewrite the history to exclude these changes? I'm not quite sure how to do this, possibly with `git rebase --interactive` or `git filter-branch`. Would you like to try? Otherwise, I'll be glad to help (through this PR, I have write access to your branch). While rebasing, we can also fold fixup commits (like Python 3 modifications) into another commit, as if you had always written correct code on first try.

My goal regarding the git history is that the logical evolution of this code is evident when debugging in the future, not necessarily that it records your actual process.

This PR will conflict with #237, which changes argument parsing to the argparse module. This will change how optional arguments are handled. I think your `parallel_callback` function can then be replaced with a `nargs='?'` argument.

After these points are addressed I'll take another look. E.g. we might be able to avoid some competing gcov processes by grouping work items for the same directory? And it might be possible to avoid spinning up more threads than needed? I don't know yet.

Please let me know if there's anything I can do to help, especially regarding Git history simplification.

JamesReynolds:
@latk After some experimentation I can compress this into a single commit without Dockerfile, .dockerignore or changes to README.rst using git rebase. That probably makes most sense as then my local figuring stuff out commits don't pollute anything?

This will conflict with #237 for sure. What is the best route though? Shall I wait until #237 is added then rebase?

I think we can certainly do better with competing gcov processes for certain build types. Maybe calculating the potential wd in one pass, then running the threads in a second pass selecting items that don't conflict? Another option, but only applicable to later GCC versions, is for two gcov processes to share a directory and get one to use '-x' to append hashes to its output.

latk:
Ok, squashing all commits into one still results in a large commit, but the changes excluding the tests are fairly compact:

    $ git diff --stat master..pr/239 gcovr/*.py
    3 files changed, 207 insertions(+), 23 deletions(-)

So :+1:, go ahead with squashing. Please make sure to keep relevant info from intermediate commit messages (you can edit a commit message via `git commit --amend`). I'll then do a proper code review in the evening or tomorrow.

I don't know whether this or #237 will be merged first, so don't rebase prematurely. Second to merge gets to resolve the conflicts :) 

I don't think `gcov -x` helps here because it hashes the filename, but doesn't generate unique filenames. Two gcov processes that generate coverage for the same file would still clash. But yes, I was also thinking about a two-pass approach.

JamesReynolds:
OK, done. I've merged all of the relevant commits into one.

With `gcov -x` I was thinking that _half_ of the threads could use it and the other half not - so those two sets wouldn't conflict. It feels a bit hacky I'll admit, but it would mean you could (in some cases) double the throughput.

gcovr/workers.py
```diff
@@ -0,0 +1,141 @@
...
+
+class LockedDirectories(object):
+    """
+    Class that keeps a list of locked directories
+    """
+    def __init__(self):
+        self.lock = Lock()
+        self.dirs = set()
+
+    def run_in(self, dir_):
```
mayeut: This shall probably be written using a while loop rather than using recursion.
This shall probably use a condition variable rather than just a lock to prevent the usage of `time.sleep`

JamesReynolds: @mayeut Fair go! I've made the change and rebased again.

mayeut: :+1:

JamesReynolds:
Merged #237

gcovr/workers.py
```diff
@@ -0,0 +1,144 @@
...
+        self.dirs.add(dir_)
+        self.cv.release()
+
+    def done(self, dir_):
+        """
+        Finished with the directory, unlock it
+        """
+        self.cv.acquire()
+        self.dirs.remove(dir_)
+        self.cv.notify()
```
mayeut: This shall be `self.cv.notify_all()` instead of just `self.cv.notify()`

JamesReynolds: @mayeut I see why... that might explain the Windows freeze that appveyor is experiencing - because the right thread isn't being woken and we're deadlocking. Sound right?

JamesReynolds: @mayeut It certainly works now whereas it didn't before!

gcovr/__main__.py
```diff
@@ -472,13 +481,16 @@ def main(args=None):
...
+    pool = Workers(options.gcov_parallel)
+    if options.verbose:
+        print("Pool started with %d workers" % pool.size())
     for file_ in datafiles:
         if options.gcov_files:
-            process_existing_gcov_file(file_, covdata, options)
+            pool.add(process_existing_gcov_file, file_, covdata, options)
         else:
-            process_datafile(file_, covdata, options)
+            pool.add(process_datafile, file_, covdata, options)
```
latk: The `covdata` dict will be updated from multiple threads, but is not synchronized. Should this access be guarded through a mutex section? Or should the workers write update commands to an output queue that is then executed on the main thread?

There may also be concurrent accesses to one covdata entry if multiple threads provide coverage for the same file, in particular for header files that are included in multiple parts of the project.

gcovr/gcov.py
```diff
@@ -700,15 +706,21 @@ def select_gcov_files_from_stdout(out, gcov_filter, gcov_exclude, logger):
...
-        active_files.append(fname)
+        if tempdir:
+            import shutil
+            active_files.append(os.path.join(tempdir, fname))
+            if chdir != tempdir:
+                shutil.copyfile(os.path.join(chdir, fname), active_files[-1])
+        else:
+            active_files.append(os.path.join(chdir, fname))
 
     return active_files, all_files
```
latk: `active_files` is the list of files that should be processed, and `all_files` the list of files that should be kept or deleted, depending on the `--keep` flag. Here, `all_files` only receive the `fname` without the `chdir` prefix. I think that is a bug unless `chdir == "."`.

Additionally, the tempdir files are not part of the all_files and will not be cleaned up until the worker exists. And we are copying and not moving the files to the tempdir even when they are guaranteed to be removed later. This isn't incorrect, but it feels inelegant. I don't know what the correct solution would be. Maybe a third return values for files that should be always deleted after processing?

gcovr/gcov.py
```diff
@@ -677,7 +683,7 @@ def run_gcov_and_process_files(
     return done
 
 
-def select_gcov_files_from_stdout(out, gcov_filter, gcov_exclude, logger):
+def select_gcov_files_from_stdout(out, gcov_filter, gcov_exclude, logger, chdir, tempdir=None):
```
latk: `tempdir=None` never happens, as a `workdir` is always set when executing through the thread. I think these arguments can be made mandatory.

gcovr/workers.py
```diff
@@ -0,0 +1,144 @@
...
+    def close(self):
+        """
+        Empty the working directory
+        """
+        import shutil
+
+        # On Windows the files may still be in use. This
+        # is unlikely, the files are small, and are in a
+        # temporary directory so we can skip this.
+        shutil.rmtree(self.workdir, ignore_errors=True)
```
mayeut: Why would the files still be in use on Windows ? This doesn't seem right. Any pointers ?

gcovr/workers.py
```diff
@@ -0,0 +1,144 @@
...
+        Get the next piece of work
+        """
+        return self.q.get(False, 5)
+
+    def wait(self):
+        """
+        Wait until all work is complete
+        """
+        for w in self.workers:
+            w.start()
```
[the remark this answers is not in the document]
mayeut: I fully agree with @latk, It would be best to start as soon as the first item is enqueued (it would be even better if `get_datafiles` returned a generator rather than a list, I have not checked if it's possible though).
It would also help get rid of  `self.q.get(False, 5)` which seems dangerous (Imagine for one reason or another, e.g. the python process is denied a cpu slot for 5 seconds, what would happen).

gcovr/__main__.py
mayeut: Good remark, access to `covdata` shall be synchronized. One other possibility not mentioned by @latk is to have one `covdata` dict per thread and do a threaded merge at the end (nb thread/2 at each merge step).

gcovr/workers.py
```diff
@@ -0,0 +1,144 @@
+# -*- coding:utf-8 -*-
+
+# This file is part of gcovr <http://gcovr.com/>.
+#
+# Copyright 2013-2018 the gcovr authors
+# Copyright 2013 Sandia Corporation
```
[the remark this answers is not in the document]
JamesReynolds: Done, will merge (won't rebase yet) when all are completed.

gcovr/workers.py
```diff
@@ -0,0 +1,144 @@
...
+class WorkThread(Thread):
+    """
+    The work thread class continuously gets work and
+    completes it
+    """
+    def __init__(self, pool):
+        """
+        Initialise with a reference to the pool object
+        which houses the queue
+        """
```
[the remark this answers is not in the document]
JamesReynolds: Done, will merge (won't rebase yet) when all are completed.

gcovr/workers.py
```diff
@@ -0,0 +1,144 @@
...
+                work, args, kwargs = self.pool.get()
+            except Empty:
+                break
+            kwargs['workdir'] = self.workdir
+            work(*args, **kwargs)
+
+    def close(self):
+        """
+        Empty the working directory
+        """
```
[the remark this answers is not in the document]
JamesReynolds: Done, will merge (won't rebase yet) when all are completed. Very good suggestion - that saves a fair few lines of code and makes much more sense.

gcovr/workers.py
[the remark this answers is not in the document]
JamesReynolds: Done, will merge (won't rebase yet) when all are completed. It makes the code neater to use a sentinel and it does mean that the threads start processing as soon as possible. Thinking about it both time.sleep and the timeout here are code smells I should've jumped on!

gcovr/workers.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+
+    def __init__(self, number=0):
+        """
+        Initialise with a number of workers
+        """
+        self.q = Queue()
+        if number == 0:
+            from multiprocessing import cpu_count
+            number = cpu_count()
```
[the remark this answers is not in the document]
JamesReynolds: Done, will merge (won't rebase yet) when all are completed. Again, much neater.

gcovr/gcov.py
JamesReynolds: I've fixed the all_files bug. 

I think the move vs copy etc... is a bit more of a thorny issue. Windows is throwing file-in-use errors when either moving or deleting. I think the best option would be to copy (as its clear what is happening and avoids a process lock if there is one), keep track of all the files, then to delete after all the threads are complete. I'm working on making Windows happy today and this is what I'll try first.

I've also had to add exception tracking to the threads so their exceptions get raised on the main thread.

gcovr/gcov.py
JamesReynolds: Done, will merge (won't rebase yet) when all are completed.

gcovr/workers.py
JamesReynolds: Indeed it does not seem right - as these are just parsed by python and not touched by gcov. I think this may be able to go once I've cured the (potential) issue with deletion of the *.gcov files. I've set up a local Windows test machine now so I should get this done today.

gcovr/__main__.py
JamesReynolds: I've created a LockedDictionary object that updates with a lock. As this is only called in one place after the recent refactor to gcov.py this seemed to be the easiest option.

However... keeping a context per thread and merging is a nice pattern and avoids further locks. It also may cure help the problem in the next comment...

JamesReynolds:
@latk @mayeut I've submitted another revision (not rebased though so you can see the change) which I think addresses all the issues. I went with a sentinel object and with a per-thread context that I merge in __main__. It solves the deadlock and crashes on Windows and I think gives marginally better performance on Sparc too.

gcovr/workers.py
```diff
@@ -0,0 +1,119 @@
...
+        work, args, kwargs = queue.get(True)
+        if not work:
+            break
+        kwargs.update(context)
+        try:
+            work(*args, **kwargs)
+        except:  # noqa: E722
+            import sys
+            exceptions.append(sys.exc_info())
+            break
```
latk: If one thread exits due to an exception, the others will continue until all work is done (or all have died). So with each exception, processing slows down further. This could e.g. happen if some gcov files have parse errors. I'm also thinking about how this integrates with Ctrl-C/`KeyboardInterrupt` exceptions.

Can we somehow turn this into fail-fast behaviour? E.g. if an exception is caught, the input queue is drained?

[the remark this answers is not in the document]
JamesReynolds: Agreed. Job for Monday!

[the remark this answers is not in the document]
JamesReynolds: I think list appends are thread-safe (http://effbot.org/zone/thread-synchronization.htm) but I'm having trouble finding a completely official line on that. If that is the case then I think two threads would always produce two, distinct, exceptions. It does specifically say that a deque does do this atomically though, so I could switch to that?

JamesReynolds: I agreed. A job for Monday - I'll make sure that if an exception is raised the queue is drained and a sentinel pushed. I think I also need to wrap the call to this in `__main__.py` in a context manager that can terminate threads if an exception is thrown prior to the sentinels being pushed. It would probably be neat to make the pool itself a context.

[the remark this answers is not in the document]
latk: OK, that sounds reasonable. There's no reason why `list.append()` would have to release the GIL, and I couldn't find anything surprising in the CPython source code. So I think we can keep using the list, iff `exceptions` is only consumed after all threads are joined (as  is the case in the current design).

gcovr/__main__.py
mayeut: Just forget my comment regarding threaded merge. This won't  do any good running only python code. I should have known better as I always try to release the GIL in long running tasks in C extensions to allow multithreading... Python only runs python code in one thread at a time.

latk: @mayeut Just to clarify, you're saying “multi-level merging is pointless in Python because merging can't happen in parallel”, but not “collecting separately and merging at the end is unnecessary”, right? So the current design (with a per-thread `covdata` dictionary) is correct as per your experience?

mayeut: Right.

mayeut: @latk, I really need to think "python". Thinking again about your question. I did not answer correctly. So, the only thing that can benefit from threading are I/O here (subprocess being part of that). Updating the coverage dictionary is pure python with no I/O, 2 threads can't really do that in parallel (compute time wise) so, we're probably better off protecting access to the dictionary rather than merging at the end since there can't really be contention here. Merging can be easier (thread safety) but not faster.

gcovr/workers.py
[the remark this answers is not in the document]
JamesReynolds: I've added a test, its a bit involved as I had to add some events to ensure that I could deterministically check that the queue was drained properly.

JamesReynolds: This is done in the latest commit, on an exception the queue is drained and there is code to ensure that the sentinels persist - so that if the drain occurs prior to pool.wait() being called the threads still stop. I've also wrapped the pool in a context manager so that keyboard interrupts are caught and the queue drained from main.

latk: Thank you for these updates. At first glance this looks quite good, but a proper review will have to wait until I'm less tired.

gcovr/tests/test_gcov_parser.py
```diff
@@ -276,3 +280,39 @@ def test_pathologic_codeline(capsys):
...
+        with Workers(threads, lambda: {'total': mutable, 'exc_raised': exc_raised, 'queue_full': queue_full}) as pool:
+            for extra in range(0, 10000):
+                pool.add(check_and_raise, extra)
+
+            # Queue is filled
+            queue_full.set()
+
+            # Wait until the exception has been completed
+            while not pool.exceptions:
+                time.sleep(0)
```
latk: Is `sleep(0)` guaranteed to yield from this thread and let other threads run? Or might this be a `while True: pass` style infinite loop?

gcovr/workers.py
```diff
@@ -95,7 +113,8 @@ def add(self, work, *args, **kwargs):
         Add in a method and the arguments to be used
         when running it
         """
-        self.q.put((work, args, kwargs))
+        if not work or not self.exceptions:
+            self.q.put((work, args, kwargs))
```
latk: This methods seems to do: add a work item to the queue unless there already was an exception. Always add the end sentinels. The check is not needed for correctness, but is only used as an optimization to avoid enqueueing items that would be drained immediately. Is that correct?

It might be a lot easier to understand if the end sentinels are added through a different method, and only keep the `if not exceptions: put(…)` check.

gcovr/workers.py
```diff
@@ -110,10 +129,19 @@ def wait(self):
         for w in self.workers:
             self.add(None)
         for w in self.workers:
-            w.join()
+            while w.is_alive():
+                w.join(timeout=0.1)
```
latk: Why is it necessary to use a timeout here?

gcovr/tests/test_gcov_parser.py
```diff
@@ -276,3 +280,39 @@ def test_pathologic_codeline(capsys):
...
+
+            # Queue should be drained and exception raised
+            exc_raised.set()
+            pool.wait()
+
+    # Outer level catches correct exception
+    assert excinfo.value.args[0] == "Number == 0"
+
+    # Fewer than 1 job per thread is executed
+    assert mutable[0] < threads
```
latk: I think it would be possible that exactly one job per thread is dequeued. Should the check be changed to “At most 1 job per thread”: `assert mutable[0] <= threads`?

gcovr/tests/test_gcov_parser.py
```diff
@@ -276,3 +280,39 @@ def test_pathologic_codeline(capsys):
...
         '(ERROR) Exiting',
         'run gcovr with --gcov-ignore-parse-errors')
+
+
+def check_and_raise(number, total, exc_raised, queue_full):
+    queue_full.wait()
+    if number == 0:
+        raise Exception("Number == 0")
+    exc_raised.wait()
+    total[0] += 1
```
latk: Increments are not threadsafe. In this case it's not strictly important as we are only asserting an upper bound on this count. However, it might be better to append an item to this array here, and use the length of the array later.

gcovr/workers.py
```diff
@@ -57,6 +57,22 @@ def locked_directory(dir_):
...
 
 
+def drain(queue):
+    """
+    Drain a queue of all items
+    """
+    sentinel_count = 0
+    while True:
+        try:
+            work, args, kwargs = queue.get(False)
```
latk: Is it correct to use nonblocking operations here? I think that would make it possible that items are added after this loop completes and before sentinels are enqueued.

Instead, it might be better to block until we get a sentinel? Or could that lead to deadlocks, which are worse than a few extra items? Or should we enqueue a sentinel before we start draining it. Or should we introduce a condition variable that controls whether the queue is open? That condvar would then be set as soon as an exception is received.

gcovr/tests/test_gcov_parser.py
JamesReynolds: It is (https://docs.python.org/3/library/sched.html) but it was a pain to find! Waiting like this is obviously not great, but I didn't want to complicate the workers code to make the test cleaner.

gcovr/workers.py
JamesReynolds: That is correct. I can make this change if you would like? I agree it would make the code neater to have an `add_sentinel` method.

gcovr/workers.py
JamesReynolds: I found the answer here (https://stackoverflow.com/questions/3788208/threading-ignores-keyboardinterrupt-exception) - it was the only way I could get a KeyboardInterrupt to get reported during a join. I don't like it myself, but I think its better than making the threads daemons.

gcovr/tests/test_gcov_parser.py
JamesReynolds: That is true - but one of those jobs is to throw an exception, so it doesn't increment the counter. ...which is not even slightly explained by my comment! Again, I could change the comment prior to the merge to say "increment" instead of "job"?

gcovr/tests/test_gcov_parser.py
JamesReynolds: That is true - but it could mean that we miss failed tests if something goes wrong in the future?

gcovr/workers.py
JamesReynolds: I'm not sure about this myself. I was aiming for simplicity with a possibility of some extra work in the case of error. Perhaps abstracting the error mechanism into the Workers class so that the worker asks the Workers class to register an exception. At that point the Workers class could put locks around adding work to the queue and raising errors + draining the queue.

I think if we don't allow an addition at the same time as raising an exception - and we use a separate mutex to the queue - we shouldn't get deadlock.

JamesReynolds:
I'm happy to do one more round of changes - but as this code is only for the error case, it may be best to refactor later. Your choice!

gcovr/tests/test_gcov_parser.py
latk: Good find! You can leave this as it is, or add a comment.

gcovr/workers.py
latk: Your call. Perhaps just add a comment. Thankyou for confirming my understanding.

gcovr/tests/test_gcov_parser.py
latk: I have a preference for correct code because it prevents “WTF!!?!”s in the future. In this case, either a fix or a comment would be fine.

gcovr/workers.py
latk: I think maybe the `if not exceptions` test in `add()` actually serves as a kind of mutex and prevents enqueueing of extra items, if the `exceptions` is appended before draining.

In any case – you are the multithreading expert here. If you think you have a better solution it might be worth a shot.

At this point, it may be best to extract all queue operations (add, sentinel, dequeue, drain) into a separate class so that it's easier to see what is going on. But that can also be done outside of this PR.

latk:
Thank you @JamesReynolds! I'm very grateful for you doing all this work and I don't want to exhaust your motivation with unnecessary nitpicking. If you're happy to go for one more round with the above comments that would be great.

gcovr/tests/test_gcov_parser.py
JamesReynolds: Added a comment.

gcovr/workers.py
JamesReynolds: I made an add_sentinel method - it fitted well with the refactor to ensure draining and adding work couldn't conflict.

gcovr/workers.py
JamesReynolds: I've added a comment and gone with timeout=1, for my sparc box that means things stop within 10 seconds (the gcov run is ~8 or 9s).

gcovr/workers.py
JamesReynolds: I've moved things into the pool and supplied a lock. I agree it makes things easier to read to have the methods split up and it means that the mutex is explicit and checked properly.

JamesReynolds:
No problem! I'd rather get this right and clean even if it takes a little longer. Should we rebase this down to a single change as well?

JamesReynolds:
Great! I've checked our SPARC improvements and it's still good. Perhaps the additional copy (which is not required in the single threaded case) adds additional I/O overhead in the tests?

PR #244, opened 2018-03-11
Add handling of gcov error message "Could not open output file"

### LINKED ISSUE #243: Gcovr does not handle "Could not open output file" error message.
#Gcov produces error message "Could not open output file" in e.g. case it tries to create temporary *.c.gcov file beyond 259 chars limitation on Windows (or path+name will exceed this limitation), but gcovr does not handle this type of error message. Therefore the gcovr continues and does not create report for this file.
Gcovr handles correctly error message "Cannot open source file" comming from gcov, but not error message "Could not open output file".

MartinXMraz:
fixes #243.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -23,6 +23,7 @@ The following developers contributed to gcovr (ordered alphabetically):
     libPhipp,
     Lukas Atkinson,
     Luke Woydziak,
+    Martin Mraz,
     Matsumoto Taichi,
     Matthias Schmieder,
     Matthieu Darbois,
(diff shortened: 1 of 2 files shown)


AUTHORS.txt
```diff
@@ -23,6 +23,7 @@ The following developers contributed to gcovr (ordered alphabetically):
     libPhipp,
     Lukas Atkinson,
     Luke Woydziak,
+	Martin Mraz,
```
latk: You used tabs instead of spaces here – could you update that?

gcovr/gcov.py
```diff
@@ -17,7 +17,7 @@
 from .utils import aliases, search_file, Logger
 
 output_re = re.compile("[Cc]reating [`'](.*)'$")
-source_re = re.compile("[Cc]annot open (source|graph) file")
+source_re = re.compile("[Cc][annot|ould not] open (source|graph|output) file")
```
latk: I think this needs to use a regex group, not a character class for `[annot|ould not]` → `(annot|ould not)`.

latk:
Thank you very much for fixing this!

PR #247, opened 2018-03-19
Fix --print-summary

thejk:
Just a quick fix for:
TypeError: print_summary() takes 1 positional argument but 2 were given


latk:
Good find! I guess we should add tests for the summary report at some point.

Before I merge, could you amend the commit to add your name to the AUTHORS file?

thejk:
Sure, always forget something ;)

Issue #235, opened 2018-03-05
Manage documentation with Sphinx


latk:
Sphinx <http://www.sphinx-doc.org/> can manage a set of RestructuredText documents. It is widely used to manage the documentation of Python projects.

Currently, our docs are structured like this:

 - `README.rst` serves as an entry point to gcovr, and must render well on GitHub and on PyPI.
 - `CONTRIBUTING.rst` is the development guide, and must render well on GitHub.
 - `AUTHORS.txt`, `CHANGELOG.txt` and `LICENSE.txt` are various metadata.
 - `doc/guide.rst` is a more in-depth user guide. It is published on the gcovr home page <http://gcovr.com/guide.html>. It includes various examples, and the `gcovr --help` output.

The `doc/examples` directory includes a few example programs that are included in the documentation. The screenshots are based on these examples.

I want to use Sphinx so that I can (at some later time):

 - split the guide into smaller documents
 - use some extension for better command line option documentation (fixed-width text is difficult to read)
 - use the readme (or something based on it) as the gcovr home page (the current homepage is super minimal, and is maintained by hand)
 - host the readme on readthedocs.io
 - build and publish new docs automatically through Travis CI.

Would anyone like to take a shot at Sphinx? This shouldn't be too difficult. Note that the `*.txt` files are not yet translated to ReST, and that the current documentation build process also executes tests and can automatically make screenshots. I would like to preserve this ability under Sphinx.

I'll be happy to assist in any way that I can.

chrta:
Maybe we could also use http://www.sphinx-doc.org/en/master/ext/extlinks.html in the changelog to link to the related issues. This is a bundled extension (no extra dependency).

latk:
@chrta That's a very good idea! Would you like to implement this as a separate pull request?

chrta:
Linking to github issues from the changelog is done in #249. This should only contain the relevant commit after #248 is merged.

PR #248, opened 2018-03-23
Create documentation with sphinx

chrta:
This is a first shot for #235

Only some initial work is done for now.

- The project was created with sphinx-quickstart
- readthedocs theme is used
- "make html" works
- the old functionality to take screenshots and execute tests should be preserved

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,8 @@
-= `gcovr` Release History and Change Log =
+``gcovr`` Release History and Change Log
 
-=== 3.4 ''(12 February 2018)'' ===
+3.4 (12 February 2018)
+----------------------
+
  - Added --html-encoding command line option (#139).
  - Added --fail-under-line and --fail-under-branch options,
    which will error under a given minimum coverage. (#173, #116)
@@ -20,7 +22,9 @@
  - Fix verbose output when using existing gcov files (#143, #144)
 
 
-=== 3.3 ''(6 August 2016)'' ===
+3.3 (6 August 2016)
+-------------------
+
  - Added CI testing using TravisCI
  - Added more tests for out of source builds and other nested builds
  - Avoid common file prefixes in HTML output (#103)
@@ -33,7 +37,9 @@
  - More careful checks for covered lines (#109)
 
 
-=== 3.2 ''(5 July 2014)'' ===
+3.2 (5 July 2014)
+-----------------
+
  - Adding a test for out of source builds
  - Using the starting directory when processing gcov filenames.
    (#42)
@@ -44,7 +50,9 @@
  - Adding logic to cover branches that are ignored (#28)
 
 
-=== 3.1 ''(6 December 2013)'' ===
+3.1 (6 December 2013)
+---------------------
+
  - Change to make the -r/--root options define the root directory 
    for source files.
  - Fix to apply the -p option when the --html option is used.
@@ -56,7 +64,9 @@
    option.
 
 
-=== 3.0 ''(10 August 2013)'' ===
+3.0 (10 August 2013)
+--------------------
+
  - Adding the '--gcov-executable' option to specify
    the name/location of the gcov executable. The command line option
    overrides the environment variable, which overrides the default 'gcov'.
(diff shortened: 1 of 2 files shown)


doc/README.txt
```diff
@@ -1,11 +1,15 @@
 This directory supports the creation of the Gcovr User Guide using
-docutils (`rst2html5.py`).
+sphinx.
+
+The following python packages have to be installed, e.g. with pip:
+
+    pip install sphinx sphinx_rtd_theme
```
chrta: Or should this be executed in a virtualenv and requirements.txt should be provided?

doc/Makefile
```diff
@@ -1,34 +1,39 @@
...
-screenshot-%.png: examples/example-%.html
+source/_static/screenshot-%.png: examples/example-%.html
 	$(SCREENSHOT_FROM_HTML) $< $@
 
 examples/example-%.html:
 	cd examples; ./example_html.sh
 
 outputs:
-	../scripts/gcovr -h > examples/gcovr.out
+	gcovr -h > examples/gcovr.out
```
chrta: How does gcovr end up in ../scripts? Maybe this is missing in some documentation or i did not find it.

latk:
Fantastic! Thank you for working on this. I'll look into this in the evening.

doc/Makefile
latk: Thank you for spotting this problem. Originally gcovr was a single file in the `scripts` directory. Fairly recently, it was refactored into a proper module that provides a `gcovr` executable upon installation. So your change is correct.

doc/README.txt
latk: Currently all development prerequisites are documented in the CONTRIBUTING document. You can mention the sphinx dependencies there. If you list the deps in a requirements.txt that would probably be even better (in which case CONTRIBUTING and doc/README and possibly also .travis.yml and appveyor.yml can be simplified).

chrta: I added a requirements.txt without pinning the versions.

chrta:
How about the structure used in http://sphinx-rtd-theme.readthedocs.io/en/latest/index.html and https://raw.githubusercontent.com/rtfd/sphinx_rtd_theme/master/docs/index.rst?
The readme at https://github.com/rtfd/sphinx_rtd_theme is very short and points to all relevant documents. The detailed documentation is only available at readthedocs.io

If i include the README.rst as it is now, the paths to the images do not work and the toc structure is not very nice.

latk:
Hmm, that's a difficult problem. I would strongly prefer to keep the README useful without having to follow external links. 

Perhaps an alternative would be to move the documentation directory to the root directory? If the `doc` directory gets in the way, we can get rid of it. Then, the `README.rst` could be used as the master document instead of a `doc/index.rst`? What do you think?

chrta:
I would keep the documentation in the docs folder. Otherwise the sphinx files, makefile etc would be in the same folder as non-doc stuff.

Maybe we can include snippets from the main README.rst in the doc/source/index.rst file, like in the guide.rst file:

~~~
.. include:: ../../README.rst
   :start-after: .. begin abstract
   :end-before: .. end abstract
~~~

This way we would reduce copy-and-paste an little bit and can setup a documentation structure freely for sphinx.

doc/source/index.rst
```diff
@@ -0,0 +1,17 @@
+gcovr
+=====
+
+.. include:: ../../README.rst
+   :start-after: .. begin abstract
+   :end-before: .. end abstract
+
```
latk: When I change the homepage to the Sphinx-generated version, I'd still like to keep the links from the [current homepage](http://gcovr.com/) here. Can we add a paragraph with the following links here so that they are still visible “above the fold”?

 - (no longer needed: link to User Guide)
 - Getting Help
    - Submit a ticket → https://github.com/gcovr/gcovr/issues
    - Stack Overflow → http://stackoverflow.com/search?q=[gcovr]
 - Download
    - PyPI → https://pypi.python.org/pypi/gcovr/ (though we could use this opportunity to switch the link to https://pypi.org/project/gcovr/)
- Source Code
    - Browse Source (GitHub) → https://github.com/gcovr/gcovr
    - Source Downloads → https://github.com/gcovr/gcovr/releases
    - Change Log → https://raw.github.com/gcovr/gcovr/master/CHANGELOG.txt

Though it might be better to include the changelog within the documentation – I think we'd only have to fix the formatting to use Restructured Text.

doc/source/index.rst
```diff
@@ -0,0 +1,17 @@
...
+.. toctree::
+   :maxdepth: 2
+   :numbered:
+   :caption: Contents:
+
+   installation
+   quickstart
+   contributing
+   license
+   guide
```
latk: For the initial switch to Sphinx-based docs, I'd initially avoid separate chapters for installation & quickstart since there are similar sections in the guide. We can later refactor that.

Since the guide is the most important part, I'd put it first:

 - guide
 - contributing
 - (changelog?)
 - license

What do you think of this structure?

doc/source/conf.py
```diff
@@ -0,0 +1,164 @@
...
+# -- Project information -----------------------------------------------------
+
+project = u'gcovr'
+copyright = u'2018, William Hart, John Siirola and and Lukas Atkinson.'
+author = u'William Hart, John Siirola and and Lukas Atkinson.'
+
+# The short X.Y version
+version = u''
+# The full version, including alpha/beta/rc tags
+release = u''
```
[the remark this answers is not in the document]
chrta: We could get htese information from the last git tag on this branch using `git describe`:

~~~
#to get the git revision (see http://dreamiurg.net/2011/10/03/using-git-to-get-version-information/)
from subprocess import Popen, PIPE

def get_version():
    #"""
    #Returns project version as string from 'git describe' command.
    #"""
    pipe = Popen('git describe --tags --always', stdout=PIPE, shell=True)
    version = pipe.stdout.read()

    if version:
        return version.decode('utf-8')
    else:
        return 'X.Y'

version = get_version().rstrip()
release = version
~~~

Is this ok?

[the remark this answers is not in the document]
mayeut: Why not use the same kind of thing as in `setup.py`:
https://github.com/gcovr/gcovr/blob/869bbc8d68c00499c1a374045f79f17f77192330/setup.py#L35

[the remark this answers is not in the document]
chrta: This should work. But using `./gcovr/version.py` would always report `3.4` for every version after 3.4. `git describe --tags --always` will report something like `3.4-65-g869bbc8` if you are not exactly on a tag. So you see, that it is an intermediate version.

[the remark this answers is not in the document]
latk: Thank you for everyone's suggestions!

Right now, the guide is only published manually for a release, so that the published guide always refers to the most recent released version.
In the long run, I'd like to automate this so that we can publish the docs for both the last release and the current development version (via readthedocs?). In either case, I'm not sure that tying a documentation build to a git commit is valuable.

If there's no simple way to include a version, I'd rather keep this blank for now and revisit this if and when giving a clear version would be necessary. E.g. I think readthedocs wouldn't use the project version from the sphinx config at all.

doc/source/guide.rst
```diff
@@ -205,7 +203,7 @@ This generates a HTML summary of the lines executed.  In this
 example, the file ``example1.html`` is generated, which has the
 following output:
 
-.. image:: ./screenshot-html.png
+.. image:: ./_static/screenshot-html.png
```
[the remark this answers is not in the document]
chrta: So you would like to put the images, that are generated by the makefile (and checked in), in the folder doc/source/ instead of doc/source/_static?

doc/source/index.rst
chrta: I included the changelog as it is now in the toctree. Later on it can be reformatted into proper rst.

doc/source/guide.rst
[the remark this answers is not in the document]
latk: Yes. I'd probably leave them in `doc/`. Which directory do you think would be best?

doc/source/index.rst
latk: I translated the changelog in chrta#1, you can merge my PR into your branch to include those changes as part of this PR.

doc/source/guide.rst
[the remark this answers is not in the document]
chrta: I would prefer a separate folder, only for the images. I like to have generated files that may be generated by the build system (like these images) in an extra folder, may be `doc/images` or something like this.

chrta:
Everything should be done now.

latk:
Ok, I'll do the merge over the weekend. Thank you very much for all your help with this issue :)

PR #249, opened 2018-04-06
doc: Link to github issues from changelog page 

chrta:
This is work in progress. It should not be merged yet, since it is based on #248 and contains its commit also.
Only the last commit is relevant for this.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,23 +3,23 @@
 3.4 (12 February 2018)
 ----------------------
 
- - Added --html-encoding command line option (#139).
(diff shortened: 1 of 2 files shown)


latk:
What remains to be done so that this PR loses its “WIP” status?

I can rebase it to the current master branch for you (I fiddled a bit with the commits of your other PR, so this PR doesn't _appear_ to apply cleanly – but that's just GitHub being confused).

chrta:
WIP status was just because the merge was not done yet.

Only the last commit should be necessary here.

PR #250, opened 2018-04-07
Drop support for Python 2.6

latk:
I've previously advocated for keeping 2.6-compatibility, but have since changed my mind. There are lots of small inconveniences when restricted to the 2.6/3.x common subset. In particular, this drastically restricts the choice of upstream modules.

Old versions of gcovr will of course continue to work. This also means there will never be a gcovr version that works well on Windows with Python 2.6, but since that OS doesn't bundle Python such a
version lock-in is less likely.

Now is a good time to drop support since the next release will likely be 4.0, and we already have a number of other somewhat breaking changes (e.g. different installation procedure). One biggie that is still pending is improved Windows filtering (see #191). I would not like to be locked in to Python 2.6 for the rest of gcovr 4.x.

Request for comments: Are there any problems that I missed? Will this make someone else's life much more difficult? I will wait at least 2 weeks before applying this change in order to provide enough time for feedback.

### DIFF
--- a/doc/source/guide.rst
+++ b/doc/source/guide.rst
@@ -326,10 +326,12 @@ Installation
 
 Which environments does ``gcovr`` support?
 
--   Python: 2.7+ and 3.4+.
-    Python 2.6 is supported on a best-effort basis.
+-   Python: 2.7 and 3.4+.
 
     The automated tests run on CPython 2.7, 3.4, and 3.5, and PyPy 2.7 and 3.5.
+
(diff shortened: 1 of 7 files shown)


doc/source/guide.rst
```diff
@@ -326,10 +326,12 @@ Installation
 Which environments does ``gcovr`` support?
 
 -   Python: 2.7+ and 3.4+.
```
hugovk: `2.7+` to `2.7`

PR #255, opened 2018-05-20
Filter option not affect when running with mingw-w64-x86_64-python3

lisongmin:
When I running gcovr with mingw-w64-x86_64-python3 on mingw64, the `-filter` option is not affect.

It seems mingw-w64-x86_64-python3 use '/' as `os.path.sep` by default, which
diff from other pythons on windows.

Modified tip:

1. Change hard code `"\\"` to `os.path.sep`. It is always safe and does not change the origin behavior.
2. uname on mingw64  env is `MINGW64_NT`,  change the logic to support both.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -37,6 +37,7 @@ The following developers contributed to gcovr (ordered alphabetically):
     Piotr Dziwinski,
     Reto Schneider,
     Robert Rosengren,
+    Songmin Li,
     Steven Myint,
     Sylvestre Ledru,
     trapzero,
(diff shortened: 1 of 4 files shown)


latk:
Thank you for this PR! I will merge your fix for now because it doesn't break anything, but I'm planning to rewrite the filtering system so that all filters use forward slashes on all systems (see the discussion on #191 and the warning in the [user guide](https://gcovr.com/guide.html#id8)). So I hope this fix becomes unnecessary soon.

latk:
@lisongmin I just implemented a new filtering system in #257 that uses forward slashes on _all_ platforms. Because you are familiar with filtering and are using Windows/MinGW, it would be very helpful if you could test that PR on your system. Thank you :)

PR #262, opened 2018-06-15
Add option to change threshold in HTML

Spacetown:
#261 
Also an option to change the title is added.

PR #263, opened 2018-06-15
New --html-title option to change the title of the HTML report.

Spacetown:
Split into seperate request like suggested in #262


latk:
Super, thanks a lot :)

Issue #260, opened 2018-06-13
Support setup.cfg when using gcovr in Python + C projects


stadelmanma:
First off thank you for this tool, it is a great alternative to lcov. The project I am currently using it on is a mix of Python and C extension modules. Would it be possible to support configuration via the "semi-standard" `setup.cfg` file as an alternative to just command line options? 

Secondly, this is beyond the scope of this particular issue but relevant to supporting Python C extensions better. An issue with using gcov on Python C extensions is is that I can't get the setup.py script to **not** recompile source files shared between extension modules that are already an object file. As a result that the timestamps between the `*.gcno` and `*.gcda` files can be mismatched and gcov emits a warning.  Maybe I'm missing the mark and there is a way to configure `setup.py build_ext` to do things more like `make`? (if so ignore this portion completely)

This is easily fixed using the code snippet below (where files have been globbed appropriately)
```Python
# gcda/gcno header format: [int32:magic int32:version int32:stamp]

# collect a timestamp from the first notes file to replicate
with open(gcno_files[0], 'rb') as fp:
    fp.seek(8)
    stamp = fp.read(4)

# replace stamp in header across all files to prevent errors
for file in gcno_files + gcda_files:
    with open(file, 'r+b') as fp:
        fp.seek(8)
        fp.write(stamp)
```  
However I don't know how you and your team feel about adding language specific logic into the utility. Supporting standard config files is one thing, but I completely understand the desire to let users deal with hacks like the one above.

latk:
I appreciate your kind words :)

### Configuration files

Config files are not yet implemented. They were discussed in #229 (incl. a proof of concept), but that code has not yet been merged. Config files are difficult because:

- we do not want to duplicate all the validation logic from argparse
- we have to think about default file locations and precedence of arguments
- we have to choose a config file format that supports multiple entries per key, e.g. to model the `--exclude` flag. This seems to rule out setup.cfg and any `configparser` based config files.

I really hope that proper config files will be implemented at some point, but I don't have the time to pick up the work myself. (hint hint)

As a workaround you can wrap gcovr in a shell script that adds your required options, e.g.:

``` bash
#!/bin/sh
mkdir -p coverage
exec gcovr --filter src/ --output coverage/index.html --html-details "$@"
```

### Usage with Python C extensions

I'm not sufficiently familiar with the Python toolchain to help you. The workaround you've found is terribly hacky, but if it works, it works :)

[...]

PR #265, opened 2018-06-16
Python c extension documentation

stadelmanma:
This adds some basic documentation to clarify how a user can use gcovr with Python C extensions. While not terribly difficult the documentation on the web is extremely sparse in this area and some of the semantics of the python extension build process make things tricky. Most notably the fact `setup.py build_ext` will recompile source code files even if they don't need to be when shared across multiple extension modules. 

Inside the cookbook document there are additional details on how to collect coverage data as well as two code snippets. The first provides and example of a route to build your modules in a fashion that permits code coverage data generation and the second is a python script that can be used to collect the code coverage data into the standard HTML reporting format. The python script is somewhat unneeded as gcovr managed to collect all of the proper files on it's own but works well if the user only has `gcov` available and/or wants the *.gcov files to be saved in the same tree structure as the source code.

If desired I can break those two code snippets into separate files stored in examples, that seems to be how the current examples are handled in guide.rst. 

related to issue #260

### DIFF
--- a/doc/source/cookbook.rst
+++ b/doc/source/cookbook.rst
@@ -1,141 +1,50 @@
 Gcovr Cookbook
 ==============
 
-Python with C Extensions
-------------------------
+How to collect coverage for C extensions in Python
+--------------------------------------------------
 
 Collecting code coverage data on the C code that makes up a Python
-extension module is not quite as straight forward as with a regular C
+extension module is not quite as straightforward as with a regular C
 program.
 
-There are two main steps that have to happen, first the
-`-coverage` flag must be added into the compilation process this can be
-easily accomplished by running `export CFLAGS="-coverage"` prior to
-running `python setup.py build_ext`.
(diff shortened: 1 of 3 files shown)


latk:
I'm currently working on your pull request – thank you for sharing your work! However, I'll edit it a bit to keep it more focused. I'll update this PR later so that you can review my changes.

The test failure was not a huge problem, just a small syntax error. I'll fix it. RST directives need an empty line between any `.. foo::`  directive and the directive body.

Anyway, under which name would you like to be credited in the AUTHORS file?

stadelmanma:
Hi @latk, thanks for cleaning up the RST I haven't written it in awhile and I didn't have time to get a proper dev environment up and running to test everything out. Also, thank you for cleaning up the general flow of content in the section, it reads much more clearly now. 

and `Matthew Stadelman` will work in the AUTHORS file.

PR #267, opened 2018-07-03
Fix empty or missing argument for `--exclude` and `--exclude-directores` options

mkurdej:


latk:
Thank you for the PR! I've merged the “check for empty” commit. If it's OK I'll skip the “check for missing argument” test cases because they effectively test the behaviour of argparse, not of gcovr itself.

If you enjoy writing tests, there still is [a lot to be covered](https://codecov.io/gh/gcovr/gcovr/tree/master/gcovr). E.g. a test for the --print-summary option would be very welcome :)

mkurdej:
Haha! We'll see :). Thanks for merging.

Issue #229, opened 2018-03-01
Project specific exclusion (configuration?) filter file


thejk:
When using gcovr in multiple projects and especially when used as generic targets like in meson build system I often wished I could add just one or more exclude rule. But also, would be nice to just be able to run `gcovr -r .. -o report` without having to remember exactly what excludes are useful/needed for this project.

So, I purpose to add something like a .gitignore file but for gcovr, like .gcovrignore (yes imaginative I know) where each line will act as an extra --exclude filter relative to the ignore file.
To start with perhaps just look in the root dir but would be nice to check all searched directories.

latk:
This might indeed be a useful addition – thanks for the suggestion.

If we are going to add any configuration file, it shouldn't only be possible to ignore/exclude some patterns, but also add other options (there are multiple different kinds of filters and excludes).

To implement a configuration file, we need to determine

 - how this integrates with argument parsing, to avoid duplicate validation code, and
 - whether a config file leads to any security considerations.

In the meanwhile, invoking gcovr through a shell script etc. would allow you to specify default arguments. E.g. as a Posix-Shell script:

``` sh
#!/bin/bash
exec gcovr --some default --args "$@"
```

thejk:
Trouble I found when doing a quick proof of concept implementation (https://github.com/thejk/gcovr/tree/ignore) is finding the files and in what directories to expect them.

If we concentrate just on the different filters:
* --filter and --exclude
* --gcov-filter and -gcov-exclude
* --exclude-directories

When searching for gcov/gcda/gcno files --exclude-directories is used.
When processing the gcov files --gcov-filter and --gcov-exclude are used.
When finding sources from the gcov files --filter and --exclude are used.
(Please correct me if I'm wrong, but to the best of my understanding that's the gist of it)

I mostly use meson for new projects and have old projects using autofoo, meson enforces that you use a separate build directory and autofoo these days encourage it.
So, I have my build directory which is not commited and rather temporary,
in the build directory all my gcda, gcno (or gcov) files are so I run gcovr with --root=build, it will find the gcda files which will have source paths like ../src/foo.c.

The trouble I have here is figuring out where a good place for the ignore files would be.
For --exclude/--filter, ie for source files, the answer is easy enough, put the file in .. or ../src (relative --root) and we can search for the ignore rules while walking the tree to find the source files. This is what my POC branch does.
[...]

PR #279, opened 2018-09-09
Move argument validation outside of argparse

stadelmanma:
Extracts the command line api configuration out into a separate module (configuration.py). The `GcovrConfigOption` classes store all of the necessary information to add themselves to an argparse parser and make that information easily accessible for any future configuration schemes. (The objects produced by `add_argument` are rather opaque.) 

Todo:
 * [ ] Add unit tests (integration tests are probably already covered)

This is a building block for eventually addressing #229

### DIFF
--- a/gcovr/configuration.py
+++ b/gcovr/configuration.py
@@ -131,8 +131,11 @@ class GcovrConfigOption(object):
         self.help = help.format(**self.__dict__)
 
     def __repr__(self):
-        r"""String representation of instance."""
+        r"""String representation of instance.
 
+        >>> GcovrConfigOption('foo', ['-f', '--foo'], help="fooify")
+        GcovrConfigOption('foo', [-f, --foo], ..., help='fooify', ...)
+        """
         kwargs = [
             '{k}={v!r}'.format(k=k, v=v)
             for k, v in sorted(self.__dict__.items())
--- a/gcovr/configuration.py
+++ b/gcovr/configuration.py
@@ -1,0 +1,471 @@
+# -*- coding:utf-8 -*-
+
+# This file is part of gcovr <http://gcovr.com/>.
+#
+# This module manages processing and validation of the configuration options
+# passed into gcovr.
+#
+#
+# Copyright 2013-2018 the gcovr authors
+# Copyright 2013 Sandia Corporation
+# This software is distributed under the BSD license.
+from argparse import ArgumentTypeError, SUPPRESS
+from locale import getpreferredencoding
+from multiprocessing import cpu_count
+import os
+
+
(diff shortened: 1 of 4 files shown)


gcovr/configuration.py
```diff
@@ -0,0 +1,475 @@
...
+
+    This class is initializer in a very similar fashion as adding arguments
+    to an argparse Parser instance with a few minor changes.
+        1. The first positional argument is the "name" (i.e. "dest" in
+           argparse speak) and is required.
+        2. A "group" keyword argument has been added which corresponds to a
+           key in the GCOVR_CONFIG_OPTION_GROUPS dictionary.
+    """
+
+    def __init__(self, name, *flags, **kwargs):
```
latk: Do we have to use `**kwargs` here, or can we enumerate all supported option-fields? Due to Python-2 restrictions we can't really combine `*flags` varargs with explicit keyword-only arguments, so we'd have to pass in an explicit list/tuple of flags. Something like:

``` python
def __init__(self, name, flags=(), group=None, help=None, action=None):

...
GcovrConfigOption('foo', ['--foo'], help="frobnicate something", action='store_true')
```

… and so on, where everything after `flags` is intended to be used as a keyword-argument. This would allow us to get rid of some `hasattr()/getattr()` calls later, and clarifies the intended API.

gcovr/configuration.py
```diff
@@ -0,0 +1,475 @@
...
+            self.const = False
+            self.default = True
+
+    def __repr__(self):
+        r"""String representation of instance."""
+        fmt = "GcovrConfigOption({name!r}, {flags}group={group!r}, {kwargs})"
+        kwargs = dict(self.__dict__)
+        for key in ["name", "flags", "group"]:
+            del kwargs[key]
+        kwargs = ['{k}={v!r}'.format(k=k, v=v) for k, v in kwargs.items()]
```
latk: dict iteration order is undefined prior to Python 3.6. To get a stable `repr()` across runs we should sort the entries, e.g. `[... for k, v in sorted(kwargs.items())]`.

gcovr/configuration.py
```diff
@@ -0,0 +1,475 @@
...
+
+        # we only want to set dest for non-positionals
+        if self.flags:
+            kwargs["dest"] = self.name
+            parser.add_argument(*self.flags, **kwargs)
+        else:
+            parser.add_argument(self.name, **kwargs)
+
+
+GCOVR_CONFIG_OPTION_GROUPS = {
```
latk: dict order is undefined prior to Python 3.6. We can use `collections.OrderedDict` here so that the groups are always in the correct order

PR #284, opened 2018-10-09
search_file: don't visit the same directory more than once

wjt:
The directory tree being scanned may in fact be a directed graph, with
cycles, due to following symbolic links. The kernel imposes a limit on
the number of symbolic links followed in a path, beyond which operations
fail with ELOOP. (According to path_resolution(7), the limit is 40 links
on current Linux.) If the graph is very branchy, it may take a very long
time to fail.

For a real-world example, the Meson build system uses gcovr to provide a
'ninja coverage-xml' target. The systemd project (which uses Meson) has
a test suite which generates a fake sysfs tree, with many circular
references, including 64 self-referential ttyX directories:

    $ ls -l build/test/sys/devices/virtual/tty/tty1/subsystem/tty1
    lrwxrwxrwx 1 wjt wjt 30 Oct  9 12:16 build/test/sys/devices/virtual/tty/tty1/subsystem/tty1 -> ../../devices/virtual/tty/tty1
    $ ls -l build/test/sys/devices/virtual/tty/tty1
    total 12
    -rw-r--r-- 1 wjt wjt    4 Oct  9 12:16 dev
    drwxr-xr-x 2 wjt wjt 4096 Oct  9 12:16 power
    lrwxrwxrwx 1 wjt wjt   21 Oct  9 12:16 subsystem -> ../../../../class/tty
    -rw-r--r-- 1 wjt wjt   16 Oct  9 12:16 uevent
    $ ls -l build/test/sys/devices/virtual/tty/tty1/subsystem/tty1
    lrwxrwxrwx 1 wjt wjt 30 Oct  9 12:16 build/test/sys/devices/virtual/tty/tty1/subsystem/tty1 -> ../../devices/virtual/tty/tty1
    $ readlink -f build/test/sys/devices/virtual/tty/tty1/subsystem/tty1
    /home/wjt/src/endlessm/systemd/build/test/sys/devices/virtual/tty/tty1

Note that each link refers to a directory which contains all the fake
tty trees again. 64 ** 40 is a very large number! On our build server, I
cancelled the build after 16 hours.

Even in less pathological cases, since search_file() only yields
absolute paths, there's no need to visit a file via more than one path.
Maintain a set of (device, inode) pairs for each directory visited, and
skip processing any directory that's been visited before.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -44,5 +44,6 @@ The following developers contributed to gcovr (ordered alphabetically):
     Steven Myint,
     Sylvestre Ledru,
     trapzero,
+    Will Thompson,
     William Hart,
     and possibly others.
--- a/gcovr/tests/linked/Makefile
+++ b/gcovr/tests/linked/Makefile
@@ -39,5 +39,6 @@ links:
 	   ln -s ../../../nested/subdir/A n;\
 	   cd ..;\
 	   ln -s m/n A;\
+	   ln -s . loop;\
 	   cd ..;\
 	fi
--- a/gcovr/utils.py
+++ b/gcovr/utils.py
@@ -7,8 +7,27 @@
 # This software is distributed under the BSD license.
 
 import os
+import platform
 import re
 import sys
+
+if not (platform.system() == 'Windows' and sys.version_info[0] == 2):
+    class LoopChecker(object):
+        def __init__(self):
+            self._seen = set()
+
+        def already_visited(self, path):
+            st = os.stat(path)
+            key = (st.st_dev, st.st_ino)
+            if key in self._seen:
+                return True
+
+            self._seen.add(key)
+            return False
+else:
+    class LoopChecker(object):
+        def already_visited(self, path):
+            return False
 
 
 def search_file(predicate, path, exclude_dirs):
@@ -21,7 +40,13 @@ def search_file(predicate, path, exclude_dirs):
     elif not os.path.exists(path):
         raise IOError("Unknown directory '" + path + "'")
 
+    loop_checker = LoopChecker()
     for root, dirs, files in os.walk(os.path.abspath(path), followlinks=True):
+        # Check if we've already visited 'root' through the magic of symlinks
+        if loop_checker.already_visited(root):
+            dirs[:] = []
+            continue
+
         dirs[:] = [d for d in dirs
                    if not any(exc.match(os.path.join(root, d))
                               for exc in exclude_dirs)]


latk:
This looks like a very sensible improvement! Thank you for discovering the problem and finding a solution :)

Unfortunately, this PR seems to break many tests under Windows, but I don't understand why. Maybe a less fragile solution would be to store the abspath in the set? If absolutely necessary, it would also be acceptable if your check gets disabled under Windows.

For the test that you modified to create many links, it seems that this isn't supported under Windows/MinGW either:

    ln: failed to create symbolic link 'loop-1' -> '.': File name too long

It might be necessary to add this test case to the list of xfail tests in `gcovr/tests/test_gcovr.py`:

``` diff
     is_windows = platform.system() == 'Windows'
     needs_symlinks = any([
-        name == 'linked' and format == 'html',
+        name == 'linked',
         name == 'filter-relative-lib',
 ])
```

wjt:
> Unfortunately, this PR seems to break many tests under Windows, but I don't understand why. Maybe a less fragile solution would be to store the abspath in the set? If absolutely necessary, it would also be acceptable if your check gets disabled under Windows.

The failures are weird… and I don't quite understand why the test fails under Python 2.7 on 32-bit Windows, but passes on 64-bit Windows with Python 3.7. Perhaps the files are being discovered in a different order because of the self-referential symlink? I'm afraid I don't have a Windows development system on hand to debug interactively. I'll try storing the abspath instead.

latk:
Oh, I missed that the tests passed under Python 3.7. This might be one of those cases where standard library support has drastically improved over time. Note that both tests use the same Windows image and MinGW version, merely the Python version is different.

If the failure is specific to 2.7, let's just disable your improvements for Python 2.x under Windows, without a fallback. After all, the code currently works fine (except in extreme cases like those you encountered). Once Python 2.7 hits its EOL in 2020 that check can be removed again.

wjt:
Are you suggesting making the fix itself conditional on the Python version, or skipping the test on 2.x? I think the former, but just want to check.

wjt:
Pushed (and squashed after the tests passed) a new version that disables the new logic under (Windows && Python 2), and factored it out to a small helper class, which I think makes both the platform-conditional behaviour and the main function easier to read.

wjt:
A fun problem I just noticed is that pytest's test collector has the same bug; if you run the test suite twice without manually cleaning out the symlink farm, then scanning for tests the second time takes forever. https://github.com/pytest-dev/pytest/issues/624

Sigh. I can just amend the Makefile to only create one cycle. This still makes test collection take longer the second and subsequent times, but at least it's measured in milliseconds! On my system, with just one link cycle, runtime of the `linked` tests (including longer collection time) across 5 runs is 2.597+/-0.3141 seconds with the fix, and 2.985+/-0.3226 seconds without it.

Issue #300, opened 2019-03-19
OOS builds and relative exclude paths


kjerstadius:
I'm using the current tip of master (c018e2b) and what I'm trying to do is to define some excludes using a gcovr.cfg file. When combining this with an OOS build (in my case Meson, but the same principle would apply for CMake too) I find that I'm unable to define exclude path(s) in a satisfactory manner. If I'm not mistaken, the problem I see arises because of the fact that relative filters are always generated relative to the current working directory, which in an OOS build will never be the same directory as the sources are in (except for generated sources, which may complicate things further). This means that in order to exclude a source file from the report, you either need to know the absolute path to the file, or the relative path to the file _from the build directory_. Since neither of these are known such that they can be defined once, the prospect of having a gcovr.cfg file with excludes defined sort of falls apart.

The way I see it there are a couple of different options that could help solving this issue (in no particular order):
- Allow relative filters to be built relative to the source root by passing an additional (optional) argument to build_filter().
- Allow gcovr.cfg files to be located in the build root.
- Add some kind of expression or syntax to tell gcovr that an exclude is or isn't relative to the source root.

I already tried the first approach by making some minimal changes on my fork ([see here](https://github.com/kjerstadius/gcovr/commit/ee966af7ec2bc0b6ded2f35e3dbf93056374c102), I've not yet made a PR of it since I wanted to discuss first). It works for my simple test and passed all CI tests, but I'm not sure whether there may be unknown (to me) side effects, or if other filters should be given the same treatment. There may also be problems in the case of generated sources, in which case the source code will be in the build directory, the name and location of which is unknown at the time of writing the cfg file, so that sort of puts us back at square one.

I don't really like the third option, but I thought I'd throw it out there. The second one I find interesting though. Since a staple of modern build systems in the concept of configuring files to generate sources based on the current environment, couldn't it make sense to support that for gcovr.cfg in OOS builds too? What I'd propose is for the logic in find_config_name() to change slightly so that if the source root and the current working directory are different, the current working dir (assumed to be the build root in OOS build) is searched first and if there's no file there, the source root is checked. This way the filter generation would remain the same, and compatibility with generated sources should be maintained.

What do you think?

latk:
[...]

PR #304, opened 2019-04-09
Add path context to filter options in config files

kjerstadius:
Addresses #300. From the "main" commit message:

> By using a dedicated type for filter options it's possible include the
> path context for a filter. This enables (relative) filters in config
> files to be treated as relative to the location of the config file,
> rather than relative to the current working directory.

Handling and behavior of CLI arguments is the same as before. Do note that the `check_non_empty()` function ended up being unused, so I removed it (in a separate commit though, so easy to roll back).

### DIFF
--- a/gcovr/configuration.py
+++ b/gcovr/configuration.py
@@ -16,6 +16,8 @@ import os
 import re
 import sys
 
+from .utils import FilterOption
+
 try:
     from typing import Iterable, Any
 except ImportError:
@@ -290,7 +292,11 @@ def _get_value_from_config_entry(cfg_entry, option):
         value = cfg_entry.value_as_bool
     elif option.type is not None:
         try:
-            value = option.type(cfg_entry.value)
+            if option.type is FilterOption:
+                value = option.type(cfg_entry.value,
+                                    os.path.dirname(cfg_entry.filename))
+            else:
+                value = option.type(cfg_entry.value)
         except (ValueError, ArgumentTypeError) as err:
             raise cfg_entry.error(str(err))
     else:
@@ -549,6 +555,7 @@ GCOVR_CONFIG_OPTIONS = [
              "Can be specified multiple times. "
              "If no filters are provided, defaults to --root.",
         action="append",
+        type=FilterOption,
         default=[],
     ),
     GcovrConfigOption(
@@ -557,7 +564,7 @@ GCOVR_CONFIG_OPTIONS = [
         help="Exclude source files that match this filter. "
              "Can be specified multiple times.",
         action="append",
-        type=check_non_empty,
+        type=FilterOption,
         default=[],
     ),
     GcovrConfigOption(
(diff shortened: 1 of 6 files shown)


gcovr/configuration.py
```diff
@@ -566,6 +567,7 @@ def merge_options_and_set_defaults(partial_namespaces, all_options=None):
         help="Keep only gcov data files that match this filter. "
              "Can be specified multiple times.",
         action="append",
+        type=FilterOption,
```
kjerstadius: This option (and `gcov_excludes` as well) previously didn't use the `check_non_empty()` function. I wasn't sure if that was deliberate or not, but in the current iteration of the code that will be a slight behavioral change. I had to modify a test case to deal with this fact.

gcovr/tests/test_args.py
```diff
@@ -103,7 +103,7 @@ def test_line_threshold_100_1(capsys):
 def test_filter_backslashes_are_detected(capsys):
     c = capture(
         capsys,
-        args=['--filter', r'C:\\foo\moo', '--gcov-exclude', ''],
```
kjerstadius: Couldn't really figure out the necessity of this argument as it seems to have no effect or relation at all to the actual test.

gcovr/utils.py
```diff
@@ -139,22 +140,31 @@ def calculate_coverage(covered, total, nan_value=0.0):
...
-        logger.warn("did you mean: {}", suggestion)
-
-    if os.path.isabs(regex):
-        return AbsoluteFilter(regex)
-    else:
-        return RelativeFilter(os.getcwd(), regex)
+class FilterOption(object):
+    def __init__(self, regex, path_context=None):
+        if not regex:
+            raise ArgumentTypeError("filter cannot be empty")
```
kjerstadius: Not sure if `ArgumentTypeError` is the most appropriate exception to use here, but I'd say it's the most straightforward to get behavior that's as similar as possible to before. I evaluated other alternatives but none really stuck. `ValueError` could be used, but using it will print a standardized error rather than a custom one. It will also print a traceback which the Argparse exception handling won't. If the behavior of the Argparse exception is to be preferred (mainly no traceback) a custom program exit function would need to be written. It would of course also be possible to leverage `check_non_empty()` to check that the value is set before passing it to the class constructor similar to how it was before, but then it would only apply to CLI args and not config file filters.

gcovr/configuration.py
latk: I'll have to think a bit more about this.

An empty filter would be a regex that always matches, and can be used to deactivate a filter. This is most important for the `--filter` option to deactivate filtering (the default would filter to files under `--root`).

But I don't think this would be useful for `--gcov-filter` or `--gcov-exclude`: it doesn't make sense to exclude all files, and accepting all files already is the default. That only leaves use cases like “deactivating filters in a configuration file” which I'm not interested in supporting at the moment.

gcovr/tests/test_args.py
latk: I now understand why `--gcov-exclude` was used here. The `capture()` test driver runs gcovr's main(). It actually searches for coverage data and creates a report. But this test is not about the reports, only about side effects. Setting `--gcov-exclude` to an empty filter ensures that no coverage data will be found, which should make the test more robust. The empty filter is the easiest way to write a filter that always matches, here: excludes all files.

Issue #1, opened 2013-09-02
Memory Usage for XML output is very high


whart222:
For my project (Gromacs) the memory usage of gcovr is extremely high when using the XML output (-x). It tops at about 1GB. This is with the version from https://software.sandia.gov/trac/fast/export/2745/gcovr/trunk/scripts/gcovr 

(from ticket 3922)

TiloW:
I'm facing a similar issue with gcovr 3.2 using too much memory for XML output.
This is especially problematic for me because it increases build time due to swapping.

I've compared the three generators on my local machine (which has sufficient memory).

|         | *.gcov | *.gcno |
|---------|--------|--------|
| Number of Files    | 388 | 612 |
| Size of Files [MB] |  14 | 141 |

|         | Plain Text | HTML | XML |
|---------|------------|------|-----|
| Peak Memory [KB] | 94280 | 97076 | 2505896 |
| Runtime         | 2.70  | 2.01 | 3.74 |

As you can see, runtime is generally not a problem but memory usage is.
XML output uses 25 times the memory of any other generator.

Is there a special reason for using `MiniDom` instead of `ElementTree` or `lxml`?

PR #307, opened 2019-05-05
Address Memory usage

jfixemer:
RE: https://github.com/gcovr/gcovr/issues/1

As suggested somewhere (regarding gcovr on the Linux kernel), detecting and using the lxml etree instead of pure python etree to bring down memory usage. 

We were trying to use this on our private codebase in a size restricted container and it failed due to memory. With this change it works  (memory saving is at least 3x).  

lxml may also be faster (possibly partially addressing the speed issue, although it is still DOM, not STREAM as suggested so still doesn't fully address that issue)

### DIFF
--- a/gcovr/cobertura_xml_generator.py
+++ b/gcovr/cobertura_xml_generator.py
@@ -217,11 +217,12 @@ def print_xml_report(covdata, options):
 
     with smart_open(options.output) as fh:
         if LXML_AVAILABLE:
-            fh.write(
+            print(
                 etree.tostring(root,
                                pretty_print=options.prettyxml,
                                encoding="UTF-8",
                                xml_declaration=True,
-                               doctype="<!DOCTYPE coverage SYSTEM 'http://cobertura.sourceforge.net/xml/coverage-04.dtd'>"))
+                               doctype="<!DOCTYPE coverage SYSTEM 'http://cobertura.sourceforge.net/xml/coverage-04.dtd'>"),
+                file=fh)
         else:
-            fh.write(etree.tostring(root, encoding="UTF-8"))
+            print(etree.tostring(root, encoding="UTF-8"), file=fh)
--- a/gcovr/cobertura_xml_generator.py
+++ b/gcovr/cobertura_xml_generator.py
@@ -31,11 +31,11 @@ def smart_open(filename=None):
     if filename and filename != '-':
         # files in write binary mode for UTF-8
         fh = open(filename, 'wb')
-    elif (sys.version_info > (3,0)):
-        #python 3 wants stdout.buffer for binary output
+    elif (sys.version_info > (3, 0)):
+        # python 3 wants stdout.buffer for binary output
         fh = sys.stdout.buffer
     else:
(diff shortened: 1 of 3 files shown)


jfixemer:
failed CI

gcovr/cobertura_xml_generator.py
```diff
@@ -174,47 +197,31 @@ def print_xml_report(covdata, options):
...
+    with smart_open(options.output) as fh:
+        if LXML_AVAILABLE:
+            print >>fh, \
+                etree.tostring(root,
+                               pretty_print=options.prettyxml,
+                               encoding="UTF-8",
+                               xml_declaration=True,
+                               doctype="<!DOCTYPE coverage SYSTEM 'http://cobertura.sourceforge.net/xml/coverage-04.dtd'>")
+        else:
+            print >>fh, etree.tostring(root, encoding="UTF-8")
```
latk: should use `fh.write(...)` or the print function `print(..., file=fh)`. Print statements only work in Python 2.

latk:
Thank you for working on this long-standing issue! I'll be happy to merge once the tests pass again. Please also add your name to the AUTHORS file.

jfixemer:
making it worse, need better local test env. sorry tried to go cheap.

latk:
No problem. It is fine to leave this issue open while you work on it.

Gcovr is primarily oriented towards Python 3 now, but trying to write code
that works under both 2 and 3 can be tricky. Here, the problem seems to be
related to the str/unicode → bytes/str change. Most file operations work on
str by default, but that means different things in the Python versions.

I'm not sufficiently familiar with the lxml  API. Maybe you just need to
avoid manually encoding the XML, or maybe you need to encode it
conditionally on the Python version?

jfixemer:
We use still use python2 where I need this to work. So testing python3 is extra and difficult (why I was going cheap and trying to use the project CI)

Probably going to need a little version detection...

Just ignore me until it works ;-)

jfixemer:
I'm not good with travis-ci / appveyor, I'm not sure if I should change one of them or how to add a matrix that does 'pip install lxml' (or if there is additional work to install lxml libraries) that would restore the lost coverage and exercise the real enhancement...

latk:
The dependency should be added to the setup.py file, this will also cause
it to be installed during all  CI tests.

However, I don't think lxml works with pypy. It may be necessary to add the
dependency as so:

    install_requires = [
      ...,
      'lxml ; implementation_name == "cpython"',
    ]

Such environment markers are specified in PEP 508.

jfixemer:
But it should work both ways -- wasn't that a comment from the earlier attempt?
So running regression with or without lxml only tests half the supported mechanism.

Well, I don't know if pypy skipped lxml all together, but it didn't fail.  I see the coverage in cobertura-xml change before and after lxml is installed so I think it hits different branches.  Not sure if this reflects well in the coverage, or if only one run is accounted for, etc.

I guess I can try some more stuff if you really think it needs to be better... (but I kind of think it is good enough).

PR #308, opened 2019-05-06
Add XML generator for sonarqube generic coverage report

akmll:
Add command line option '--sonarqube' with output report file name to generate sonarqube generic coverage report as described in https://docs.sonarqube.org/latest/analysis/generic-test/

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -26,6 +26,7 @@ The following developers contributed to gcovr (ordered alphabetically):
     libPhipp,
     Lukas Atkinson,
     Luke Woydziak,
+    Leon Ma,
     Marek Kurdej,
     Martin Mraz,
     Matsumoto Taichi,
--- a/README.rst
+++ b/README.rst
@@ -123,7 +123,7 @@ To generate code coverage report for sonarqube:
 
 ::
 
-    gcovr -r . --xml-pretty --sonarqube coverage.xml
+    gcovr -r . --sonarqube coverage.xml
 
 This will create coverage.xml containing the sonarqube generic coverage report described at:
 `<https://docs.sonarqube.org/latest/analysis/generic-test/>`_
--- a/README.rst
+++ b/README.rst
@@ -119,6 +119,15 @@ You can also generate detailed HTML reports:
 
 Gcovr will create one HTML report per source file next to the coverage.html summary.
 
+To generate code coverage report for sonarqube:
+
+::
+
+    gcovr -r . --xml-pretty --sonarqube coverage.xml
+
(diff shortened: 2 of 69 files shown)


latk:
Thank you for this PR, that looks fantastic. It is fine to leave the PR open while you work on it. The Travis test failures seem to only be about code style issues (unused variables).

What I need from you before I can merge:

* Please add your name to the AUTHORS file
* Please add a mention of this format to the README file.
* The recent pull request #307 moves to the lxml backend which reduces memory usage. I think this would be a good idea here as well. If you run into any Python 2 vs 3 problems, it's fine if new features only work under 3.
* I need at least one test case for the generator. This involves:
  1. creating a new target in the test's Makefile, 
  2. running it manually to create the output file
  3. manually verifying the output
  4. moving the file into the test's reference folder
  5. extend the gcovr/tests/test_gcovr.py test runner to run this additional target if it is available

Please let me know if you run into any problems.

akmll:
Thank you for your comments. I am quite a newbie in python, I will update the unit tests and create pull request later. It makes no sense to add uncovered code in a code coverage tool.

akmll:
Sorry, I didn't notice it was reopened. I am quite a newbie in python so this might take time, and I use msys2 on windows, it seems some of the test cases would fail.

latk:
No worries. I'm thankful for any work that you can do!

The tests assume a specific GCC version (GCC 5) and may fail under
different versions. This is not a problem, as long as Travis/Appveyor are
green. If extending the tests is too tricky then I can look into it myself.
TBH this is the first time in forever that a new test target is added, so
this would be nontrivial for me as well.

The part about tests that I definitely need from you is that you confirm
that the generated XML is accepted by Sonarqube, for example by using your
modified version of gcovr in your build pipeline. If that seems to work,
I'm confident it will be all right for others as well.

Again, thank you very much for working on this, and please reach out if I
can help.

akmll:
Finally I have made the unit tests to success. The uncovered lines of code share across the generators, so I think it is OK.

I have tested the generator with a real project under enterprise environment and it works like a charm, the coverage information is just the same as the html report generated.

I read in the PR #307 that the memory usage is high with xml.minidom. I was thinking about generating the xml text directly since the format is rather simple. I am not very familiar with python, so I am not sure if that would help.

Another thing needs to mention is that there is no stdout print option for this sonarqube generator, I could add it easily if there is such scenario.

latk:
OK, I've merged this manually! I decided not to use the string-based output generation but rather ported to lxml. Your idea to pass the output file name via `--sonarqube` rather than `-o` was good, so I added that to all other formats as well.

Thanks again for working on this!

PR #315, opened 2019-05-31
Fixed issue under gcov 8 where an un-hit section could overwrite a hit in templated code

obiphi:
Hi,

When working with GCC 8 (ref #228) I found a bug whereby if the gcov file is arranged thus:
```
------------------
Foo<int>::Foo():
function Foo<int>::Foo() called 1 returned 100% blocks executed 100%
        1:    7:  Foo(): b (1000) {}
------------------
Foo<char>::Foo():
function Foo<char>::Foo() called 0 returned 0% blocks executed 0%
    #####:    7:  Foo(): b (1000) {}
------------------
```
instead of:
```
------------------
Foo<char>::Foo():
function Foo<char>::Foo() called 0 returned 0% blocks executed 0%
    #####:    7:  Foo(): b (1000) {}
------------------
Foo<int>::Foo():
function Foo<int>::Foo() called 1 returned 100% blocks executed 100%
        1:    7:  Foo(): b (1000) {}
------------------
```
then the line in question is flagged as un-hit, because the un-hit indicator `####` occurs after the hit indicator. In addition, if there are multiple instantiations of the template with different types only the hit count for the last one is used. 

I've modified an existing test to demonstrate & test the first issue, and validated that the change fixes the issue I see in my code as well. I have done my best to understand and follow the contributions rules and have run the build under Travis CI to verify the changes I've made. 

Please let me know if there are any other changes I can make to this PR to get it accepted.

Thanks,
Phil

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -40,6 +40,7 @@ The following developers contributed to gcovr (ordered alphabetically):
     Mikk Leini,
     Nikolaj Schumacher,
     Piotr Dziwinski,
+    Phil Clapham,
     Richard Kjerstadius,
     Reto Schneider,
     Robert Rosengren,
--- a/gcovr/gcov.py
+++ b/gcovr/gcov.py
@@ -308,13 +308,13 @@ class GcovParser(object):
             if is_non_code(code):
                 self.coverage.line(self.lineno).noncode = True
             else:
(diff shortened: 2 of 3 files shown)


latk:
This looks excellent! Thank you for the fix :)

Issue #282, opened 2018-10-03
Support GCOV intermediate format


marxin:
For GCC 9.1 I'm planning to come up with JSON format of the intermediate representation.
It's feasible for consumers of the information like lcov and gcovr
Feel free to comment my patch request:
https://gcc.gnu.org/ml/gcc-patches/2018-09/msg01628.html

Example output:
https://users.suse.com/~mliska/tmp/tramp.json

marxin:
> Thank you, this looks great! There's nothing wrong with the previous intermediate format, but moving to JSON will make future improvements easier (both for GCC and downstream users like gcovr). I look forward to implementing support for it, because it's much easier to work with than the human-readable output we are currently parsing….

Thanks for very useful feedback!

> 
> Unfortunately I won't be able to _move_ to your JSON format because old GCC versions and `llvm-cov gcov` will have to be supported, but it will be great to be able to offer a more robust parser for modern GCC as an option.

I understand that, but hopefully in the future it will be main format used for GCC.

> 
> A few concerns/questions/observations:
> 
>  It is great that you are including a `current_working_directory` key! Are the file names resolved relative to this working directory, or are the file names taken directly from the gcno notes? This matters if gcov is not invoked from the same directory as gcc, which happens to be one of the difficult problems for gcovr.

current_working_directory is taken from .gcno file and file names as well. That should be enough to locate source files.

> 
>  These JSON files can get quite big, but gcov and any parsers will have to load them into memory entirely. Please consider whether a mixed line-based/JSON format would work, i.e. something like [JsonLines.org](http://jsonlines.org/) with one "file" JSON document per line and perhaps a header with metadata. That's technically not true JSON but much more Unixy. Line-based output would be more valuable than compressed output to me.

Well you will have one JSON per object file. Even if you have very many object files, then it's expected that one can load into memory JSON for one of them. Then you can release it.

> 
>  Will it be possible to pipe the JSON to stdout instead of writing to a file? It seems so. Great!

Yes.

> 
>  No format is self-describing. The source code in your patch makes the format clear, but real documentation would avoid misunderstandings. E.g. the intermediate format shows a vaguely BNF-ish example in the docs. I'd consider writing:
> 
> > The JSON output is structured as follows:
> > ```json
> > {
> >   "version": "<gcov_version>",
> >   "current_working_directory": "<cwd_absolute>",
> >   "files": [<file>...]
> > }
> > ```
> > 
> > 
> > where each `<file>` has the form:
[...]

Issue #10, opened 2013-09-02
add "--add-tracefile" like lcov


whart222:
I want to merge two gcda result into one.
lcov can use --add-tracefile do the work, how gcovr do it? Thanks

wichtounet:
Such a feature would indeed be great. I also have the need for it with several object directories.

verbitan:
This is exactly the feature I'd like as well, so a vote for this issue from me :+1:

pbregener:
I'm interested in this feature as well 👍

pbregener:
@whart222 As you opened this issue 3 years ago and now seem to be the maintainer of gcovr - any thoughts on this?

pbregener:
ping @whart222

latk:
While working on a new internal data model (191b30cd3d7a0377341dcfc42b8ea24681614dce) I realized that merging coverage data is not actually difficult – such merging is already done internally to combine the coverage from multiple object files. All that is missing is to serialize and deserialize that data. In particular:

 - a new generator that dumps the complete coverage data into a file, preferably as JSON.
 - an option to load such a files so that additional coverage can be merged. “Steal” the design of that option from lcov?

This file format should allow some sanity checks, for example that all runs must use the same options.

The file format could be modeled after the upcoming gcov json format (see #282) which would allow for some interesting synergies, but this is not necessary.

If anyone would like to take on this work, please do :) As per the [contribution guidelines](https://github.com/gcovr/gcovr/blob/master/CONTRIBUTING.rst#how-to-submit-a-pull-request): if you need help please ask in the [Gitter chat](https://gitter.im/gcovr/gcovr), discuss the design on this issue, or open a WIP pull request.

CezaryGapinski:
Hi!

I am trying to make add-tracefile option with JSON files. I have noticed that something can be wrong with `LineCoverage` class [constructor](https://github.com/gcovr/gcovr/blob/1f5af01d1a347c917939df34082254544d85a5f1/gcovr/coverage.py#L68) and [update method](https://github.com/gcovr/gcovr/blob/1f5af01d1a347c917939df34082254544d85a5f1/gcovr/coverage.py#L106).
Maybe I missed something, but `noncode` fileld should be set to True as a default value?
During update when one of the object has `noncode` field cleared and the other has it set that means the line is still real code, filed shouldn't be set and line should be taken into coverage?

latk:
[...]

PR #326, opened 2019-08-17
Add --add-tracefile feature to merge coverage data in JSON format

CezaryGapinski:
This PR implements merging coverage data from files in JSON format.

PR helps to merge results, simillary to other tools like lcov. Files to combine are in JSON format recommended by @latk in #10, created partially by @opikalo in PR #302 and use update methods in coverage data model. The file structure is based on JSON intermediate format implemented in gcov mentioned in #282 with structure and additional keys specific to gcovr.
gcovr always create reports with source files paths relative to root directory, therefore JSON files structure relative to root. It also helps to combine results from different builds on separate computers or different directories with the same source code directory structure.


gcovr/tests/add_coverages/Makefile
```diff
@@ -0,0 +1,40 @@
...
+json_foo:
+	./testcase_foo
+	$(GCOVR) -d --json-pretty -o coverage_foo.json
+
+json_bar:
+	./testcase_bar
+	$(GCOVR) -d --json-pretty -o coverage_bar.json
+
+txt: json_foo json_bar
+	$(GCOVR) -a coverage_foo.json -a coverage_bar.json -o coverage.txt
```
latk: I really love this test! It's a great demonstration of the --add-tracefile functionality.

In the future, I hope to port the reference data of the other tests to JSON so that I don't have to check the HTML and XML data every time. This makes the following things possible or a lot easier:

* faster test suite
* smaller test suite
* changes to the HTML format will become easier because fewer files will have to be updated
* supporting multiple compilers in the test suite instead of just GCC 5
* the XML tests will become less important, so I can probably remove the pyutilib dependency \o/

So thank you a lot for working on this cornerstone feature!

gcovr/tests/add_coverages/reference/coverage_bar.json
```diff
@@ -0,0 +1,167 @@
+{
```
latk: Do the tests have to include the individual files like `coverage_{foo,bar}.json` or is it sufficient to only compare the final coverage.json?

gcovr/tests/add_coverages/Makefile
```diff
@@ -0,0 +1,40 @@
...
+	$(GCOVR) -a coverage_foo.json -a coverage_bar.json -x -o coverage.xml
+
+html: json_foo json_bar
+	$(GCOVR) -a coverage_foo.json -a coverage_bar.json --html-details -o coverage.html
+
+sonarqube: json_foo json_bar
+	$(GCOVR) -a coverage_foo.json -a coverage_bar.json --sonarqube sonarqube.xml
+
+json: json_foo json_bar
+	$(GCOVR) -a coverage_foo.json -a coverage_bar.json --json-pretty -o coverage.json
```
latk: The `json` make target will only be invoked if you add it to the list of targets in `gcovr/tests/test_gcovr.py`. The `run` target is not actually used by the test suite. Additionally, the test driver needs to be updated with `json=lambda x: x` in SCRUBBERS and `json='coverage*.json'` in OUTPUT_PATTERN.

CezaryGapinski: Ok, so I have to extend this line and add 'json':
`@pytest.mark.parametrize('format', ['txt', 'xml', 'html', 'sonarqube', 'json'])`
and next:

- extend all tests Makefiles to use json rule
- extend all reference directories with json coverage?

latk: I think the tests will fail if you don't update the other Makefiles to include a `json` target. But to keep this PR reasonably small, you don't have to create JSON reference files. The JSON target can be empty.

gcovr/configuration.py
```diff
@@ -421,6 +421,16 @@ def choose(_cls, choices, default=None):
              "The --root is the default --filter.",
         default='.',
     ),
+    GcovrConfigOption(
+        "add_tracefile", ["-a", "--add-tracefile"],
+        help="Combine the coverage data from JSON files. "
```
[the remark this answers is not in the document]
CezaryGapinski: > Perhaps the main() should have additional checks so that it can warn when unused options like filters were specified on the command line (but not when they are just loaded from the config file).

OK, just mention about that no coverage data will be collected from gcov as it is already done for other options like - - branches, - - sort-uncovered, - -sort-percentage or do something more sophisticated in the main()? What with other options like for example calling the gcovr to XML output with unused the - - sort-uncovered option?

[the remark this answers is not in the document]
latk: Hmm, let's forget about adding checks, since the other options don't have them either. That's a problem for a future enhancement.

So as long as the --add-tracefile help message is updated to clearly note that it will not run gcov to collect new coverage data, that is fine by me.

gcovr/json_generator.py
```diff
@@ -0,0 +1,71 @@
...
+#
+# Produce gcovr JSON report
+#
+def print_json_report(covdata, output_file, options):
+    r"""produce an JSON report in the format partially
+    compatible with gcov JSON output"""
+
+    gcovr_json_root = {}
+    gcovr_json_root['gcovr/format_version'] = JSON_FORMAT_VERSION
+    gcovr_json_root['gcovr/file_paths'] = []
```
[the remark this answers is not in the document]
CezaryGapinski: I renamed these keys to `files` and `file` in the new version.

gcovr/json_generator.py
```diff
@@ -0,0 +1,71 @@
...
+    if options.output:
+        with open(options.output, 'w') as output:
+            write_json(gcovr_json_root, output)
+    else:
+        write_json(gcovr_json_root, sys.stdout)
+
+
+def _gcovr_lines_to_gcovr_json_lines(lines):
+    gcovr_json_lines = []
+    for no in sorted(lines):
```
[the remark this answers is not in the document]
CezaryGapinski: Looks nice. Please review my new version of PR where all functions to convert from/to JSON are refactored.

gcovr/tests/add_coverages/reference/coverage_bar.json
CezaryGapinski: Agreed. These files actually are not necessary to check during tests.

gcovr/tests/add_coverages/Makefile
CezaryGapinski: I have added to Makefiles empty json target.

gcovr/json_generator.py
```diff
@@ -0,0 +1,71 @@
...
+    return gcovr_json_lines
+
+
+def _gcovr_branches_to_gcovr_json_branches(branches):
+    gcovr_json_branches = []
+    for no in sorted(branches):
+        gcovr_json_branch = {}
+        gcovr_json_branch['count'] = branches[no].count
+        gcovr_json_branch['fallthrough'] = branches[no].fallthrough is not None
+        gcovr_json_branch['throw'] = branches[no].throw is not None
```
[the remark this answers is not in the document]
CezaryGapinski: Agreed.

README.rst
```diff
@@ -21,12 +21,13 @@ The ``gcovr`` command can produce different kinds of coverage reports:
...
 -  ``--xml``: machine readable XML reports in Cobertura_ format
 -  ``--sonarqube``: machine readable XML reports in Sonarqube format
+-  ``--json``: JSON report with source files structure and coverage
 
 Thus, gcovr can be viewed
 as a command-line alternative to the lcov_ utility, which runs gcov
 and generates an HTML-formatted report.
 The development of gcovr was motivated by the need for
-text summaries and XML reports.
+text summaries, XML reports and gcov JSON intermediate format.
```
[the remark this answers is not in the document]
CezaryGapinski: Agreed.

gcovr/__main__.py
```diff
@@ -240,47 +241,61 @@ def main(args=None):
...
-                covdata[fname] = cov
-            else:
-                covdata[fname].update(cov)
-        toerase.update(context['toerase'])
-        rmtree(context['workdir'])
-    for filepath in toerase:
-        if os.path.exists(filepath):
-            os.remove(filepath)
+
+    if options.add_tracefile:
```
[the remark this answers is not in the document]
CezaryGapinski: Looks better, I've added it in separate commit. collect_coverage_from_{tracefiles, gcov} functions have also logger parameter.

gcovr/configuration.py
```diff
@@ -590,6 +590,27 @@ def choose(_cls, choices, default=None):
...
+        type=OutputOrDefault,
+        default=None,
+        const=OutputOrDefault(None),
+    ),
+    GcovrConfigOption(
+        "prettyjson", ["--json-pretty"],
+        group="output_options",
+        help="Pretty-print the JSON report. Implies --json. Default: {default!s}."
+             "OUTPUT is optional and defaults to --output.",
+        nargs='?',
```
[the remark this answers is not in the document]
CezaryGapinski: Fixed in new version.

gcovr/json_generator.py
```diff
@@ -0,0 +1,129 @@
...
+    json_lines = [_json_from_line(lines[no]) for no in sorted(lines)]
+    return json_lines
+
+
+def _json_from_line(line):
+    json_line = {}
+    json_line['branches'] = _json_from_branches(line.branches)
+    json_line['count'] = line.count
+    json_line['line_number'] = line.lineno
+    json_line['unexecuted_block'] = line.count == 0
```
latk: I'm not sure what the correct semantics for the `unexecuted_block` field are. Gcovr does not currently parse per-block data and cannot confidently say whether a line contains unexecuted blocks. Blocks have to do with partial coverage: a line can be covered but still contain unexecuted blocks, e.g. `if (foo()) bar();` where `foo()` always returns false.

I would get rid of this field in order to avoid incompatibilities with the gcov JSON format.

When loading, this field is used to populate the `noncode` property. That's a good idea, so maybe it would be best explicitly to store a `gcovr/noncode` field instead?

gcovr/json_generator.py
```diff
@@ -0,0 +1,129 @@
...
+    json_branch['throw'] = bool(branch.throw)
+    return json_branch
+
+
+def _lines_from_json(file, json_lines):
+    [_line_from_json(file.line(json_line['line_number']), json_line) for json_line in json_lines]
+
+
+def _line_from_json(line, json_line):
+    line.noncode = json_line['unexecuted_block']
```
latk: see above comment about the `unexecuted_block` field.

gcovr/tests/add_coverages/reference/coverage.json
```diff
@@ -0,0 +1,167 @@
+{
+    "gcovr/format_version": 0.1,
```
latk: This file contains a space after the comma, which causes all tests to fail except those under 2.7. There are two alternatives how this can be resolved:

* in the test runner, a JSON scrubber is added that trims all lines to remove the trailing space if any is present
* in the test runner, JSON files aren't compared textually but the JSON is loaded into a data structure that is then compared, similar to the current XML tests. This could be done in a scrubber or in a custom ASSERT_EQUALS.

gcovr/json_generator.py
CezaryGapinski: How about just to remove `unexecuted_block` filed (or `gcovr/noncode`after renaming)?
During conversion JSON file to `covdata` I can update the `noncode` property after fetching the `lineno`.

```diff
--- a/gcovr/json_generator.py
+++ b/gcovr/json_generator.py
@@ -92,7 +92,6 @@ def _json_from_line(line):
     json_line['branches'] = _json_from_branches(line.branches)
     json_line['count'] = line.count
     json_line['line_number'] = line.lineno
-    json_line['unexecuted_block'] = line.count == 0
     return json_line
 
 
@@ -114,8 +113,8 @@ def _lines_from_json(file, json_lines):
 
 
 def _line_from_json(line, json_line):
-    line.noncode = json_line['unexecuted_block']
     line.count = json_line['count']
+    line.noncode = line.count == 0
     _branches_from_json(line, json_line['branches'])
```

gcovr/tests/add_coverages/reference/coverage.json
CezaryGapinski: I can explicit set separators in json.dump, like here:

```diff
--- a/gcovr/json_generator.py
+++ b/gcovr/json_generator.py
@@ -37,7 +37,7 @@ def print_json_report(covdata, output_file, options):
 
     write_json = json.dump
     if options.prettyjson:
-        write_json = functools.partial(write_json, indent=PRETTY_JSON_INDENT)
+        write_json = functools.partial(write_json, indent=PRETTY_JSON_INDENT, separators=(',', ': '))
 
     if options.output:
         with open(options.output, 'w') as output:
```

JSON output in pretty-print mode will be without space at the end of line with comma, what can be useful if somebody wants to parse JSON file with external custom tool. Space at the end of line with comma can be annoying.

> 
- in the test runner, JSON files aren't compared textually but the JSON is loaded into a data structure that is then compared, similar to the current XML tests. This could be done in a scrubber or in a custom ASSERT_EQUALS.

It can help to pass test when one the file is in pretty-print mode and second without it, so I have to implement this comparison.

gcovr/json_generator.py
latk: No, noncode means that a line should not be included in the coverage calculation. That's not the same as an uncovered line. The two alternatives are:

* do not output noncode lines to the JSON. This is correct, but lossy.
* change `unexecuted_block` to something like `gcovr/noncode` or `gcovr/excluded`.

I'll leave it to you to determine the better approach.

gcovr/tests/add_coverages/reference/coverage.json
latk: Oh, I didn't know you could change the separators! If setting that option fixes the problem, that seems like the easiest solution. Do whatever *you* think is best, as long as the tests pass on Python 2.7 and 3.6.

CezaryGapinski:
Third version of PR. Changes:
- fix in gcov's `unexecuted_block` field - replaced to `gcovr/noncode` field
- added separators and sort_keys fileds to force ordering in dictionaries in JSON (useful during comparison between reference and result file, where they serialized to text representation)

gcovr/json_generator.py
CezaryGapinski: I think, now it should be correct. `gcovr/noncode` filed has `line.noncode` value.

gcovr/json_generator.py
CezaryGapinski: Now `line.noncode` has value from JSON `gcovr/noncode` field.

gcovr/tests/add_coverages/reference/coverage.json
CezaryGapinski: In my last PR I have added `separators=(',', ': ')` and `sort_keys=True` to the `json.dump()`.
Dictionaries in python are not ordered, therefore `coverage data` can be generated in different ways in JSON output files. `sort_keys=True` preserves alphanumeric ordering for dictionaries used in `json.dump()`. Content of output files will be generated always in the same order what is useful to compare JSON output file like a plain text files during tests.

I think thanks to that tests for Python 2.7 and 3.6 were passed.

Actually I though about comparing JSON structures during tests after gathering information from `json.load()`, convert this to the `coverage data` and next add `compare_coverages()` function to utils.py and use it, but maybe is better to differentiate between pretty-print and non pretty-print mode and compare raw text files. After conversion from JSON file to `coverage` we will lose information about pretty printing. What if in the future something goes wrong with pretty-print during adding some new features. `coverage data` will be ok, but pretty-print can be damaged without information about that during test.
It is always decision to make, usage of `separators=(',', ': ')` and `sort_keys=True` looks like to be the easiest one.

opikalo:
Anything I can do to help this PR to land?

CezaryGapinski:
> Anything I can do to help this PR to land?

For me this PR looks ok, maybe I missed something, but I do not have any feedback from @latk what should I improve.
I guess he has to decide if its is good enough to merge it.

latk:
This took a lot longer than expected, but I _finally_ merged this excellent PR. It seems I have become the bottleneck for gcovr development. If anyone would like write access in order to merge changes without having to wait for me, I'd be happy to arrange that.

CezaryGapinski:
Hi!
No problem at all. You did a lot of work with this project and you should continue!
Python is not my primary language, but If you want I can help. I just need some time to go back to this project and analyse it more deeply.

PR #328, opened 2019-08-27
Support JSON output for Coveralls.io

zfields:
[Coveralls.io](https://coveralls.io) is a code coverage web application, similar to codecov.io. They are lacking support for CPP projects, so this is a great opportunity for exposure for both `gcovr` and `coveralls`. My company uses Coveralls.io, and I have experience with Gcovr from my last position, so I thought I would like to contribute.

This PR produces a Coveralls compatible JSON file. The file is created as expected, and can be uploaded successfully. Hopefully I've done everything correctly, I have attempted to follow the pattern of the recently added Sonarqube generator and testing.

I am still not certain if the `SCRUBBERS` and `ASSERT_EQUALS` arrays are necessary if you are not generating output on `stdout`. I have implemented most, if not all, of the logic for the tests, but I disabled it because it was comparing the `example_json.json` file to an empty string returned by the `subprocess` invoked via the test framework.

The disabled logic is pulling the coverage down, so it can be quickly remedied once I know how to proceed.

Thanks,
Zak


zfields:
@latk I am stuck in need of guidance. Sorry to single you out, but you are one of the top and most recent contributors and I hope you can help me find someone to answer my questions. It's been a couple weeks and no one has commented on the PR.

latk:
Hi Zachary! Thank you for waiting despite the silence. And thank you for working on this feature. It looks _very_ useful so I'll be glad to eventually merge this. At first skimming, your code looks reasonable :)

---

> I am still not certain if the SCRUBBERS and ASSERT_EQUALS arrays are necessary

The test runner looks for files matching a particular pattern, looks for a matching filename in the `reference/` directory of the test, scrubs both files, and compares the result. The scrubber must be present but could be an identity function. The assertion can be empty. The stdout output is not directly checked.

The example runner uses the same infrastructure but looks at the stdout of the example. You could change the example runner so that it can also look at files. Alternatively, you can write the coveralls JSON to stdout by not mentioning a name on the command line:

``` diff
- gcovr -r . --coveralls example_json.json
+ gcovr -r . --coveralls
```

By the way, in the code and especially in the tests you refer to the format as “JSON”. Unfortunately this conflicts with the JSON format that is implemented in #326, just not yet merged. Please use the name `coveralls` for your format.

The example output file includes personal data and data that will become stale with future commits. Perhaps it would be better to write the example script so that it sets up a new git repo in a tempdir with fake commit data that's easier to verify.

---

Just to be clear, there will be no automatic upload? That's good because that keeps the security surface of gcovr more manageable.

---

> In order to get the "Python 3.4" tests to run with setup.py, I had to pin the version of lxml to v4.3.5 (two versions old).

If that fixes the tests for the time being that's OK. I plan to just drop support for Python 3.4 instead.

---

I see you've included a library for reading commit metadata. I'd rather not do this as it complicates licensing. If a comparable library is available on PyPI we can add a dependency on that, or we can write our own code to generate the required JSON structure.

gcovr/tests/test_gcovr.py
```diff
@@ -69,6 +81,12 @@ def assert_xml_equals(coverage, reference):
     compare_xml(reference_repn, coverage_repn, tolerance=1e-4, exact=True)
 
 
+def assert_json_equals(coverage, reference):
```
CezaryGapinski: I guess in this function you have content of two raw text files and it means you should compare coverage and reference as a text or use json.load(), and next compare python objects (I think you can use `compare_repn()` function for this from `pyutilib` like for `assert_xml_equals()`).

I used `json.dump()` with sort_keys=True during generation of json file to keep keys sorted and next compare coverage and reference as a two texts. Please check my Pull Request #326 because I had the same problems like you have :).

It is just my suggestion and @latk is an expert here :).

zfields: Thanks @CezaryGapinski! I appreciate the help. I can't wait to get this merged in.

zfields:
@latk Thank you for reaching out! As soon as I am able, I will update the PR accordingly. I am excited to contribute, and I can't wait have this change in a versioned tag on `gcovr`.

This not only helps my company, but fills a giant void in the Coveralls community!

zfields:
Changed `json` --> `coveralls`. I will continue to address code review concerns as time permits.

zfields:
@latk 

> I see you've included a library for reading commit metadata. I'd rather not do this as it complicates licensing. If a comparable library is available on PyPI we can add a dependency on that, or we can write our own code to generate the required JSON structure.

This is difficult for me. I'm a C++ developer and I don't develop in Python, in fact it has been quite difficult for me to bring this PR as far along as I have. It is my understanding, the **Apache 2.0** license is extremely similar to **MIT** lincense. I have placed it in it's own distinct subfolder, which should mark a clean seperation - they should not clash. [According to Wikipedia](https://en.wikipedia.org/wiki/Comparison_of_free_and_open-source_software_licenses) the only difference is in regard to trademarking.

If the licensing is still too much to overcome, I would request help from someone with advanced Python skills to please understand the output _(without looking at the source)_, and rewrite the functionality under the MIT license.

zfields:
> Just to be clear, there will be no automatic upload? That's good because that keeps the security surface of gcovr more manageable.

That is correct. There is NO upload capability, only JSON file creation.

latk:
* Due to merging the --json format (#326), there are now lots of smaller merge conflicts in the test Makefiles.
* I am going to stay firm on the licensing issue. Instead, we can use the original (and now vastly improved) code from <https://github.com/coveralls-clients/coveralls-python>. We can add an optional dependency to that package, i.e. using [`extras_require`](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) and adding a runtime check along the lines `try: import coveralls; except ImportError: error("please pip install coveralls")`
* Today I committed the change to drop support for Python 3.4, so that it's no longer necessary to pin the lxml version.

gcovr/coveralls_generator.py
```diff
@@ -0,0 +1,113 @@
...
+    root = {}
+
+    # Capture timestamp
+    root['run_at'] = datetime.utcnow().strftime("%Y-%m-%d %H:%M:%S UTC")
+
+    # Pull environment variables
+    root['repo_token'] = environ.get('COVERALLS_REPO_TOKEN')
+
+    # Consume Travis CI specific environment variables _(if available)_
+    root['service_job_id'] = environ.get('TRAVIS_JOB_ID')
```
Spacetown, line 45: Can you also consume the Jenkins variables?

gcovr/coveralls_generator.py
```diff
@@ -0,0 +1,113 @@
...
+    if (root['service_job_id'] is not None):
+        root['service_name'] = "travis-ci"
+        root['service_number'] = environ.get('TRAVIS_BUILD_NUMBER')
+        root['service_pull_request'] = environ.get('TRAVIS_PULL_REQUEST')
+    else:
+        root['service_name'] = ""
+        del root['service_job_id']
+
+    # Add last git commit information
+    root['git'] = gitrepo(options.root_dir)
```
Spacetown, line 55: Is coverall tied to GIT?
In Jenkins this is also available as environment, I guess in Travis also. If you run the command localy you don't realy need the commit and therefore the dependency can be dropped.

setup.py
```diff
@@ -25,7 +25,7 @@
       packages=['gcovr'],
       install_requires=[
           'jinja2',
-          'lxml',
+          'lxml==4.3.5',
```
Spacetown, line 28: Do not fix the version.

zfields:
I made this change to help the company I worked for at the time. I have completely lost context. If someone can build on the work I've done here to support `coveralls.io`, I would appreciate the help. Otherwise, I no longer work for that company, and at this time, I do not have time to get this over the finish line.

Spacetown:
Ok. I'll try to take over this branch.

zfields:
@Spacetown I do appreciate you taking over the branch, but I would have appreciated you leaving my original commit. You've removed my name completely from this contribution. :cry: :broken_heart:

Spacetown:
> @Spacetown I do appreciate you taking over the branch, but I would have appreciated you leaving my original commit. You've removed my name completely from this contribution. 😢 💔

Sorry for that. Because all makefiles have changed I had to squash before rebase. I'll add you again.
First I wanted to create an own branch but the PR can't be updated and I don't want to create a new on. Feel free to force push your branch and I'll create a new PR referencing this one.

Spacetown:
@zfields I've added you as co-author. Sorry again, the squash removed the information. But the rebase of a single commit took about 1 hour because ofresolving the conflicts in the test files.

zfields:
@Spacetown Cheers! Much appreciated! :+1:

AUTHORS.txt
```diff
@@ -56,4 +56,5 @@ The following developers contributed to gcovr (ordered alphabetically):
     Will Thompson,
     William Hart,
     Antonio Quarta,
+    Zachary J. Fields
```
CezaryGapinski, line 59: Definitely authors list is not in the alphabetical order :).

doc/source/guide.rst
```diff
@@ -341,6 +341,49 @@ instead.
...
+
+    gcovr --coveralls coverage.json
+
+The ``--coveralls-pretty`` option generates an indented JSON output
+that is easier to read.
+
+If available, environment variable COVERALLS_REPO_TOKEN will be
+consumed and baked into the JSON output.
+
+If running ain a CI additional variables are used:
```
CezaryGapinski, line 358: should be "in" instead of "ain"?

gcovr/__main__.py
```diff
@@ -419,6 +420,14 @@ def print_reports(covdata, options, logger):
                 "CSV output skipped - "
                 "consider providing output file with `--csv=OUTPUT`.")))
 
+    if options.coveralls:
```
CezaryGapinski, line 423: `or options.coveralls_pretty`

gcovr/coveralls_generator.py
```diff
@@ -0,0 +1,197 @@
...
+        CurrentBranch = os.environ.get('BRANCH_NAME')
+    elif (os.environ.get('GCOVR_TEST_SUITE') is not None):
+        json_dict['service_name'] = "gcovr-test-suite"
+        json_dict['service_job_id'] = 'id'
+        json_dict['service_number'] = 'number'
+        json_dict['service_pull_request'] = 'pr'
+        CurrentCommit = None
+        CurrentBranch = 'branch'
+
+    git = shutil.which('git')
```
CezaryGapinski, line 105: Well, I don't know how useful it is but maybe is better to use something like `--gcov-executable` option when we can point out to particular `gcov` exacutable. We can also set `gcov` exacutable by `GCOV` variable. Maybe it is good to to the same for `git`.

AUTHORS.txt
Spacetown: I'll sort the list.

doc/source/guide.rst
Spacetown: Sure.

gcovr/__main__.py
Spacetown: Sure.

gcovr/coveralls_generator.py
Spacetown: For me the option make sense for gcov. There can be installed several versions at the same time. But for git I don't know if this is an use case.
I can add the option if you think thats an use case.

CezaryGapinski: Ok, it was just proposal. I also can't find any practical example where it can be used, so leave it as it is.

Spacetown:
@zfields You can remove your branch. 🎉

zfields:
@Spacetown Thank you so much for all your help!

latk:
I'm glad this feature was dragged over the finish line (thanks Spacetown!)

Unfortunately the Appveyor tests are now failing on the master branch, it seems that `"service_pull_request"` is set to `null` rather than the empty string as expected. How can we properly override such environment effects within our test suite?

chrta:
Discalimer: I don't know about appveyor etc

> Unfortunately the Appveyor tests are now failing on the master branch, it seems that "service_pull_request" is set to null rather than the empty string as expected. How can we properly override such environment effects within our test suite?

In that case it seems that APPVEYOR_PULL_REQUEST_NUMBER is not set in the environment and `json_dict['service_pull_request'] = os.environ.get('APPVEYOR_PULL_REQUEST_NUMBER')` should be `json_dict['service_pull_request'] = os.environ.get('APPVEYOR_PULL_REQUEST_NUMBER', "")` if this must not be None, but an empty string.

Spacetown:
According to the docs (https://docs.coveralls.io/api-reference) the key `service_pull_request`is optional. I would prefer to remove the key if the value is `None` or an empty string for all CI's.

Spacetown:
Master is fixed. Sorry.

PR #334, opened 2019-09-26
Update gcov.py

alex43dm:
UnicodeDecodeError: 'utf-8' codec can't decode byte

### DIFF
--- a/gcovr/gcov.py
+++ b/gcovr/gcov.py
@@ -642,6 +642,7 @@ def run_gcov_and_process_files(
     # NB: Currently, we will only parse English output
     env = dict(os.environ)
     env['LC_ALL'] = 'en_US'
+    env['LANGUAGE'] = 'en_US'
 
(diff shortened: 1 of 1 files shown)


latk:
Thank you for this pull request! I was not sure why this change was necessary, but it seems that gcov uses the GNU gettext translation software, which has a complex interaction with environment variables:

> GNU gettext gives preference to LANGUAGE over LC_ALL and LANG for the purpose of message handling, but you still need to have LANG (or LC_ALL) set to the primary language; this is required by other parts of the system libraries.
>
> — https://www.gnu.org/software/gettext/manual/gettext.html#The-LANGUAGE-variable

So if a user has LANGUAGE set, overriding LC_ALL is not enough – we need to override both variables.

Thanks to this change, the language override should be correct. However, I'm not yet sure whether this also forces UTF-8 encoding, so future bugs are still possible.

PR #335, opened 2019-10-27
Change relative imports to absolute imports.

Sinclair-John:
PEP8:
> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages)...

Also something like `python gcovr\gcovr\__main__.py --version` will not work with relative imports.

### DIFF
--- a/gcovr/__main__.py
+++ b/gcovr/__main__.py
@@ -39,22 +39,22 @@ from os.path import normpath
 from tempfile import mkdtemp
 from shutil import rmtree
 
-from .configuration import (
+from configuration import (
     argument_parser_setup, merge_options_and_set_defaults,
     parse_config_file, parse_config_into_dict, OutputOrDefault)
-from .gcov import (find_existing_gcov_files, find_datafiles,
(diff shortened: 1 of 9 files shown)


latk:
Thank you for working on improvements to gcovr! However, I have to refuse these changes.

1. PEP 8 should not be taken too seriously. And while it recommends against relative imports in the general case, it does not outlaw them entirely. I believe that PEP-8 does not recommend against relative imports as used by gcovr.

2. Gcovr is designed so that it must be installed as a Python module, ideally using `pip`. The main file cannot be executed directly. Once it is installed as a module, it can be invoked like `python3 -m gcovr`.

    This is intentional: pip is necessary anyway in order to install the various dependencies. Installation is also able to automatically create the correct wrapper scripts under windows.

3. This change is incorrect, and breaks gcovr under Python 3 (see the Travis/AppVeyor test results). For example, you propose that we `import utils` – but no such module is known to Python. Importing `gcovr.utils` might work and would comply with the PEP-8 guidance, but as discussed above I believe that level of compliance is unnecessary.

PR #336, opened 2019-10-27
Use --coverage instead of -fprofile-arcs -ftest-coverage

Sinclair-John:


Spacetown:
Why should the documentation be changed?

Issue #277, opened 2018-08-27
Configure tab size for HTML report


itavero:
At work we use hard tabs (so no spaces) in our code.
It would be nice if we could configure the CSS `tab-size` property when calling Gcovr.
That way we can set it to `3` (in our case) and have the HTML reports be aligned properly.

In most browser it defaults to `8`, but I would say a default of `4` or even `2` would also make a lot of sense (that are defaults used by a lot of IDEs).

latk:
Thanks for the idea! I won't add an option specifically for tab size, but I too want more configurable CSS.

Currently, the CSS is inlined into the HTML. That's unnecessary duplication (70% of the gcovr code base consists of HTML test files…) and makes the reports difficult to style.

I'd like to see command line options so that:

 - the CSS can be written into a separate file
 - the HTML reports can reference an external file (so that the external CSS is shared by all reports)

This would allow you to provide a patched CSS file instead. It would also make the gcovr test suite much smaller. Finally, external CSS might help with the perpetual Jenkins-CSP problems (#125).

I am not sure how these options should be called and how they interact, so that all of the following cases are covered:

 - the CSS is inlined into each HTML report – what is currently being done
 - the CSS is written to an separate file that is linked by each HTML report
 - external CSS is provided that is linked by each HTML report. Gcovr's CSS isn't used.
 - CSS is written into a separate file **and** external CSS is provided. The external CSS is linked by each HTML report. The external CSS can `@import` gcovr's CSS file to override its contents.

Would you like to try implementing this? I'm afraid the `gcovr/html_generator.py` code is a bit of a mess right now, but only the `data['CSS']` variable plus the I/O should be relevant here.

itavero:
That also sounds like a very good suggestion. My calendar is full for the next few weeks, but I'll see if I can find some time after that.

Spacetown:
Found this old issue.
Now the CSS is an external file and this enhancement can be implemented quite easy.
If an external file is given should the the file be copied beside the HTML, the given path added as it is or changed to an absolute path? There could also be an additional parameter to control what to do.

latk:
Resolved by #377: CSS tab size can now be specified via `--html-tab-size` option. Inclusion of external CSS hasn't been implemented yet.

Issue #160, opened 2016-12-11
Is it possible to hyperlink and nest directories in --html-details output?


rsmz:
I have a project with thousands of C++ files organized in a large tree, however, the gcovr html details output presents them all in a single flat table.  Is there an option to nest and hyperlink all the html pages just like the source code tree, as well as links to parent directory pages?  I think lcov did something like that.  If not, I'd be happy to contribute if given some hints on where to start.  Each row for a sub-directory could show totals.  I would also have a link to a flat view like now for each directory.

Also, the html table doesn't maximize to the width of the browser, so my C++ file name links are wrapped while there remains ample unused space in the left/right margins.

latk:
Thank you for your suggestion. This is an often-requested feature, and I too want to see this happen.

However, I'm planning to do larger changes to gcovr's architecture and to the HTML report templating. After those changes have gone through, we can revisit this issue. Implementing nested reports will be much easier after untangling the current HTML generation code.

macetw:
@latk Lukas, I'm working on this feature: my project needs this. I'm pretty satisfied but not fully complete on this feature.

I'm curious if the desired behavior is to always show the nested tree or also show a nested tree. That is, do we allow the option of both a flat-list and a hierarchical list?

I'm not yet clear on whether I can contribute my changes back to the source, but I sure hope I can. It wasn't terribly difficult, and I'd be happy to be a part.

Spacetown:
Thank you for working on this. Please share your work as draft PR than we can discuss about it.

chreniuc:
Any news on this? I've noticed that this was requested 6 years ago.

Spacetown:
I'm also waiz8for a response of @macetw .

zchrissirhcz:
@Spacetown @macetw I suggest hte official gcovr team / organization do the request feature instead of the infinite long waiting, since there is no public branch or commit in @macetw 's fork demonstrate that there is a (draft) PR.

PR #342, opened 2019-11-28
Use bootstrap 4

chrta:
This is a subtask from #43 and is based mainly on
gcovr commit bc2abd20ffb0a684cc3c407d7c8e60124809a146


chrta:
You may see it as an RFC, since i do not know how the output should look and i am a total beginner regarding html and css.

latk:
Thank you for working on this! The HTML reports are overdue for a facelift. I've previously avoided changes here since the HTML reports are the main way that gcovr is tested, but now the JSON reports are taking up that role. For the build to pass all of the HTML reference files in all of the tests would have to be updated, but before doing that work we need some discussion first:

1. Do we want external resources for the HTML files?
2. How do we want the HTML reports to look?
3. Do we want to use Bootstrap?

## External resources

Previously, the HTML reports consisted of self-contained files. This made them easy to share, but had two drawbacks:

* the use of inline `style="..."` attributes caused problems with the [Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) of some servers, notably when used with Jenkins (#125)
* small changes to the CSS required ~50 reference files to be updated which is a lot of tedious work

Extracting the CSS into separate files mitigates these issues, but makes it more difficult to share the reports.

My intuition would be to introduce a new command line option (e.g. `--html-css-external`) that controls whether the styles are linked or included.

## Report layout

Here are screenshots from reports created by the proposed change:

| Summary | Detail |
|------------|-------|
| gcovr-proposed-summary | gcovr-proposed-details |

This definitely looks better, although I see two issues:

* the source code lines are too loose with the Bootstrap style
* applying a pretty style on these reports doesn't fix their main issue: that they are layouted in a 90s style manner with HTML tables, which makes them difficult to change (and totally unresponsive!)

If we are going to touch the HTML reports anyway, it may end up being simpler to re-create them from scratch, using modern approaches. Earlier this year I tried doing some incremental improvements, but it was incredibly tedious.

## Do we even want to use Bootstrap?

Bootstrap makes it easy to quickly get a decent-looking website, but it involves including a lot of CSS that we don't really need: modern HTML/CSS is pretty good by itself, as long as we're not interested in supporting Internet Explorer :)
For example, [flexbox](https://caniuse.com/#feat=flexbox) and [grids](https://caniuse.com/#feat=css-grid) are now widely supported, which obsoletes one of the main use cases of Bootstrap.

## Suggested approach

I see the proposed change as a great improvement over the status quo, but as more of a bandaid than a permanent solution. I am therefore not enthusiastic about switching the styles to Bootstrap (but would of course be prepared to change my mind).

Instead, if you have the bandwidth for doing so, it might be a better idea to rewrite the HTML generator more or less from scratch. Aside from the `--object-directory` option, **the HTML generator is the biggest chunk of technical debt in gcovr**. The first step would be to rewrite the generator to output the data in a less insane HTML structure (tables only for tabular data), the second step would be to apply nicer styling and layouting.

I can help with these steps (providing guidance and reviews re Python, Jinja, HTML, CSS) but don't currently have the time to take on this work myself. That you aren't a web whiz isn't a problem but may actually be a benefit, since the goal would be to produce something less complicated than the status quo. Less complicated means that future improvements will be become easier.

Working on the HTML generator could optionally tie in with these other tasks/efforts:

* getting rid of Python 2 compatibility
* the Jenkins CSP problem (mentioned above)
* making source lines linkable (#180)
* making the CSS or even the HTML templates user-configurable/themeable (#277)
* making nested (per-directory) HTML reports possible (#160)

What do you think?

chrta:
Thank you for your detailed feedback. I was expecting some kind of discussion, that is why i did not adapt the tests etc. If a modern look can be achieved with plain css and html, i also think this is better.

If you would like a rewrite of the html generator, i think this would be good:

1. Rewrite the data model jinja2 uses. I would prefer something that is independent from the output format (html). Bring in all the html-specific stuff in the jinja2 templates. Later on the same data could be used by different exporters to create other output formats using other jinja2 templates.
   * use some kind of nested dicts for the data, use classes where it makes sense
   * keep html completely out of the data
   * only change this in `html_generator.py` in the beginning, do not touch other code
   * only python3
   *  keep the produced html as it is now in the master branch (or as close as possible)

2. In the next steps i would try to incrementally improve the html output you mentioned in you "Suggested Approach"

I would spare some time to improve gcovr, but first i would like to discuss how to go on. I would like to do this incrementally to have multiple small pull requests.

What are your thoughts, especially about point 1.?

latk:
Ok, that approach sounds sensible! Yes, the HTML as part of the data model is not exactly good, but at the time it was the least invasive approach that let us start using Jinja templates :)

Splitting the work into multiple incremental PRs is a very good idea. However, the master branch should always be in a releasable state, which implies that the test suite should pass. This is effectively incompatible with the sweeping HTML changes that are necessary.

I'll therefore open a new html-rewrite branch as a target for incremental changes (tomorrow?), and will make sure that  it doesn't diverge too far from master.

* before opening that branch, I will remove the CI tests for Python 2
* once that branch is created, we can remove the old HTML reference files. That way the tests will still run, but actual HTML output is ignored.
* various incremental changes are made…
* at some point, test reference files are added again
* once the rewrite branch has settled, it will be merged and changes can continue on master

Actually, it's not necessary for the HTML tests to run for every test case since we now have the comprehensive JSON tests. Some time during the rewrite would be an opportunity to change the test approach, but doing actual HTML changes has priority – there's only so much yak shaving we can do without losing the actual goal out of sight, and I've already moved the goal from “CSS redesign” to “HTML generator rewrite”…

Some further pointers:

* Avoid a reusable data model but stay specific to the two HTML variants (summary + details). There already is a shared data model in `gcovr/coverage.py`. However, it may be OK to extend the shared model with convenience functions, e.g. for calculation coverage percentages. The shared model doesn't contain source code lines, and should likely stay that way.

* There's a Gitter chatroom for gcovr that can be used for smaller discussions. Still, most discussion should stay on GitHub for visibility + linkability.

* Initially, staying close to the current HTML is a good idea. But don't stick to it too closely where it is too restrictive. E.g. the summary header at the top of each HTML file currently consists of a single `<table>`, but splitting it into multiple tables will likely be a lot easier.

latk:
I've created a `html-rewrite` branch that can be used as a target for incremental work. Python 2 support has been dropped, and the old HTML reference files have been removed, that way you have a clean slate for any rewrites.

Have fun!

chrta:
I am closing this, since the html-rewrite branch is merged and we do not want to use bootstrap

PR #350, opened 2020-02-04
Added decision coverage to HTML output

glenntws:
Hello everybody,

my company is active in embedded software development for the automotive industry. To achieve certain levels of security (ISO 26262, ASIL A,B and C), "taken at least once" is completely overkill and "branches executed" is not usefull to see, if certain decisions have been made by the software.

The result is that I implemented a new decision coverage feature which extends the already existing outputs.

A new test case also is integrated.

The strategy how the decisions get analyzed is the following:

1. Try to find if-/if-else-/switch-/while-/for-/do-while-statements and check for the consecutive lines, if they are executed
2. If this isn't possible or not accurate enough (ternary expression or branches in one line), use the info of the branch coverage (only working when it's a simple condition, not a complex one with multiple conditions. for further info on why that is, just ask me.)
3. If both strategies don't work, mark the decision as uncheckable. A warining is given in the header of the HTML file.

Maybe someone is interested in integrating this output into other file formats. HTML was the most important for us for now.


latk:
Thank you for working on this! And for investing the effort of contributing your changes upstream! Many users have been confused by gcovr's branch coverage, so this might present a useful alternative.

However, I'm somewhat uneasy with munging source code to attribute branch coverage to individual control flow constructs, as I discussed [somewhere in the gcovr FAQ](https://gcovr.com/en/4.2/faq.html#why-does-c-code-have-so-many-uncovered-branches). It's worth pointing out that nothing\* in gcovr is currently C or C++-specific, but that gcovr should work with any language that's supported by GCC/LLVM. So while I look forward to merging this, I'll have to think about what the default behaviour should be (balancing usability, correctness, back-compat, and so on).

\* --exclude-unreachable-branches is a heuristic that assumes C-like curly-brace languages

This PR has some interactions with other work:

* There's some other work on the HTML generator going on right now, which lives on the html-rewrite branch (and I have some unpushed changes as well). I'm sorry you had to slog through the old code. Don't rebase right now, but the other changes have priority.
* Your work largely consists of parser/datamodel changes, and changes to the HTML reports. This is separable. So maybe it does make sense to split off the HTML changes into another PR, so that this PR can proceed faster?
* When updating the internal datamodel, it would be good to also update the JSON reader/writer since that's needed for the add-tracefile feature.

Those are just some things of the top of my head. I'm looking forward to look at this PR in more detail later, perhaps as early as this weekend.

latk:
I have pushed some changes to the html-rewrite branch. Could you redo your changes to the HTML reports on that branch, i.e. rebase the entire pull request? The rewritten HTML generator is much simpler, and details about branch coverage is now shown in a popup like this:

Screenshot_20200208_175008

You could add the detailed messages about decision coverage to this popup. You can also rename the Branch column to show Decision data, but only if the user opts in. Perhaps a `--decision-coverage` command line flag would make sense? You can add flags in the configuration.py file, e.g. see the `--branches` option.

Also, I really do not understand the decision parsing code in `gcov.py`. Is it necessary to do this within the main GcovParser, or could it be extracted to a separate class or module?

latk:
Thank you for rebasing the pull request! I just pushed two further commits to the html-rewrite branch which should fix some of the test failures you were seeing, but the CI failures also indicate a typo in your `calculate_decision_coverage()` function. It seems you still have some old HTML test files. They would have to be regenerated, I think.

I'll take a deeper look into your changes over the next couple of days. Displaying decision coverage alongside branch coverage data is fine by me.

During the last few days I found and fixed some bugs so I am planning to cut a release within the next few weeks, and I want this decision coverage feature to be part of the release. With regards to this feature, the blockers are:

* getting the tests to pass (but that won't be difficult)
* writing documentation on the feature (in the docs on HTML output, in the FAQ section on branch coverage)
* including decision coverage in the JSON output (and documenting the format)

If you have time to work on this that would be great, otherwise I'll do a bit but I will be slower.

glenntws:
Yeah, I'm currently fixing some typos I made and some other refinements. Also adding the feature to enable the decision coverage analysis via argument.

Since I think that this feature is pretty relevant and my company also is interested in that, I'm able to spend quite a bit of time for this feature, so it's no problem for me to do as many of the tasks as possible!

I'm guessing with a new HTML output and this feature, the next release will be a new major?

glenntws:
Okay, so besides the documentation (working on that right now), most things are done.

However, I still have some questions regarding the JSON-output: Right now - no matter if the --decisions argument is used or not - a dict called "decision" is created for every line, it's just empty for every line if the feature is disabled.

I'm thinking if this should stay that way, or if the output of the "decision" dict should be disabled completely when the feature isn't enabled.

And another notice: The HTML test should pass now, but it seems that it fails because of the different timestamps in reference and test. Somehow the test runner doesn't seem to change the timestamps to zeros... I didn't have time to check that in detail though.

latk:
Thank you so much, this is converging to a state that's ready for merging. I will merge the html-rewrite into master prior to this, so the effective “size” of this pull request will be smaller.

I'd be happy if you could add some more comments to the decision coverage analysis to make it clearer _why_ you are doing something. This isn't necessarily obvious to me. You can add example C code snippets in these comments to show when which heuristic applies. Perhaps this could even be a doctest, like:

``` python
def prep_decision_string(code):
    """remove comments from  a line of code.

    >>> prep_decision_string('  int x = foo(); // fooify something')
    '  int x = foo(); '
    """
    ...
```

The HTML test uses some “scrubber” regexes to remove timestamps and specific versions, but these were broken during the HTML rewrite. I can fix them tomorrow, or you can try to update them.

gobater:
@glenntws: cool that you mad this. I was also looking for this feature.
C1 -> Branch Coverage (If/Else)
mcdc -> is actually what was shown in gcovr!

Spacetown:
I've rebased the branch to current master and created a PR to the fork of @glenntws.

macetw:
Is this PR to be abandoned? There are tons of conflicts here.

Spacetown:
@macetw I think a cherry pick based on the current master would be best.

Spacetown:
@latk I've squashed the branch for better handling of rebase because there where several changes which where undone.
The test data was also outdated. Can I merge it or do you want to review the changes?

gsemet:
Hope this can be integrated !

gcovr/templates/source_page.html
```diff
@@ -108,6 +125,40 @@
...
+            {% endif %}
+            {% endif %}
+            {% else %}
+            {% if loop.index0 % 2 == 0 %}
+            <div class="uncheckedDecision">? Decision 'true' couldn't be analyzed.</div>
+            {% else %}
+            <div class="uncheckedDecision">? Decision 'false' couldn't be analyzed.</div>
+            {% endif %}
+            {% endif%}
+            {% endfor %}
```
latk, line 156: There's a lot of business logic going on here. Could some of it be moved into the Python code? At least, the decision name (true/false) could be handled outside of the template.

gcovr/tests/update-data/reference/coverage.update-data.h.html
```diff
@@ -0,0 +1,517 @@
+
+<html>
+
```
latk, line 3: Reference files like these seem to be a leftover from the time when we didn't separate references by compiler name. I think these should be removed.

gcovr/writer/html.py
```diff
@@ -517,6 +565,35 @@ def source_row_branch(branches):
...
+            'name': decision_id,
+            'count': decision.count,
+        })
+
+    return {
+        'taken': taken,
+        'uncheckable': uncheckable,
+        'total': total,
+        'decisions': items,
+    }
```
latk, line 594: Should add a `name` field here to simplify the template. I think the value of that field would be `name = 'true' if decision_id % 2 == 0 else 'false'`.

gcovr/writer/json.py
```diff
@@ -248,6 +250,19 @@ def _json_from_branch(branch):
...
+def _json_from_decision(decisions):
+    json_decision = {}
+    if len(decisions) == 1:
+        json_decision["is_switch"] = True
+        json_decision["count_true"] = decisions[0].count
+    elif len(decisions) == 2:
+        json_decision["is_switch"] = False
+        json_decision["count_true"] = decisions[0].count
+        json_decision["count_false"] = decisions[1].count
+        json_decision["uncheckable"] = decisions[0].uncheckable
```
latk, line 262: Is it guaranteed that there will be exactly 0, 1, or 2 decisions on the line? Because the JSON format MUST round-trip losslessly. A data format like this that interprets the decision as a “switch” might be more immediately useful, but it would greatly limit how the JSON format can be used by gcovr itself, for example to merge coverage data from multiple test runs.

latk (changes requested):
I won't pretend to understand the heuristics that parse the source code, but for reasonably-formatted code it should be fine. And since this feature is opt-in, the general design is OK.

There are various points that should be addressed before this can be released, though. Not fixing these issues would produce significant technical debt down the line.

For example, the JSON format does not look like a lossless serialization of our data model. That the decision coverage analyzer opens the source code files by itself required significant changes throughout other code.

doc/examples/example_html.details.css
```diff
@@ -346,6 +356,21 @@ span.uncoveredLine
...
+
+.notTakenDecision
+{
+    color: Red!important;
+}
+
+.uncheckedDecision
+{
+    color: DarkOrange!important;
+}
```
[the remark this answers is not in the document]
latk: (ignore this comment, that was an example output file, not a template file :facepalm:)

gcovr/configuration.py
```diff
@@ -623,6 +623,13 @@ def merge_options_and_set_defaults(partial_namespaces, all_options=None):
              "For text report only.",
         action="store_true",
     ),
+    GcovrConfigOption(
+        "show_decision", ["--decisions"], config='txt-decision',
```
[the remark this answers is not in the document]
Spacetown: I was also wondering about this. I'll remove the config key and will do the same thing for `--branches`.

gcovr/coverage.py
```diff
@@ -61,6 +61,52 @@ def update(self, other):
...
+    def __init__(self, count, uncheckable):
+        # type: (int, Optional[bool]) -> None
+        assert count >= 0
+        self.count = count
+        self.uncheckable = uncheckable
+
+    @property
+    def is_covered(self):
+        # type: () -> bool
+        return self.count > 0
```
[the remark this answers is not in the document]
Spacetown: I think the writer should take care of this. If we return False in this case, the information is lost.

gcovr/gcov_parser.py
```diff
@@ -347,6 +352,10 @@ def parse_coverage(
     for function in state.deferred_functions:
         _add_coverage_for_function(coverage, state.lineno + 1, function, context)
 
+    if flags & ParserFlags.PARSE_DECISIONS:
+        decision_parser = DecisionParser(filename, coverage, source_encoding, logger)
+        decision_parser.parse_all_lines()
+
```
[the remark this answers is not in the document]
Spacetown: It was only a squash and rebase from my site. This was also my intention.

gcovr/writer/json.py
Spacetown: If there are 0 decisions the function isn't called. I think a key/value would be better than using an index and a representation in JSON like e.g.:
```json
{
   "switch" : { "count_true": 3 }
}
```
```json
{
    "count_false": 1,
    "count_true": 3,
    "uncheckable": false
}
```

gcovr/templates/source_page.html
Spacetown: I'll also invert the if, I hate a negated if with an else branch.
```
if xxx == False
   ...
else
   ...
endif
```

gcovr/gcov_parser.py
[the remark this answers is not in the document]
Spacetown: This should also be possible for the HTML writer but I think we will get memory troubles for big projects.

gcovr/configuration.py
```diff
@@ -617,12 +617,19 @@ def merge_options_and_set_defaults(partial_namespaces, all_options=None):
         default=None,
     ),
     GcovrConfigOption(
-        "show_branch", ["-b", "--branches"], config='txt-branch',
+        "show_branch", ["-b", "--branches"],
```
latk, line 620: For the --branches flag, removing the `txt-branches` config key is a backwards-incompatible change. Existing configuration files would no longer work.

gcovr/coverage.py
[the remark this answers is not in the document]
latk: Ok, let's leave it like this for now.

As I'm looking through the code I'm getting the feeling that this decision coverage data model isn't a good fit for the actual data, but that's not something that has to be addressed before merging. Can be refactored later if a better data model becomes clear.

gcovr/coverage.py
```diff
@@ -61,6 +61,47 @@ def update(self, other):
...
+    def update(self, other):
+        # type: (DecisionCoverage) -> None
+        r"""Merge DecisionCoverage information"""
+        self.count += other.count
+        self.uncheckable = self.uncheckable or other.uncheckable
+
+    def update_count(self, count):
+        # type: (int) -> None
+        r"""Increase the counter by the given number"""
+        self.count += count
```
latk, line 102: All calls `a.update_count(b)` can be replaced with `a.count += b`, so this method isn't needed.

gcovr/gcov_parser.py
```diff
@@ -276,6 +280,7 @@ def parse_coverage(
     logger: Logger,
     exclude_lines_by_pattern: Optional[str],
     flags: ParserFlags,
+    source_encoding: str,
```
latk, line 283: Source encoding is now unused. Removing this parameter requires updating the tests that call this function.

gcovr/templates/source_page.html
```diff
@@ -108,6 +125,28 @@
...
+      <td class="linedecision">
+        {% if row.linedecision %}
+        <details class="linedecisionDetails">
+          <summary class="linedecisionSummary">{{row.linedecision.taken}}/{{row.linedecision.total}}</summary>
+          <div class="linedecisionContents">
+            {% for decision in row.linedecision.decisions %}
+            {% if decision.uncheckable %}
+            <div class="uncheckedDecision">? Decision '{{decision.name}}' couldn't be analyzed.</div>
+            {% else %}
+            {% if decision.taken %}
```
latk, line 138: Jinja supports `elif` as in Python, which avoids the need to nest if/else.

gcovr/writer/json.py
latk: I think the current JSON data model is better than the `DecisionCoverage` data model at representing the result of the decision coverage analysis. Indeed, using indexes to disambiguate the true/false decisions is confusing.

I'd leave the JSON format as it is, or use a more extensible representation like

```json
{
  "type": "switch",
  "count": 3
}
```

```json
{
  "type": "conditional",
  "count_true": 1,
  "count_false": 3,
}
```

```json
{
  "type": "uncheckable"
}
```

Whether using a representation like `{ "switch": {...} }` or `{ "type": "switch", ... }` is mostly a matter of style. The former is more human-readable, the latter is typically easier to handle in software. I have no strong feelings about this.

latk (approved):
More comments! My greatest concerns have been addressed, so in principle this is ready to merge.

I now understand that the JSON data model is a more accurate description of the decision coverage analysis than the internal data model, so the JSON format is OK (but could maybe improved).

Most important remaining change: not removing the `txt-branch` config key, as removing it would break backwards compatibility.

gcovr/writer/json.py
```diff
@@ -229,6 +229,8 @@ def _json_from_lines(lines):
 def _json_from_line(line):
     json_line = {}
     json_line["branches"] = _json_from_branches(line.branches)
+    if line.decisions:
+        json_line["decision"] = _json_from_decision(line.decisions)
```
latk, line 233: It just occurred to me that this field should be namespaced to avoid incompatibilities with the upstream gcov JSON format.

```suggestion
        json_line["gcovr/decision"] = _json_from_decision(line.decisions)
```

Would require corresponding changes to the JSON reader.

gcovr/writer/json.py
Spacetown: I'll improve the JSON data model with the type, this is easier to understand. I also check if this data model can be applied to `DecisionCoverage`.

gcovr/configuration.py
Spacetown: Ok, but this key is also not only for txt output.

latk: I know, but we didn't think about that at the time when this config key was introduced :)

If we want to break backwards compatibility, then deliberately. We can do that if we really want to: the documentation on config files still says this feature is still experimental, after all.

Spacetown: I've undone the change and we can create a seperate PR to cleanup the configuration keys.

gcovr/writer/json.py
Spacetown: I've added seperate coverage classes and changed the JSON using the `type` key.

gcovr/coverage.py
```diff
@@ -197,8 +260,10 @@ def update(self, other):
         for branch_id, branch_cov in other.branches.items():
             self.branch(branch_id).update(branch_cov)
 
-        for decision_id, decision_cov in other.decisions.items():
-            self.decision(decision_id).update(decision_cov)
+        if self.decision is None:
+            self.decision = other.decision
+        else:
+            self.decision.update(other.decision)
```
latk, line 266: This update could fail for conflicting decision types, but I don't think that can occur in practice. So I'm fine with this for now. Not sure what an elegant solution would be.

gcovr/coverage.py
```diff
@@ -213,14 +278,23 @@ def branch_coverage(self):
...
+                unchecked = True
+            elif self.decision.is_conditional:
+                total = 2
+                if self.decision.count_true > 0:
+                    cover += 1
+                if self.decision.count_false > 0:
+                    cover += 1
+            elif self.decision.is_switch:
+                if self.decision.count > 0:
+                    cover += 1
```
latk, line 295: I think a switch-condition must set `total = 1`. The changes to the test reference files look suspicious.

gcovr/coverage.py
```diff
@@ -213,14 +278,23 @@ def branch_coverage(self):
...
-        unchecked = 0
-        for decision in self.decisions.values():
-            if decision.is_covered:
-                cover += 1
-            elif decision.is_uncheckable:
-                unchecked += 1
+        unchecked = False
+        if self.decision is not None:
+            if self.decision.is_uncheckable:
+                unchecked = True
```
latk, line 286: Not sure how uncheckable decisions should be handled. The comment in the decision analyzer suggests that this decision type is used if there might be *more* than two alternatives. So setting `total = 2` in that case might be reasonable. Must be `total = 1` at least. For comparison, the current state of the HTML writer uses the equivalent of `total = 1`.

latk (commented):
More comments! The new coverage data model raises questions how totals should be calculated.

But I'm really happy that we're managing to tear through the backlog of pull requests! We're close to getting this one over the finish line as well.

One thing that I noticed is that there are no docs for this feature, currently. I think I can write a section about this feature *after* the PR is merged. (Have been toying with the docs anyway, will submit a PR soonishly.)

gcovr/coverage.py
Spacetown: That can occure if the preprocessor is used and different code is generated but I think for the moment I've no solution for this.

gcovr/coverage.py
Spacetown: `+= 1` results in a 1 but we can also set it hard to 1.
In the test only unchec kable decisions are changed because ther was a true and false part and now it's only one decision, shall a uncheckable be counted twice?

gcovr/coverage.py
Spacetown: But it can also be 3, 4 or ... what is the correct value. What about using `number of branches / 2`?
The coverage in % depends on the used value.

gcovr/coverage.py
Spacetown: Forget my comment, the snippet is incrementing `cover`. You're right.

gcovr/coverage.py
latk: `2` is a lower bound on the true value. The number of branches is the upper bound on the true value. I think it makes sense (for simplicity) to either offer a most optimistic or most pessimistic estimate of the decision coverage.

With `total = 2` we're choosing the most optimistic estimate. But I think that will be fine, especially since it might still underestimate the true coverage (we count uncheckable decisions as _entirely_ uncovered).

I'm not entirely sure what decision coverage is supposed to represent, since the metric being calculated here is a lot like our branch coverage with all the weird C++-isms stripped out. If decision coverage is supposed to be branch coverage for lines that *look* like they should contain a conditional, then using the branch coverage information for total/covered would allow a more accurate estimation for uncheckable lines.

Not sure what to do here, so I'd rather leave it as it is now (total = 2) and maybe revisit this later.

Spacetown:
Then I'll merge back after fixing the CI.

Issue #338, opened 2019-11-06
Aggregation of coverage measurements for code with ifdef


peter-toft-greve:
I have tried to raise a question on how to aggregate coverage measures for code with ifdef
in https://stackoverflow.com/questions/58632496/how-to-get-100-coverage-with-gcovr-with-ifdef-code
Basically - how can I summarize or aggregate test results from my code when tested with different 
compile options. I cannot make this work.

I guess that I probably should have asked here directly.
If answered I will cross-answer the stackoverflow my self.

latk:
I saw your StackOverflow question, and it motivated me to finally release
the necessary functionality (it had already been implemented for a while).

Please see the documentation on combining/merging coverage data:
https://gcovr.com/en/stable/guide.html#combining-tracefiles
In a nutshell, you create a JSON report with the coverage of each
configuration,
and can then use the -a/--add-tracefile option to merge the JSON files, and
create your desired report format.
This requires that you update to gcovr 4.2.

peter-toft-greve:
@latk This sounds brilliant. Will test as the first thing tomorrow morning. Thanx a mil. More to follow :)

peter-toft-greve:
@latk This is a PURE homerun - thanx a mil!! 
I will answer on StackOverflow and credit you (+reference the documentation)

For sake of completeness - I take it that this cannot be done prior to gcovr 4.2?

latk:
Thank you. Yes, the necessary features (JSON output, merging) were first released with gcovr 4.2.

peter-toft-greve:
@latk  One suggestion for further updates; that 
gcovr --add-tracefile run-*.json
is supported - to avoid to name the *json files

peter-toft-greve:
@latk as promised > I have updated https://stackoverflow.com/questions/58632496/ and credited you. THANX!!!

PR #351, opened 2020-02-15
Extensions and fixes for json_generator

CezaryGapinski:
In this PR I try to:
- solve problem with generating output report when '--output' option is missing for json_generator
- add feature suggested in issue #338 about adding wildcards to '--add-tracefile' option

I need a review for the second feature, because I am not sure whether is correctly implemented or should be done totally different.

### DIFF
--- a/doc/source/guide.rst
+++ b/doc/source/guide.rst
@@ -339,6 +339,11 @@ Next, merge the json files and generate the desired report::
 
     gcovr --add-tracefile run-1.json --add-tracefile run-2.json --html-details coverage.html
 
+You can also use unix style wildcards to merge the json files without
+duplicating :option:`gcovr --add-tracefile`. With this option you have to
+place your pathnames with wildcards in double quotation marks::
+
+    gcovr --add-tracefile "run-*.json" --html-details coverage.html
 
 The gcovr Command
 -----------------
--- a/gcovr/__main__.py
+++ b/gcovr/__main__.py
@@ -38,6 +38,7 @@ from argparse import ArgumentParser
 from os.path import normpath
 from tempfile import mkdtemp
 from shutil import rmtree
+from glob import glob
 
 from .configuration import (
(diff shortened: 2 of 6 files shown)


gcovr/__main__.py
```diff
@@ -259,13 +260,17 @@ def main(args=None):
 def collect_coverage_from_tracefiles(covdata, options, logger):
     datafiles = set()
 
-    for trace_file in options.add_tracefile:
-        if not os.path.exists(normpath(trace_file)):
+    for trace_files_regex in options.add_tracefile:
+        trace_files = glob(trace_files_regex)
```
latk: Do we want to use `glob(..., recursive=True)` in order to allow `**/*.json` style patterns? I see no particularly good use case for this, but I also see no good reason to disallow recursive glob patterns.

doc/source/guide.rst
```diff
@@ -339,6 +339,11 @@ Next, merge the json files and generate the desired report::
 
     gcovr --add-tracefile run-1.json --add-tracefile run-2.json --html-details coverage.html
 
+You can also use wildcards to merge the json files without duplicating
+:option:`gcovr --add-tracefile`. With this option you have to place your
+file name with wildcards in quotation marks::
+
+    gcovr --add-tracefile "run-*.json" --html-details coverage.html
```
latk: This should also be documented in the --help entry for the --add-tracefile option.

gcovr/__main__.py
CezaryGapinski: I can see no reason to disallow `recursive=True` so I have added this option in next version of patch. Thanks for hint!

CezaryGapinski:
@latk Thank you for your review. I was not aware about your fix for missing `--output`. I only noticed `html-rewrite` branch name and I have assumed that there is no changes for json output. I should check it more deeply. Sorry for that :).

PR #356, opened 2020-03-16
Add --exclude-lines-by-pattern to filter out source lines by regex

sinelaw:
Use case:

We have some C code that includes macros such as ASSERT, PANIC, etc. We don't need those lines to be covered.

1. The legacy code base is big and can't be changed easily to include EXCL-style markers
2. Due to preprocessor macro magic going on, it's very difficult to include the EXCL-style markers in the code anyway

It's much easier to simply ignore those lines by some simple regular expression. This pull request adds that capability.


latk:
Thank you for working on this! I totally get why this functionality is needed, and the coverage exclusion mechanism is due for an overhaul anyway.

But I'm not entirely sure whether adding a new mechanism is appropriate? Would it be possible to instead make the `GCOVR_EXCL_LINE` mechanism user-configurable so that they can write down their own patterns in the config file? The minimally invasive change towards that would be to split up the `exclude_line_pattern` to handle LINE patterns separately from START/STOP patterns, and then provide the line pattern through a config variable rather than hardcoding it as a global variable. However, that does require some changes throughout the gcov.py code, e.g. we would have to replace this:

``` python
exclude_line_pattern = re.compile(r'([GL]COVR?)_EXCL_(LINE|START|STOP)')
...

        if exclude_line_flag in line:
            excl_line = False
            for header, flag in exclude_line_pattern.findall(line):
                if self.parse_exclusion_marker(header, flag):
                    excl_line = True
```

with something like:

``` python
exclude_line_pattern = re.compile(r'([GL]COVR?)_EXCL_(START|STOP)')
...

        excl_line = False
        for header, flag in exclude_line_pattern.findall(line):
            if self.parse_exclusion_marker(header, flag):
                excl_line = True

        # check for user-provided line exclusion patterns,
        # defaulting to r'([GL]COVR?)_EXCL_LINE'
        if options.exclude_line.find(line):
            excl_line = True
```

jenswet:
Allowing a regex pattern for line exclusion is a great idea. We have the same problem as @sinelaw. 

@latk I think a new mechanism is a good idea. We could keep the standard solution people are already using and still exclude specific legacy patterns additionally. In my case I would like to exclude all LOG_* macros. It is not easily possible to add some marker in all these lines / would decrease the readability too much.

sinelaw:
@latk I don't think I'll have the time to reimplement this in another way - any chance this is good enough as-is (or with minor modifications)?

sinelaw:
(rebased and pushed unmodified)

latk:
> @latk I don't think I'll have the time to reimplement this in another way - any chance this is good enough as-is (or with minor modifications)?

I'd really like to avoid adding multiple overlapping exclusion mechanisms because that is technical debt down the line. But I think that the changes I have suggested above are fairly minor. Essentially, I'm just asking for removing line exclusion from the legacy exclusion mechanism, and adding the old line pattern as a default to your mechanism.

* In `gcov.py`:

  ``` diff
  -exclude_line_pattern = re.compile(r'([GL]COVR?)_EXCL_(LINE|START|STOP)')
  +exclude_line_pattern = re.compile(r'([GL]COVR?)_EXCL_(START|STOP)')
  ```

* In `gcov.py`, removal of dead paths from the `parse_exclusion_marker()` method and its callsite

* Adding a default value to the --exclude-lines-by-pattern option: `.*[GL]COVR?_EXCL_LINE`

If you (Noam) don't want to do this, I'd be happy if anyone else adopts the pull request. Because this feature is really good, I just don't want it to box us in later.

sinelaw:
I made the suggested changes, fixed a few bugs in the feature itself on the way.

PR #358, opened 2020-03-17
Fix: incorrect path for header can still generate html-details reports

ensky:
Currently still some cases (as added in testcase) caused header file path parsing fails, however these cases is unlikely to (or hard to) fix. It seems that we should by-pass these cases and allow users to generate html-details report even if header file is missing.


latk:
Making the writers more resilient sounds absolutely sensible! Thank you for working on this. However:

* There are some larger pending changes in the html-rewrite branch that I should merge first.
* I agree with writing the complete report even when there was an error, but not with swallowing the exception entirely. I think it would be better to *defer* the exception until after all pages have been written.

Since other stuff is currently more important I won't merge right now, but I want to revisit this idea in the future.

ensky:
Thanks for reviewing! 

In my case, there's many header's location being parsed with errors, so what I did is just print out the warning message to indicate that there's a error in header parsing. It'll print something like:

 `File Generation failed: include/myheader.h, reason=FileNotFoundError: [Errno 2] No such file or directory`

which I found pretty informative. What are your thoughts on this? Will throwing a exception directly after html parsing process better?

latk:
>
> Will throwing a exception directly after html parsing process better?


I think so. If there's an error, the exit code of gcovr should reflect
this. Gcovr is used in some automated builds, and if gcovr cannot create
the coverage report the whole build should fail. Otherwise no one would
notice the problem. I'm more concerned with CI than with users who may or
may not see error messages on their console.

However, there's some precedent for allowing the user to selectively ignore
*some* gcovr errors. E.g. `--gcov-ignore-parse-errors` was necessary as a
stopgap measure to deal with changes of gcov's output format. Something
like `--html-ignore-unknown-files` might be acceptable in this case.
Resilience is good, but swallowing errors must not be the default.

ensky:
> I'm more concerned with CI than with users who may or may not see error messages on their console.

I agree with your concern. Indeed we should return error codes if any error occured.

gcovr/tests/subfolder-includes/Makefile
```diff
@@ -0,0 +1,27 @@
+all:
+	make -C subfolder
+
+run: txt xml html sonarqube json
```
Spacetown, line 4: Only run HTML test because change is only for this format.

gcovr/html_generator.py
```diff
@@ -245,12 +248,14 @@ def print_html_report(covdata, output_file, options):
...
-                    source_row(ctr, line.rstrip(), cdata.lines.get(ctr))
-                )
+        try:
+            with io.open(data['FILENAME'], 'r', encoding=options.source_encoding, errors='replace') as INPUT:
+                for ctr, line in enumerate(INPUT, 1):
+                    data['ROWS'].append(
+                        source_row(ctr, line.rstrip(), cdata.lines.get(ctr))
+                    )
+        except IOError as e:
+            logger.warn('File Generation failed: {filename}, reason={reason}', filename=data['FILENAME'], reason=repr(e))
```
Spacetown, line 258: The HTML doesn't cotain data if source isn't found. Would it be better to use something like `!!! File not found !!!` for the first line and keeep the lines for the rest of the data empty?

ensky: I'll change it to !!! File not found: {filename}, reason={reason}

Please check if this is what you suggested, thx!

ensky:
@Spacetown I rebased, updated the test data and changelog. Please check, thanks!

gcovr/tests/subfolder-includes/reference/gcc-5/coverage.lib.h.4acb6a61176c15c0bede65bfbcd51fe2.html
```diff
@@ -0,0 +1,78 @@
...
+<div class="file-source">
+  <table>
+    <tr>
+      <th align="right">Line</th>
+      <th align="right">Branch</th>
+      <th align="right">Exec</th>
+      <th align="left">Source</th>
+    </tr>
+
+
```
Spacetown, line 68: My suggestion was to add something like this instead of a empty html:
```html
    <tr class="source-line">
      <td class="lineno">-</td>
      <td class="linebranch">
      </td>
      <td class="linecount"></td>
      <td class="src coveredLine">!!! File not found !!!</td>
    </tr>
```

gcovr/html_generator.py
```diff
@@ -382,13 +383,16 @@ def print_html_report(covdata, output_file, options):
...
+        try:
+            with io.open(data['filename'], 'r', encoding=options.source_encoding,
+                         errors='replace') as source_file:
+                lines = formatter.highlighter_for_file(data['filename'])(source_file.read())
+                for ctr, line in enumerate(lines, 1):
+                    data['source_lines'].append(
+                        source_row(ctr, line, cdata.lines.get(ctr))
+                    )
+        except IOError as e:
+            logger.warn('!!! File not found: {filename}, reason={reason}', filename=data['filename'], reason=repr(e))
```
Spacetown, line 395: The message should only contain the text withou exlamation marks because the logge prepends text.
```suggestion
            logger.warn('File {filename} not found: {reason}', filename=data['filename'], reason=repr(e))
```

Also shall a flag be set to remember the error. This flag must be returned by the function and collected in https://github.com/gcovr/gcovr/blob/4240ddcd212276a6b03a00586fdc859024f66391/gcovr/__main__.py#L438 to set the exit code for a generator error.

Spacetown (changes requested):

gcovr/html_generator.py
ensky: Generators currently don't report errors. I change this behavior to try to fit your suggestion. Please check for my latest commits. Thanks!

gcovr/__main__.py
```diff
@@ -425,17 +430,19 @@ def print_reports(covdata, options, logger):
...
+
     for output_choices, generator, on_no_output in generators:
         output = OutputOrDefault.choose(output_choices, default=default_output)
         if output is not None and output is default_output:
             default_output_used = True
             if not output.is_dir:
                 default_output = None
         if output is not None:
-            generator(covdata, output.abspath, options)
+            generator_error_occurred = generator(covdata, output.abspath, options) or generator_error_occurred
```
Spacetown, line 445: To see that the errors are or'ed.
```suggestion
            generator_error_occurred = generator_error_occurred or generator(covdata, output.abspath, options)
```

gcovr/__main__.py
```diff
@@ -290,7 +290,12 @@ def main(args=None):
     logger.verbose_msg("Gathered coveraged data for {} files", len(covdata))
 
     # Print reports
-    print_reports(covdata, options, logger)
+    error_occurred = print_reports(covdata, options, logger)
+    if error_occurred:
+        logger.error(
+            "Error occurred while printing reports"
+        )
+        sys.exit(1)
```
Spacetown, line 298: Exit code 1 is used for command line errors. 7 is the next free one. Where are the codes documented???
```suggestion
        sys.exit(7)
```

gcovr/tests/subfolder-includes/Makefile
```diff
@@ -5,7 +5,7 @@ run: html
 
 html:
 	./subfolder/subfolder/testcase
-	$(GCOVR) -d --html-details -o coverage.html
+	$(GCOVR) -d --html-details -o coverage.html || true
```
Spacetown, line 8: Check the correct exit code and not nly failing.

gcovr/html_generator.py
```diff
@@ -392,13 +393,18 @@ def print_html_report(covdata, output_file, options):
                         source_row(ctr, line, cdata.lines.get(ctr))
                     )
         except IOError as e:
-            logger.warn('!!! File not found: {filename}, reason={reason}', filename=data['filename'], reason=repr(e))
+            logger.warn('File {filename} not found: {reason}', filename=data['filename'], reason=repr(e))
+            data['source_lines'].append(
+                source_row(0, '!!! File not found !!!', None)
```
Spacetown, line 398:
```suggestion
                for ctr in Range(1,max(cdata.lines.keys())):
                    data['source_lines'].append(
                        source_row(ctr, '!!! File not found !!!' if ctr == 1 else '', cdata.lines.get(ctr))
                    )

```

Spacetown (changes requested):
Can you add a missing file with coverage data, e.g. a header with an inline function or a C file?

gcovr/__main__.py
ensky: I wouldn't suggest to do that because it's logically different. If I changed to `generator_error_occurred or generator()` than if error occurred generator won't be called.
```python
>>> def go():
...   print 'asdf'
...
>>> True or go()
True
>>> go() or True
asdf
True
```

Spacetown: Good point. If equating on left side of a or is True, the right site isn't evaluated.
I think a if clause, setting the flag in the body is clearer.

gcovr/__main__.py
ensky: Hi @Spacetown. I don't think there exists a document related to the exit code currently. Maybe create an issue for it?

Spacetown (approved):
LGFM

Spacetown:
Thanks for the effort.

PR #359, opened 2020-03-17
Fix header with relative path coudn't be searched

ensky:
Please refer to the testcase.
If headers were included via compile flag -I, gcov will generate a relative path like ../include/xxx.h, which cannot be detected by gcovr. This commit add support to lookup relative path (related to .o file) to fix this problem.


latk:
Thank you for looking into this. I agree that this heuristic for finding files can make sense, but I'm not sure it is always applicable. AFAIK the paths reported by gcov are relative to the compiler's working directory, but that's not always the same as the directory of the object file! We need to consider the cases (listing the source, include, and build dirs for each case):

1. that everything is in the same directory:  
   a.  `gcc -I ../includes foo.c -o foo.o` (/project/src, /project/includes, /project/src)  
   b. `gcc -I includes foo.c -o foo.o` (/project, /project/includes, /project)

2. that the compiler runs in the source directory but outputs to a build directory:  
   a.  `gcc -I ../includes foo.c -o ../build/foo.o` (/project/src, /project/includes, /project/build)  
   b.  `gcc -I includes foo.c -o build/foo.o` (/project, /project/includes, /project/build)

3. that the compiler runs in the build directory
    a. `gcc -I ../includes ../src/foo.c -o foo.o` (/project/src, /project/includes, /project/build)  
    b. `gcc -I ../includes ../foo.c -o foo.o` (/project, /project/includes, /project/build)

4. that the compiler runs in a separate directory  
    a. `gcc -I includes src/foo.c -o build/foo.o`  (/project/src, /project/includes, /project/build)

I think – but am not sure – that your heuristic would be wrong in cases 2 and 4. That doesn't mean I won't merge your heuristic, but we should order the `guess_source_file_name_heuristics()` in a manner that minimizes false matches.

What are your thoughts on this? At this point you probably have a better understanding about file resolution than me :)

ensky:
Wow, thanks for organizing the four cases so systematically! I think you're totally right about this four cases.

I would suggest that instead to put my heuristic in the first priority, maybe we should just simply put it in the last priority, just before the fail-back?

latk:
>
> I would suggest that instead to put my heuristic in the first priority,
> maybe we should just simply put it in the last priority, just before the
> fail-back?
>

That sounds very sensible. Honestly, I don't know if the fallback can ever
produce a correct/useful result. Your heuristic looks very similar, but
more correct because it doesn't throw away the path information.

ensky:
@latk Thanks for the review! I've pushed a new commit to fix the review.

ensky:
duplicated with #410 and #410 is merged.

PR #361, opened 2020-04-01
Add --no-markers option to ignore exclusion markers

Irfy:
My team needed a way to show true coverage vs. coverage with exclusions in an easy and accessible way. `lcov` already has `--no-markers` so I found it meaningful to implement in the same way for `gcovr` as well.


Irfy:
https://travis-ci.org/github/gcovr/gcovr/builds/669664039 says that everything passes, but it shows as `Pending` here, anybody know what's up?

@latk did you have time to have a look at this? The implementation diff is 19 +- lines, the rest comes from detailed tests.

The main purpose of this PR is to be able to do things like "we're not using this code right now in production", "this is non-critical code", etc. and get coverage numbers for the parts that matter, but still be able to get coverage for the whole thing as well.

Irfy:
I've looked at #356 and here's my opinion:

- The problems these two PRs are trying to solve are *thoroughly orthogonal*.
  - The fact that a part of this PR can be emulated with the other is, IMHO, incidental.
- That you cannot ignore begin/end markers in #356 is not a minor thing -- it's a complete deal-breaker for the actual problem my PR is solving (unless I'm the only one using begin/end markers :) )

In addition to that, the option, spelled exactly as in my PR, is already well-established with `lcov`, so it would be commonly understood if someone is coming from that background. (This is not a big argument, just a nice add-on)

latk:
Ok, that's a reasonable argument.

To merge this PR, it would be necessary to rebase on the current `master` branch and to regenerate the HTML files. It may also be helpful to rename the internal `no_markers` variable to something like `respect_exclusion_markers`, in order to avoid double negations.

Feature partity with lcov/fastcov/grcov is generally desirable, but it's OK if different tools have different strengths and different conventions. It's therefore not necessary to call this option `--no-markers`, but I can't think of any better name.

Irfy:
#356 is giving me some issues with my exclusion marker option. Originally, I did this:

```python
        if respect_exclusion_markers and exclude_line_flag in line:
            for header, flag in exclude_line_pattern.findall(line):
                self.parse_exclusion_marker(header, flag)
```

But now, the new `exclude_lines_by_pattern_regex` mechanism is performing the line exclusion through its default pattern of `[GL]COVR?_EXCL_LINE` immediately afterwards:

```python
        if exclude_lines_by_pattern_regex:
            if exclude_lines_by_pattern_regex.match(code):
                self.excluding.append(False)
```

So when the user doesn't pass `--exclude-lines-by-pattern`, that option is internally active and its functionality overlaps with that of the common exclusion marker mechanism.

Now, when `--no-markers` comes into play, I have to disable the exclusion done by #356 because of the default regex:

```python
        if respect_exclusion_markers and exclude_lines_by_pattern_regex:
            if exclude_lines_by_pattern_regex.match(code):
                self.excluding.append(False)
```

However, this introduces an **artificial** conflict between the two options: one can never combine  `--no-markers` with `--exclude-lines-by-pattern`, because the former has to override the latter -- and that, without there being a technical reason that would make the override absolutely necessary.

In my opinion it would have been better that each exclusion mechanism which is semantically distinct from others is self-contained and enabled or disabled by an option, and not let the semantics of one (explicit exclusion markers) seep into the other (user-defined general line matching for exclusion), because then we cannot control them individually, as is the case now.

In order to achieve this clear mechanism separation, I would have to undo the changes #356 did to `parse_exclusion_marker`, i.e., let the flag `LINE` be handled there and simply return true, and reintroduce the `excl_line` variable which tracks whether the current line should be excluded or not and exclude it once after both checks in `parse_line`.

I'll be pushing these changes for review, with a separate commit for the #356 changes so it can be reviewed on its own merit.

(Also, `--exclude-lines-by-pattern` is in the wrong section -- gcov -- I moved it to gcovr as should be)

latk:
Thank you @Irfy for thinking about the interactions of various options. That is very important because the next release commits us to a design, and your above comment has some good insight.

I disagree with your conclusion though. Yes, `--no-markers` and setting an explicit line exclusion marker is at odds, so it “shouldn't” occur in any reasonable invocation.

But I see `--no-markers` as a kill switch for exclusion markers, whereas `--exclude-lines-by-pattern` is more of a configuration option, similar to how the `--html-title` has no effect for XML output. This matters in particular:

* when setting a line exclusion marker in the config file, but wanting to override it for one invocation
* when using a build system that provides some default command line arguments, but lets you append your own options, e.g. in a Makefile: `gcovr --exclude-lines-by-pattern='//\s*nocover\b' $(GCOVR_OPTIONS)`

I would therefore suggest:

1. please keep the general behaviour of #356 intact
2. have --no-markers disable --excludes-lines-by-pattern as shown in your code snippet

Your other improvements like moving the option to the correct section or your idea to skip regex matches if the line is excluded anyway, those improvements are very good.

If you have other ideas regarding the design of these features I'd like to hear them, in particular how we could expose start/stop markers as configuration options. And now that I'm thinking more about the context of these options, maybe we should rename `--exclude-lines-by-pattern` to `--exclude-line-marker`?

Spacetown (approved):
LGFM.

PR #363, opened 2020-04-19
WIP: add json summary (per-file and global) to the existing --json output

Irfy:
I'm making this PR after personally abandoning #302 as it has 92 commits deviating from gcovr:master (undoing previous changes/fixes in the process).

This PR adds coverage summary to the JSON output, analogous to `--html` output. Each JSON object for a file now looks like this:

    {
        "file": "...",
        "lines": [ ... ],
        "branch_summary": {
            "coverage": <float>,
            "exec": <int>,
            "total": <int>
        },
        "line_summary": {
            "coverage": <float>,
            "exec": <int>,
            "total": <int>
        }
    }

The "branch_summary" and "line_summary" sub-objects are also added to the top-level, containing global coverage numbers, identical to those printed to STDOUT with `--print-summary`.

The change is backwards-compatible (it's a pure extension of current `--json` output).

In a second step, I could move the line information to a new `--json-detail` option, breaking backwards compatibility, but becoming consistent with `--html` vs. `--html-details` options (I prefer this much more over having `--json` behave like `--html-details` and having a `--json-summary` option to compensate).

Please let me know if you prefer this approach over #302 . (I haven't written new or updated existing tests until I see if you're interested in merging this approach)

### DIFF
--- a/gcovr/json_generator.py
+++ b/gcovr/json_generator.py
@@ -10,7 +10,7 @@ import os
 import sys
 import functools
 
-from .utils import presentable_filename, Logger
+from .utils import presentable_filename, Logger, calculate_coverage
 from .coverage import FileCoverage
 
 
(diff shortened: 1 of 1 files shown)


CezaryGapinski:
Why not just upgrade JSON_FORMAT_VERSION to 0.2 and use your proposal as a default JSON?

Have you tried to merge previous JSON file to your new propsed format by `--add-tracefile` option? I am not sure about create new `--json-detail` option. Maybe JSON file should be self-contained to do not making  confusion during merging.

latk:
Thank you for working on this. I like the idea of adding summary information directly into the JSON output, for convenience of downstream users. However:

* I'll give #366 precedence for JSON-summary support
* If we add extra keys `foo` to the normal JSON output, they should be namespaced like `gcovr/foo` in order to maintain forward compatibility with the upstream gcov format.

  Some people disagree with me on this, but I find it easier to put the foot down now rather than having to potentially deal with large-scale breakage in a couple of years. Gcovr is broken enough as it is :)

Spacetown:
Closed because duplicate of #366.

PR #366, opened 2020-04-26
Adding json_summary changes as one commit

opikalo:
This is just #302 rebased into a  single commit, to clean up commit history.

JSON summary output allows pushing basic coverage data for processing using other downstream tools or storing in DB. 

EDIT: After some discussion, we are keeping `--json-pretty`  behavior identical for backward compatibility. If you want pretty summaries, you would need `--json-pretty-summary`


CezaryGapinski:
Hi @opikalo

Why not just use `--json-summary` to report this summary in JSON format and use `--json-summary-pretty` to do this in pretty way and leave `--json` and `--json-pretty` as it is? That would be something similar to `--print-summary`.

opikalo:
@CezaryGapinski: should be good to go.

CHANGELOG.rst
```diff
@@ -62,6 +62,7 @@ Improvements and new features:
    are equivalent. (:issue:`291`)
  - Override gcov locale properly. (:issue:`334`)
  - Make gcov parser more robust when used with GCC 8. (:issue:`315`)
+ - Added -json-sumamry option to generate json summary report. (:issue:`302`)
```
CezaryGapinski, line 65: Small typo here :)

gcovr/json_generator.py
```diff
@@ -31,21 +49,65 @@ def print_json_report(covdata, output_file, options):
...
+# Produce gcovr JSON summary report
+#
+def print_json_summary_report(covdata, output_file, options):
+
+    total_lines = 0
+    total_covered = 0
+
+    json_dict = {}
+    json_dict['current_working_directory'] = options.root
+    json_dict['gcovr/format_version'] = JSON_FORMAT_VERSION
```
CezaryGapinski, line 70: This types of keys gcovr/* like gcovr/format_version were added as a some extension keys which don't exist in intermediate JSON format from GCOV. I suppose your JSON summary has nothing common with the GCOV intermediate JSON, so you can change the gcovr/format_version key to just the format_version (or summary_format_version) = JSON_SUMMARY_FORMAT_VERSION.
I would prefer to separate the JSON_FORMAT_VERSION and the JSON_SUMMARY_FORMAT_VERSION to emphasise that they have nothing common.

gcovr/json_generator.py
```diff
@@ -67,7 +68,7 @@ def print_json_summary_report(covdata, output_file, options):
 
     json_dict = {}
     json_dict['current_working_directory'] = options.root
-    json_dict['gcovr/format_version'] = JSON_FORMAT_VERSION
+    json_dict['gcovr/summary_format_version'] = JSON_SUMMARY_FORMAT_VERSION
```
CezaryGapinski, line 71: What do you think about change:
    json_dict['gcovr/summary_format_version'] = JSON_SUMMARY_FORMAT_VERSION
to:
    json_dict['summary_format_version'] = JSON_SUMMARY_FORMAT_VERSION

without `gcovr` prefix? As it is mentioned in the [User Guide](https://github.com/gcovr/gcovr/blob/master/doc/source/guide.rst#id18) `gcovr/` was used to differentiate between GCOV keys and gcovr keys. This summary doesn't base on some external format, so there is nothing to differentiate.

gcovr/tests/add_coverages/Makefile
```diff
@@ -8,15 +8,15 @@ all:
...
 
 json_foo:
 	./testcase_foo
-	$(GCOVR) -d --json-pretty -o coverage_foo.json
+	$(GCOVR) -d --json --json-pretty -o coverage_foo.json
 
 json_bar:
 	./testcase_bar
-	$(GCOVR) -d --json coverage_bar.json
+	$(GCOVR) -d --json -o coverage_bar.json
```
CezaryGapinski, line 19: It's turned out in this commit 3aa4a34c7fe78ca5081df9e2951a4450e89f3192 output should work even without -o, so maybe it is better to leave at least one test to check correct behaviour for this issue?

CezaryGapinski:
For me this PR after minor changes will look good, but it must be reviewed by @latk. Perhaps it would be nice to add some description in the [User Guide](https://github.com/gcovr/gcovr/blob/master/doc/source/guide.rst).
@Irfy - do you have something do add or change based on your PR #363?

gcovr/tests/add_coverages/Makefile
latk: I touched that file in the html-rewrite branch (now merged). The different JSON reports in that Makefile now exercise different -o/--json combinations:

https://github.com/gcovr/gcovr/blob/1f17d4427c301f2a2a8bad50991f161a0e45aa5c/gcovr/tests/add_coverages/Makefile#L13-L23

gcovr/json_generator.py
opikalo: @CezaryGapinski `json` is notorious about not telling what this payload is. I would rather keep the `gcovr` in place instead of ambiguous `summary_format_version`.

opikalo:
@latk - appreciate a review

gcovr/json_generator.py
```diff
@@ -31,21 +50,65 @@ def print_json_report(covdata, output_file, options):
...
+    def _summarize_file_coverage(coverage):
+        filename = presentable_filename(coverage.filename,
+                                        root_filter=options.root_filter)
+
+        if options.show_branch:
+            total, cover, percent = coverage.branch_coverage()
+            uncovered_lines = coverage.uncovered_branches_str()
+        else:
+            total, cover, percent = coverage.line_coverage()
+            uncovered_lines = coverage.uncovered_lines_str()
```
latk, line 89: Instead of switching between line and branch coverage, the summary can reasonably show both. Compare #376, as they essentially show the same data.

gcovr/json_generator.py
```diff
@@ -31,21 +50,65 @@ def print_json_report(covdata, output_file, options):
...
+#
+# Produce gcovr JSON summary report
+#
+def print_json_summary_report(covdata, output_file, options):
+
+    total_lines = 0
+    total_covered = 0
+
+    json_dict = {}
+    json_dict['current_working_directory'] = options.root
```
latk, line 70: Should the cwd be root, or root relative to the output file, or the actual current working directory? Should it be a relative or absolute path? Depends on the exact intention for this field.

gcovr/tests/oos/Makefile
```diff
@@ -24,6 +26,10 @@ html: coverage.json
...
 	$(GCOVR) -a $< --sonarqube sonarqube.xml
 
+json_summary:
+	build/testcase
+	$(GCOVR) $(GCOVR_TEST_OPTIONS) -d --json-summary -o summary_coverage.json
+
 clean:
 	rm -f build/*
-	rm -f coverage*.* sonarqube*.*
+	rm -f coverage*.* sonarqube*.* summary_coverage.json
```
latk, line 35: reference file is not checked in

gcovr/json_generator.py
opikalo: I understand the idea, and the data is essentially right there; however, currently, the text/tabulated summary does not show branch. Are we departing from that? What should be the behavior of the summary with and without show_branch option?

gcovr/json_generator.py
opikalo: I think I placed it thinking that it will allow to differentiated multiple analysis from different trees; should we rename it to be `root` to be consistent?

gcovr/tests/oos/Makefile
opikalo: Good catch, readding it back.

gcovr/json_generator.py
latk: The text report switches between line/branch coverage for space reasons, not because this is fundamentally useful. For a machine-readable report, text width does not matter. Including this data is effectively zero cost but makes the summary much more useful.

It may be useful to use the same field names as the CSV report (#376) uses for columns.

gcovr/json_generator.py
latk: One of the problems we have with gcov is that it doesn't make quite clear which paths are relative to what, so I'd like to avoid making similar problems for downstream. I think that relevant designs could include:

* all file paths are relative to the summary report location, and no root/cwd is needed; or
* there is a root or cwd field that is relative to the summary report location, and the reported file paths are relative to the root – similar to the HTML `<base>` element.

  ```suggestion
      json_dict['root'] = os.path.relpath(options.root, output_file)
  ```

Both designs have things that are in their favour, but I think that most use cases are more interested in the path relative to root, not relative to the report.

In any case, it should be possible to correctly interpret the report without having to know how gcovr was invoked.

CezaryGapinski:
@opikalo Can you rebase your branch on the current master? For me `--json-summary` feature is really cool, and it will be great to merge it. I have noticed something weird  in the [commits history](https://github.com/gcovr/gcovr/pull/366/commits). It is hard to noticed what has changed :). I have tried to create your [changes](https://github.com/CezaryGapinski/gcovr/commits/json_summary_squashed) on my fork, so you can take a look if you like it or arrange it differently and of course remove me as an committer :). I know it is a small change, but it helps to keep a linear history.

PR #367, opened 2020-04-26
Add html and json templates

opikalo:
> @latk: If you want to do me a huge favour and have an hour to spare: please check out the html-rewrite branch and regenerate the HTML test reference files (and add JSON reference files). They were deleted for a large refactoring. Go into each test directory, run the Makefile, and relevant generated files into the reference directory (might have to create it). That's the next step that I've been procrastinating.

This should do it. I left the regeneration code in place: I can remove it too. I think I did this right. There is a small change in scraping regexp to accommodate new HTML rewrite.


latk:
Thank you so much! Adding that regeneration code is elegant. The CI failures are to be expected since the underlying code is probably slightly broken, but those don't hinder merging this PR. I'll do it as soon as I'm back at my development machine.

PR #372, opened 2020-05-08
Set make variable TEST_OPTS as environment variable inside docker.

Spacetown:
To get the full assertion text or run a specific test you have to set TEST_OPTS. This is not possible if test run inside the docker container.
The option is set as environment variable which is used by the make inside docker.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,8 +49,10 @@ test:
 doc:
 	cd doc && make html O=-W
 
+docker-qa: export TEST_OPTS := $(TEST_OPTS)
+
 docker-qa: | docker-qa-build
-	docker run --rm -v `pwd`:/gcovr $(QA_CONTAINER)
+	docker run --rm -e TEST_OPTS -v `pwd`:/gcovr $(QA_CONTAINER)
 
 docker-qa-build: admin/Dockerfile.qa requirements.txt doc/requirements.txt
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -50,7 +50,7 @@ doc:
 	cd doc && make html O=-W
 
 docker-qa: | docker-qa-build
-	docker run --rm -v `pwd`:/gcovr $(QA_CONTAINER)
+	docker run --rm -e TEST_OPTS="$(TEST_OPTS)" -v `pwd`:/gcovr $(QA_CONTAINER)
```
latk, line 53: While this is certainly an improvement, I foresee shell quoting issues when the TEST_OPTS involve more complicated arguments, for example `make docker-qa TEST_OPTS='-k "xml and simple"'`. I think this can be avoided by using a target-specific export statement `docker-qa: export TEST_OPTS := $(TEST_OPTS)`, then:

```diff
-	docker run --rm -e TEST_OPTS="$(TEST_OPTS)" -v `pwd`:/gcovr $(QA_CONTAINER)
+	docker run --rm -e TEST_OPTS -v `pwd`:/gcovr $(QA_CONTAINER)
```

Spacetown: I've overlooked the option to supress the value in the command line of docker to use the current host value. Updated PR.

PR #373, opened 2020-05-09
Enable filter option in add JSON tracefiles mode

CezaryGapinski:
The main goal of this PR is to enable filtering when we are in mergin JSON mode.

I have noticed problem with `Branch Id` for generating detailed html output from JSON. For my PR problem stop exists after merge `rewrite-html` branch, but I think it is just because something has changed with the `Branch Id` enumerations. I think now it starts from 0 instead of 1, but it is not the same `Id` as in the real gcov file.

Here is src#main.cpp.gcov file for filter-relative-lib test.

```
        -:    0:Source:src/main.cpp
        -:    0:Graph:/home/cg/repos/gcovr/gcovr/tests/filter-relative-lib/project/main.gcno
        -:    0:Data:/home/cg/repos/gcovr/gcovr/tests/filter-relative-lib/project/main.gcda
        -:    0:Runs:5
        -:    0:Programs:1
        -:    1:bool no(bool);
        -:    2:int yes(int);
        -:    3:
function main called 5 returned 100% blocks executed 71%
        5:    4:int main(int argc, char *argv[])
        -:    5:{
        5:    6:   if (no(true)) {
call    0 returned 5
branch  1 taken 0 (fallthrough)
branch  2 taken 5
    #####:    7:      return yes(7);
call    0 never executed
        -:    8:   }
        -:    9:   else {
        5:   10:      return yes(10);
call    0 returned 5
        -:   11:   }
        -:   12:}
```

We have branches with `Ids` 1 and 2 (because Id = 0 is for call function) but after `html-rewrite` these values are replaced to 0 and 1.

Based on the last example from [gcov](https://gcc.gnu.org/onlinedocs/gcc/Invoking-Gcov.html#Invoking-Gcov) depends on the source code the branches `Ids` not necessarily starts from 0 or 1. It depends what is happening in the particular line of code.

You can find the result for incorrect behaviour on my `apply_filter_for_json_merge_before_html_rewrite branch` [here](https://travis-ci.com/github/CezaryGapinski/gcovr/jobs/330717331).

I am not sure if it is a real problem (is anybody interested `Branch Id` in the html or JSON output?), maybe it can be omitted in the html output. I didn't find that is really stored even for intermediate GCOV JSON.


gcovr/json_generator.py
```diff
@@ -71,6 +72,20 @@ def gcovr_json_files_to_coverage(filenames, covdata, options):
             file_path = os.path.join(
                 os.path.abspath(options.root),
                 os.path.normpath(gcovr_file['file']))
+
+            # Return if the filename does not match the filter
+            # Return if the filename matches the exclude pattern
```
Spacetown, line 77: Comment should be moved to if clause and Return can be replaced with ignore.

CezaryGapinski: Thanks! It was copy-paste error from `gcov.py` file :).

CezaryGapinski:
Rebased and push forced on the latest master branch.

PR #375, opened 2020-05-14
fix wrong details html filename in windows

sgheppy:
in windows html details files created like sysrcfile.C:_restofpath.html


latk:
Thanks for finding and fixing this issue! Cross platform path handling is awful :joy:

Just two potential enhancements:

* would you like to add your name to the AUTHORS file? 
* could you add a comment near your change to explain for future readers *why* escaping `:` is necessary? Usually it would be better to create a test case, but here that might be difficult.

sgheppy:
> Thanks for finding and fixing this issue! Cross platform path handling is awful 😂
> 
> Just two potential enhancements:
> 
> * would you like to add your name to the AUTHORS file?
> * could you add a comment near your change to explain for future readers _why_ escaping `:` is necessary? Usually it would be better to create a test case, but here that might be difficult.

When using gcovr in windows the _make_short_sourcename function in html_generator module return something like that: 
`C:\Work\aq_space\BlueNRG_PSLIB_v3x_Test_Component\PYDTM_LLC\eclipse_proj\gcovr_files\llc_coverage.C:_Work_aq_space_BlueNRG_PSLIB_v3x_Test_Component_stack_repo_Stack_lib_core_csp_llc_src_LL_scheduler.c.html`.
I don't know with no error is generated when windows create file with this path but that file is create and readable by browser but isn't reachable by file managers.

I'll implement a test case for that.

Thank you for great tool.

Spacetown:
> I don't know with no error is generated when windows create file with this path but that file is create and readable by browser but isn't reachable by file managers.
You can't create a file with a : in windows, see [Naming Files](https://docs.microsoft.com/de-de/windows/win32/fileio/naming-a-file).

gcovr/tests/test_html_generator.py
```diff
@@ -0,0 +1,22 @@
...
+                                                     ('C:\\gcovr', 'C:\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\source.c'),
+                                                     ('C:/gcovr', 'C:\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\source.c'),
+                                                     ('C:/gcovr_files', 'C:\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\other_dir\\project\\source.c')
+                                                     ])
+@pytest.mark.skipif(sys.platform != 'win32', reason="not in Win31")
```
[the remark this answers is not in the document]
sgheppy: Definitely a typo. Thank you

gcovr/tests/test_html_generator.py
```diff
@@ -0,0 +1,22 @@
...
+def test_windows__make_short_sourcename(outfile, source_filename):
+
+    result = _make_short_sourcename(outfile, source_filename)
+
+    assert result.find(':') < 0 or (result.find(':') == 1 and result.find('C:') == 0)
```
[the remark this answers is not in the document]
sgheppy: Thank you for your remarks. I'm agree with you, this is crearer.

sgheppy:
I' found another error, in case of using _combination of tracefiles_ feature the value of parameter a `filename` is like `/dir1/dir2/source.c`  in windows too.
The char '/' is not escape in windows.
Can I add that fix in this PR?

latk:
Oh, good catch! Yes, you can update the PR to fix this as well.

gcovr/html_generator.py
```diff
@@ -373,6 +376,7 @@ def _make_short_sourcename(output_file, filename):
         # filename if it exceeds common filesystem limitations
         if len(os.path.basename(sourcename)) < 256:
```
Spacetown, line 377: The limit is 260 characters (MAX_PATH).

sgheppy: Thank you for pointing.
Maybe it's needed to consider all path and not only sourcename.
What do you think?

Spacetown: All paths should respect this limit. What about using only the basename of the file extended by the md5sum of the whole path e.g.:
`path/to/source.c` --> `source.c.883f9ee88c14d9c77addd640015ee77d.html`

sgheppy: There is already that:
```python
        longname_hash = "_" + hex(zlib.crc32(longname.encode('utf-8')) & 0xffffffff)[2:]
        longname = longname[(len(sourcename) - len(longname_hash)):]
```
Isn't good enough?

Spacetown: I suggested to always use the naming schema instead of removing the known bad characters and use a hash if the name gets to long.

sgheppy: Sorry I didn't undestand that before.
I'm agree with you, i think is the best approach to avoid path errors.

Would you like me to do it?

Spacetown: I think @latk should decide this. I also prefer md5 instead of crc32.

sgheppy:
> Seems to be easier and robuster than removing known bad characters.

but a lot of tests fail, as soon as possible i try to fix those fails

Spacetown:
> > Seems to be easier and robuster than removing known bad characters.
> 
> but a lot of tests fail, as soon as possible i try to fix those fails

You can use the opion ``

> > Seems to be easier and robuster than removing known bad characters.
> 
> but a lot of tests fail, as soon as possible i try to fix those fails

You can use the option `TEST_OPTS=--generate_reference` and remove the old file if the content is equal. The main file can be updated with `TEST_OPTS=--update_reference`.

sgheppy:
sorry but i don't understand. 
if i type

```CC=gcc CXX=g++ GCOV=gcov PYTHON=python TEST_OPTS=--generate_reference make test``` 
I've got that error:
```ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: --generate_reference
  inifile: None
  rootdir: c:\Users\antonio\Documents\GitHub\gcovr```

Spacetown:
Rebase your branch first. This option is new. In old version you have to modify the init value in the script.

sgheppy:
> > > Seems to be easier and robuster than removing known bad characters.
> > 
> > 
> > but a lot of tests fail, as soon as possible i try to fix those fails
> 
> You can use the opion ``
> 
> > > Seems to be easier and robuster than removing known bad characters.
> > 
> > 
> > but a lot of tests fail, as soon as possible i try to fix those fails
> 
> You can use the option `TEST_OPTS=--generate_reference` and remove the old file if the content is equal. The main file can be updated with `TEST_OPTS=--update_reference`.

I think that create difficult for testing.
I think the md5 calculate are different for windows case and for linux case due to the different path strings.

Spacetown:
What about normalizing? Make relative to root and change path separateor to slash?

sgheppy:
Is that ok?

Spacetown:
I'll try to check it this week.

Spacetown:
Thanks for this fix.

sgheppy:
Thank you for your patience

PR #376, opened 2020-05-26
Add --csv output

sinelaw:
This adds a summarized CSV output format.
It includes only the totals/covered/percent per file, without details on the covered lines/branches.

### DIFF
--- a/gcovr/csv_generator.py
+++ b/gcovr/csv_generator.py
@@ -1,0 +1,47 @@
+# -*- coding:utf-8 -*-
+
+# This file is part of gcovr <http://gcovr.com/>.
+#
+# Copyright 2013-2018 the gcovr authors
+# Copyright 2013 Sandia Corporation
+# This software is distributed under the BSD license.
+
+import sys
+import csv
+
+from .utils import sort_coverage, presentable_filename
+
+
+def print_csv_report(covdata, output_file, options):
+    """produce gcovr csv report"""
+    if output_file:
+        with open(output_file, 'w') as fh:
+            _real_print_csv_report(covdata, fh, options)
+    else:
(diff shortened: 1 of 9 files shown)


gcovr/csv_generator.py
```diff
@@ -0,0 +1,47 @@
...
+        if options.show_branch:
+            total, cover, percent = coverage.branch_coverage()
+        else:
+            total, cover, percent = coverage.line_coverage()
+        return (filename, cover, total, percent)
+
+    writer = csv.writer(OUTPUT)
+    writer.writerow('filename', 'covered', 'total', 'percent')
+    for key in keys:
+        writer.writerow(_summarize_file_coverage(covdata[key]))
```
latk, line 47: instead of switching between line/branch coverage summaries, couldn't we show all? I'd add the following columns:

* filename
* lines covered
* lines total
* lines percent
* branches covered
* branches total
* branches percent

We should also be explicit about how the percent fields are represented. If it's called percent, does it have a `%` sign? Or do we show fractions like 1/10 = 10% = `0.1`?

sinelaw: all good points. push-forced fixes to export all of those, and changed the percentages to [0,1.0] range

latk:
Looks good! :+1: 

Now all that is missing is fixing the linter error from the Travis build, and a teeny tiny test case (add another known format in test_gcovr.py, update a single test Makefile to output CSV, and add a reference file)

sinelaw:
@latk that wasn't so much fun (windows CR/LF issues) but it passes now

latk:
Thanks for working on this!

CSV is useful, but of course a rather underspecified format so such problems aren't entirely surprising. It could be that in the future, we might need to expose CSV formatting options?

PR #377, opened 2020-05-27
Configure tab size for HTML source page #277

Spacetown:


gcovr/templates/style.css
```diff
@@ -241,6 +241,9 @@ span.uncoveredLine
 
     font-family: monospace;
     white-space: pre;
+
+    tab-size: 3;
+    -moz-tab-size: 3;
```
latk, line 246: I respect both sides in the tabs vs spaces debate, but `tab-size: 3` feels like it would make nearly no one happy :)

If we're going to set a default, maybe `tab-size: 4`? Should we introduce a `--html-tab-size` option, or is that a YAGNI feature right now?

Btw it also seems like a .DS_Store file made it through the gitignore?

Spacetown: I'll add an option for this. Shall we add the CSS only if the option is set?
I'll remove the .DS_STORE and add this to .gitignore.

latk: Since the default tab size is `8`, it should always be set explicitly. The Linux kernel is the only project I know that assumes `8`, everyone else seems to default to something in the range 2–4. Using `4` is probably a sane default.

Issue #296, opened 2019-02-14
HTML Title inconsequential


dscole:
First I would say I know almost nothing about html. Hopefully, what I say still makes sense :)

It appears the argument I pass to `--html-title` ends up being set in the html report's `<head>` section's title, not the `<body>`'s title. The effect I'm seeing is that the `--html-title` argument controls the name of the tab in my browser, but the report itself still has the title of `GCC Code Coverage Report`.
Usually my browser tabs are not even wide enough to even show the full title. I think it makes more sense for the `--html-title` to set the title of the report I see, rather than then tab name.

I tried looking how I would make this change myself in the code, but there are just so many instances of `GCC Code Coverage Report` in the repo, that I didn't know which I actually need to change.

latk:
Hi, I actually have a private branch where I've started cleaning up the
HTML templates (mostly removing the table-based layout and Internet
Explorer compatibility things). Instead of changing text occurrences
manually I instead run a Perl script with some regexes over the test suite …

Your point that the title is pointless is very good, and I'll try to fix
that the next time I touch Gcovr. However, I cannot give you an ETA when
that will be. It's great that you've looked for ways to fix this yourself
(thank you!) but in this particular case I'd prefer to do it myself to
prevent merge conflicts.

>

dscole:
Yeah absolutely, I prefer not touching anything html since I know so little about it.

Spacetown:
Shall the "GCC Code Coverage Report" be replaced or extended by the given title?

Spacetown:
We have two defaults for the title:
- Title of HTML page. https://github.com/gcovr/gcovr/blob/1f17d4427c301f2a2a8bad50991f161a0e45aa5c/gcovr/configuration.py#L595
- Heading in HTML report. https://github.com/gcovr/gcovr/blob/1f17d4427c301f2a2a8bad50991f161a0e45aa5c/gcovr/templates/base.html#L19
I think there should only be one title to keep the template simple. I prefer changing the default in the configuration.py to the value of the heading in the template.

latk:
I have no idea why the title ever said “Head”. Ideally, the HTML title would just default to `GCC Code Coverage report`, and would provide both the `<title>` and the `<h1>`. Of course, this hasn't been done yet because that would require all the HTML reference files to be updated.

Spacetown:
Before I introduced the '--title' option the title was hardcoded to "Head", this was the reason for using this as a default. But I'll change the default and update the test data.

[...]

PR #378, opened 2020-05-27
Use HTML title also for report heading (#296)

Spacetown:
- Default for option --title changed to "GCC Code Coverage Report".


latk:
Brilliant, thanks!

Issue #374, opened 2020-05-09
Extend tests to use an unified diff instead of a simple assert


Spacetown:
It would be easier to check the tests if the output is an unified diff with a context ov several lines instead of a full diff.
If the test fails the clean should be skipped for easier update of the reference. The output directories should be ignored by git.

latk:
I fully agree with all of these things. Making it easier to update the reference files is very important. 

Some context/background:

Currently HTML is just compared as a string, and the diff is produced by pytest. The diff is affected by pytest's verbosity setting: too high and the entire string is shown, too low and the diff may be abbreviated. Maybe there's a pytest plugin or something that produces a proper diff.

I'd actually prefer a proper HTML diff that ignores whitespace etc, but I haven't found a good solution. LXML has basic support for HTML diffs, but seems difficult to integrate. We currently use a similar diffing approach for XML files via the `pyutilib` library (by gcovr's original authors at Sandia) but I want to get rid of that depencency (large dependency for little value).

Feel free to change the gitignore settings, as long as the reference directories are not ignored. IIRC this could require a pattern such as:

``` .gitignore
# in gcovr/tests/.gitignore
# ignore all HTML files
*.html
# except in reference directories
!*/reference/*
```

Spacetown:
Currently there is used following code for the check:
https://github.com/gcovr/gcovr/blob/1f17d4427c301f2a2a8bad50991f161a0e45aa5c/gcovr/tests/test_gcovr.py#L232-L233

The output comes from the assert. What about using difflib and assert that the result of the diff is "equal"?

Spacetown:
The .gitignore works as suggested. I've also added CSS files and moved the unit test parts from the top level .gitignore to gcovr/tests/.gitignore.
I've added two options:
- `--skip_clean`: Skip the clean step after the compare (keep the output data).
- `--update_reference`: Copy the test result to the reference if they differ.

I prefer out of source builds for easier .gitignore but this will be a big change.

latk:
Resoled via #379 :tada:

PR #379, opened 2020-05-27
Extend tests to use an unified diff in the assert #374

Spacetown:
- Add option --update_reference to update the reference data if there are differences.
- Add option --skip_clean to keep test output.

### DIFF
--- a/gcovr/tests/test_gcovr.py
+++ b/gcovr/tests/test_gcovr.py
@@ -8,18 +8,18 @@ import re
 import shutil
 import subprocess
 import sys
+import difflib
 
 from pyutilib.misc.pyyaml_util import compare_repn as compare_xml
 from pyutilib.misc.xmltodict import parse as parse_xml
-
-
-GENERATE_REFERENCE = False
 
 python_interpreter = sys.executable.replace('\\', '/')  # use forward slash on windows as well
 env = os.environ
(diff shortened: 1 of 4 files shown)


gcovr/tests/test_gcovr.py
```diff
@@ -232,15 +240,20 @@ def test_build(compiled, format, available_targets):
...
         with io.open(reference_file, encoding=encoding) as f:
             reference = scrub(f.read())
 
         if assert_equals is not None:
             assert_equals(coverage, reference)
         else:
-            assert coverage == reference, "coverage={}, reference={}".format(
-                coverage_file, reference_file)
+            diff_out = list(difflib.unified_diff(reference.splitlines(keepends=True), coverage.splitlines(keepends=True), fromfile=reference_file, tofile=coverage_file))
+            len_diff_is_zero = True if len(diff_out) == 0 else False
```
latk, line 252: Explicitly listing true/false seems more complicated than just using the condition. Consider:

```suggestion
            diff_is_empty = len(diff_out) == 0
```

Spacetown: I'll change this and add an option '--generate-reference' which will only generate non existing references.

Spacetown: The option `--generate-reference` only creates non existing references.
The option `--update-reference` updates all existing references if there are changes in the normalized files.

PR #380, opened 2020-05-29
Add option --html-css for user defined styling.

Spacetown:
Add optional feature from #277. Given CSS is not rendered by Jinja and replaces the rendered data.
Added function to check given input files in command line and return the absolute path. This can also used for other options where the file must exist.


gcovr/configuration.py
```diff
@@ -38,6 +38,18 @@ def check_percentage(value):
...
 
 
+def check_inputfile(value):
+    r"""
+    Check that the input file is present. Return the full path.
+    """
+
+    if not os.path.isfile(value):
+        raise ArgumentTypeError(
+            "Element {value} is not a file.".format(value=value))
```
latk, line 48: The word “element” might be confusing. I'd rephrase the message like this to guide the user in the correct direction:
```suggestion
            "Should be a file that already exists: {value!r}".format(value=value))
```

Showing the `!r` repr is useful if the filename contains trailing spaces or nonprintable characters

gcovr/configuration.py
```diff
@@ -587,6 +599,14 @@ def choose(_cls, choices, default=None):
         default=None,
         const=OutputOrDefault(None),
     ),
+    GcovrConfigOption(
+        "html_css", ["--html-css"],
+        group="output_options",
+        type=check_inputfile,
+        metavar="CSS",
+        help="Use given CSS stylesheet for the HTML report. Default is '{default!s}'.",
+        default=None
```
latk, line 608: If default is None, the current help message might be confusing. Depending on the exact role of this style sheet, the message should possibly say

> Override the default style sheet for the HTML report.

gcovr/tests/test_gcovr.py
```diff
@@ -105,7 +109,7 @@ def compiled(request, name):
         assert run(['make', 'clean'], cwd=path)
 
 
-KNOWN_FORMATS = ['txt', 'xml', 'html', 'sonarqube', 'json', 'csv']
+KNOWN_FORMATS = ['txt', 'xml', 'html', 'css', 'sonarqube', 'json', 'csv']
```
latk, line 112: I'm not a fan of introducing a `css` output format for the tests since this makes the test suite slower, and the Makefiles become unintuitive. I understand that you're doing this in order to set an `OUTPUT_PATTERN`. But I think it would be better to expand the output patterns mechanism to allow for multiple patterns per type?

We need functionality to check CSS files in order to test the existing --no-html-self-contained option. But for this PR we can just output self-contained HTML files, without writing external CSS?

gcovr/tests/html-css/Makefile
```diff
@@ -0,0 +1,17 @@
+all:
+	$(CXX) -fprofile-arcs -ftest-coverage -fPIC main.cpp -o testcase
+
+run: html css
+
+html:
+	./testcase
+	$(GCOVR) -d --html-details --html-css=style.css -o coverage.html
```
latk, line 8: Test can be minimized by avoiding --html-details, or by setting --html-self-contained
```suggestion
	$(GCOVR) -d --html --html-css=style.css -o coverage.html
```

gcovr/tests/html-css/style.css
```diff
@@ -0,0 +1,344 @@
+:root { font-family: sans-serif; }
```
latk, line 1: Test can be minimized by replacing custom CSS with a placeholder like `/* custom CSS */` – anything as long as it's clearly distinct from the default CSS. If we decide to turn custom CSS into Jinja templates we'd also need to include an example variable somewhere.

gcovr/configuration.py
Spacetown: Ok.

gcovr/configuration.py
Spacetown: Ok.

gcovr/tests/test_gcovr.py
Spacetown: Ok. I'll create a seperate PR to use a list of patterns.

gcovr/tests/html-css/Makefile
Spacetown: Ok.

gcovr/tests/html-css/style.css
Spacetown: Ok.

Spacetown:
> * should the CSS file _override_ or _augment_ the default styles?
>   I have no preference, but both could be useful? Overriding seems more general, and should likely be prioritized as you have done.

I also think overriding is more general. The user will always copy the default stylesheet and adapt it.

> * should the CSS file be used as-is, or should it be interpreted as a Jinja template? That would allow access to the various variables like colours or tab-size.
>   I think that if the CSS overrides the default then it should be a Jinja template to have access to the config variables. If it merely amends the CSS that might not be necessary, assuming a long-term plan to expose the config variables as CSS variables at some point in the future.
>   Executing user-provided templates _is_ security-relevant, but less so than --gcov-executable so that would be allright.

Then I first have to read the documentation of Jinja :-). I didn't use a template because I thought that the user will not use theexisting colors and can hardcode the tabsize. This are the only variables at the moment.

gcovr/tests/test_gcovr.py
Spacetown: #383

Issue #346, opened 2019-12-11
Early fail when HTML details output directory does not exist


0ge:
If `gcovr` is run with `--html-details`, but the designated output directory does not exist, gcovr will run the analysis (which may take some time). It will only then discover that the output directory does not exist, and fail with `IOError: [Errno 2] No such file or directory`.

From a user perspective, it would be great if this was checked prior to the analysis. It may need some consideration if multiple output formats are specified.

latk:
Absolutely, it would be a good idea to check up front that the output files can be written, without already creating them.

However, I'm not sure whether this can be done in a cross-platform manner. On Unix, I'd check whether the target directory exists and that the write-bit is set in the mode of the target directory, as retrieved by the stat() function. Maybe just checking that the target directory exists would be sufficient.

I'd be glad to merge a pull request with a best-effort check that does something like:

``` python
# pseudocode
for outfile in [options.output, options.html, options.html_details, options.xml, ...]:
    if outfile is empty:
       continue
    if not writeable(dirname(outfile)):
        error("directory {} is not writeable", dirname(outfile))
```

PR #382, opened 2020-05-31
Early fail when output can not be created #346

Spacetown:
The options for output files are checked if the file is writable and the parent directory is writable.
Checks extended woth output files in non existing folders. Don't now how to check a write protected folder.


gcovr/configuration.py
```diff
@@ -50,6 +50,30 @@ def check_inputfile(value):
...
+    """
+
+    if value is not None:
+        if os.path.isfile(value) and not os.access(value, os.W_OK):
+            raise ArgumentTypeError(
+                "File {value} exist but is not writable".format(value=value))
+
+        dirname = os.path.dirname(value)
+        if dirname == '':
+            dirname = '.'
```
latk, line 65: I thought hard about this line yesterday. This is correct for command line arguments. But this is not correct when the path is read from the config file.

This is not your fault, but the current code is buggy. We special-case filters in the argument parsing code to fix up relative paths, maybe we need something similar for files?

Should we fix it as part of this PR, or separately?

gcovr/configuration.py
```diff
@@ -50,6 +50,30 @@ def check_inputfile(value):
...
+        if not os.path.isdir(dirname):
+            raise ArgumentTypeError(
+                "Directory {dirname} does not exist".format(dirname=dirname))
+
+        if not os.access(dirname, os.W_OK):
+            raise ArgumentTypeError(
+                "Directory {dirname} is not writable".format(dirname=dirname))
```
latk, line 72: Error messages can be designed along different dimensions:

What the error message tells us:

* stating an error, negative: The directory is not writable.
* positive, describing an expectation or pointing towards a solution: The directory has to be writable.

How the error message is phrased:

* natural phrasing, easier to read: The directory foo/bar has to be writable.
* structured, easier for online searches: The directory has to be writable: foo/bar

I have a mild preference for positive+structured, but won't insist. This is really not important compared to other stuff.

Spacetown: I agree with you and will change this.

gcovr/configuration.py
Spacetown: This is not only for configuration files. Another solution here is to make the path absolute and return this path as it is done for input files.
Paths in config files should be treated relative to the config file and changed to an absolute path while reading the config file. I'll check this with this PR.

Spacetown:
I've tried to add a test with a temporary file and using chmod but it didn't work. Maybe this is because I'm running the test in Docker.

gcovr/configuration.py
Spacetown: Special treatment of check_input_file, check_output_file and OutputOrDefault. The new test uses the --html-css and --output options in the configfile.

gcovr/tests/test_args.py
```diff
@@ -79,6 +79,41 @@ def test_invalid_objdir(capsys):
...
 
+def helper_test_non_existing_directory(capsys, option):
+    c = capture(capsys, [option, 'not-existing-dir/file.txt'])
+    assert c.out == ''
+    assert 'The output directory does not exist: \'not-existing-dir\'' in c.err
+    assert c.exception.code != 0
+
+
+def test_non_existing_directory_output(capsys):
+    helper_test_non_existing_directory(capsys, '--output')
```
latk, line 90: `@pytest.mark.parametrize(…)` to the rescue? That way the helper could serve as the single test function.

latk:
Testing permissions is going to be somewhat nonportable, at least when Windows is involved. Otherwise, `chmod 444 file` should create a file that is readable and deletable but not writable. Unfortunately, the tests in Docker run as root and ignore the missing write permission. 

But given the difficulties involved, it's OK if the test is skipped. Gcovr isn't that critical, so some amount of testing in “production” is fine :)

gcovr/tests/test_args.py
Spacetown: Ok.

Spacetown:
What about to change to the python way. Instead of doing all the checks for the permission try to write the file and unlink it?

Spacetown:
I'm trying to solve the problem found within the new config test. Jinja fails if the template is in a sub directory, in docker it works.

gcovr/tests/test_args.py
Spacetown: I've checked the `paramtrize` but didn't get it to run.

gcovr/configuration.py
Spacetown: After changing to the python way we have no cover problem anymore and only one error message containing the OS error message:
```python
raise ArgumentTypeError("Could not create output file {value!r}: {error}".format(value=value, error=e.strerror))
```

gcovr/tests/test_args.py
latk: ok, no problem

latk:
> What about to change to the python way. Instead of doing all the checks for the permission try to write the file and unlink it?

Well, there are tradeoffs. WIth the Python Way you presumably mean “it's better to ask forgiveness than permission”. That's generally sensible, but is also what we were doing previously: just try to create the report, and if it fails, it fails. Here, our goal is not to create the report, but to quickly alert the users about problems.

* checking write permission and existence of directories
  (-) complicated
  (+) non-destructive
* open the output file and delete afterwards
  (+) simple
  (-) has the effect of deleting the file

This differs in the outcome if gcovr crashes between the test and writing the report. I think there's a 10% chance of someone complaining about this in the future, but the simple destructive check is not incorrect.

> Should this be solved in this PR or should I create a separate one?
> --> Solved in this PR by replacing FileSystemLoader with FunctionLoader for user template.

Thanks for the fix! Jinja is geared towards a web context where there are different security concerns than we have, so apparently it takes some extra work to load arbitrary files.

Spacetown:
> * open the output file and delete afterwards
>   (+) simple
>   (-) has the effect of deleting the file

In the tools we write at work it's always done this way. The users often don't read error messages or check the exit code. If there is a old output from the last run it's used.
Therefore we first delete all output products.

Should I change back to the non-destructive way and we live with the uncovered branches?

latk:
I've merged as-is. The tradeoff between different approaches relates to the concept of exception safety e.g. in C++ programming: things are safe if we end up in a valid state (and deleting files is valid), but under a stronger exception safety guarantee changes would be atomic: the attempted action either fails or produces an effect.

Since gcovr isn't a concurrent database system and since we don't make particular guarantees, deleting files during the check is fine.

Issue #329, opened 2019-08-28
Headers containing definitions are filtered in out-of-source build


mjvankampen:
It seems that headers that contain definitions (such as templates) that are included in other cpp files get a path in gcov that is lower-case on windows (not sure on other OS). This means they do not match the root path anymore. Could the root filter be case-insensitive or do you feel this is an issue with gcov/gcc (6.3.0) and it should be fixed there? For me this seems like the directory is matching so this should not be filtered.

```
Finding source file corresponding to a gcov data file
  currdir      C:\Soft_Dev\590008\JLSCONTROL\JLSOCSOURCE\build
  gcov_fname   C:\Users\mjvk\AppData\Local\Temp\tmp00sj52vb\c~#soft_dev#590008#jlscontrol#jlsocsource#master#mastercommon#include#mastercommon#votercoremessages.hpp.gcov
               ['        -', '    0', 'Source', 'c:/soft_dev/590008/jlscontrol/jlsocsource/master/mastercommon/include/mastercommon/votercoremessages.hpp\n']
  source_fname C:\Soft_Dev\590008\JLSCONTROL\JLSOCSOURCE\build\core\src\CMakeFiles\libocscore.dir\CoreScheduler.cpp.gcda
  root         C:\Soft_Dev\590008\JLSCONTROL\JLSOCSOURCE\MASTER
  fname        c:\soft_dev\590008\jlscontrol\jlsocsource\master\mastercommon\include\mastercommon\votercoremessages.hpp
Parsing coverage data for file c:\soft_dev\590008\jlscontrol\jlsocsource\master\mastercommon\include\mastercommon\votercoremessages.hpp
  Filtering coverage data for file c:\soft_dev\590008\jlscontrol\jlsocsource\master\mastercommon\include\mastercommon\votercoremessages.hpp
```

latk:
Thank you for raising this issue. I think you're right that gcovr should switch to case-insensitive matching in this case.

I have not previously done that because I'm not sure which precise paths should be case-insensitive on which operating system. If I remember correctly, case insensitivity is not a feature of Windows but an optional feature of the NTFS filesystem, but it would be impractical to try to detect that. I'm also not sure how that would interact with environments like MSYS.

Do you have more insight regarding when case-folding should be applied? Should this be user-configurable? Or should gcovr just ignore case under all circumstances on Windows? From the output you show it seems like gcc just lowercases all filenames, which could be done as well.

[...]

PR #386, opened 2020-06-03
Compare paths case insensitive #329

Spacetown:
If HOME directory is case insensitive the pattern match is also done case insensitive.
Test for exclude filter adapted to give filename with wrong case if testcase.exe exists in filesystem.

### DIFF
--- a/gcovr/utils.py
+++ b/gcovr/utils.py
@@ -181,7 +181,9 @@ FilterOption.NonEmpty = NonEmptyFilterOption
 class Filter(object):
     def __init__(self, pattern):
         cwd = os.getcwd()
-        is_fs_case_insensitive = os.path.exists(cwd.upper()) and os.path.exists(cwd.lower())
+        # Gueesing if file system is case insensitive.
+        # The working directory is not the root and accessable in upper and lower case.
+        is_fs_case_insensitive = ( cwd != os.path.sep ) and os.path.exists(cwd.upper()) and os.path.exists(cwd.lower())
         flags = 0
         if is_fs_case_insensitive:
             flags |= re.IGNORECASE
--- a/gcovr/utils.py
+++ b/gcovr/utils.py
@@ -181,12 +181,10 @@ FilterOption.NonEmpty = NonEmptyFilterOption
 class Filter(object):
     def __init__(self, pattern):
         cwd = os.getcwd()
-        # Gueesing if file system is case insensitive.
-        # The working directory is not the root and accessable in upper and lower case.
+        # Guessing if file system is case insensitive.
+        # The working directory is not the root and accessible in upper and lower case.
         is_fs_case_insensitive = ( cwd != os.path.sep ) and os.path.exists(cwd.upper()) and os.path.exists(cwd.lower())
-        flags = 0
-        if is_fs_case_insensitive:
-            flags |= re.IGNORECASE
+        flags = re.IGNORECASE if is_fs_case_insensitive else 0
         self.pattern = re.compile(pattern, flags)
 
     def match(self, path):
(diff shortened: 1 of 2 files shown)


Spacetown:
Changed from HOME to cwd.
I thought about a scenario with SCM ClearCase which we are using at work. This SCM uses a virtual file system driver called MVFS. This can be configured to be case sensitive even on Windows.
I think that using the CWD is the best choice because this is normaly in the filesystem of the source files. Using python directory can lead to a false detection if it is installed localy and the sources are on a different drive.

gcovr/utils.py
```diff
@@ -180,7 +180,12 @@ def __init__(self, regex, path_context=None):
 
 class Filter(object):
     def __init__(self, pattern):
-        self.pattern = re.compile(pattern)
+        cwd = os.getcwd()
+        is_fs_case_insensitive = os.path.exists(cwd.upper()) and os.path.exists(cwd.lower())
```
latk, line 184: I agree that using cwd is sensible. But there might be the case that `cwd = '/'` on Unix, which would falsely suggest a case-insensitive file system. In principle arbitrary non-alphabetic paths are possible, and I'm not sure to which degree this should be handled.

I'd suggest just a check `is_fs_case_insensitive = (cwd != '/') and ...`, ignoring other paths that are case-invariant because nearly all projects will use a sane folder layout in practice.

Spacetown: I've used os.path.sep instead of '/'.

gcovr/utils.py
```diff
@@ -180,7 +180,14 @@ def __init__(self, regex, path_context=None):
 
 class Filter(object):
     def __init__(self, pattern):
-        self.pattern = re.compile(pattern)
+        cwd = os.getcwd()
+        # Gueesing if file system is case insensitive.
+        # The working directory is not the root and accessable in upper and lower case.
+        is_fs_case_insensitive = ( cwd != os.path.sep ) and os.path.exists(cwd.upper()) and os.path.exists(cwd.lower())
+        flags = 0
```
chrta, line 187: Do we need other flags here in the near future? If not this could be changed to:
```
flags = re.IGNORECASE if is_fs_case_insensitive else 0
```

I am not sure if this coding style is ok here.

gcovr/utils.py
```diff
@@ -180,7 +180,14 @@ def __init__(self, regex, path_context=None):
 
 class Filter(object):
     def __init__(self, pattern):
-        self.pattern = re.compile(pattern)
+        cwd = os.getcwd()
+        # Gueesing if file system is case insensitive.
```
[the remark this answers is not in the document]
Spacetown: Changed

gcovr/utils.py
```diff
@@ -180,7 +180,14 @@ def __init__(self, regex, path_context=None):
...
+        # The working directory is not the root and accessable in upper and lower case.
```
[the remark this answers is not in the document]
Spacetown: Changed

gcovr/utils.py
Spacetown: Changed. The `x if y else z` is often used.

gcovr/utils.py
```diff
@@ -180,7 +180,12 @@ def __init__(self, regex, path_context=None):
 
 class Filter(object):
     def __init__(self, pattern):
-        self.pattern = re.compile(pattern)
+        cwd = os.getcwd()
+        # Guessing if file system is case insensitive.
+        # The working directory is not the root and accessible in upper and lower case.
+        is_fs_case_insensitive = ( cwd != os.path.sep ) and os.path.exists(cwd.upper()) and os.path.exists(cwd.lower())
```
latk, line 186: Linter doesn't like extra spaces.
```suggestion
        is_fs_case_insensitive = (cwd != os.path.sep) and os.path.exists(cwd.upper()) and os.path.exists(cwd.lower())
```

Spacetown:
@latk: AppVeyor hangs on master branch but the gcovr team has no access to the project. Please can you check this?

latk:
Build should be fixed now

Issue #347, opened 2019-12-13
Allow generating text report in addition to any of the other output formats


Ram-Z:
Currently when choosing any of the other output formats it's not possible to get the text report printed to stdout or even a file.

I've tried the following, but neither work.

```
$ gcovr --sonarqube coverage.xml --output
usage: gcovr [options] [search_paths...]
gcovr: error: argument -o/--output: expected one argument
$ gcovr --sonarqube coverage.xml --output -
(WARNING) --output='-' option was provided but not used.
$ gcovr --sonarqube coverage.xml --output /dev/stdout
(WARNING) --output='/dev/stdout' option was provided but not used.
$ gcovr --sonarqube coverage.xml --output report.txt
(WARNING) --output='report.txt' option was provided but not used.
```

latk:
Thank you for pointing this out! You're right, there is no `--txt` option. Text output is the default whereas all other formats must be explicitly selected. The `--output` option just provides the default target for the first active output that wasn't given an explicit output file, and is not specific to text output.

The `-s/--print-summary` option predates support for multiple output formats and produces something like the text report, but always on stdout.

So what is your intention here?

* To get a quick coverage summary on the console, please try `--print-summary`.

* If you want to save the text report for further processing/later consumption, we can discuss whether to add a `--txt` option. That would be nice for consistency's sake, but I don't really see the use case: all other formats include significantly richer data.

Ram-Z:
I'd like the text output also. It's mostly to be printed out to stdout for a quick overview from a CI build. All the other outputs cannot be parsed by a human.

`--print-summary` is what I'm using now, but I think it doesn't provide enough information.

I've opened the issue mostly to address the inconsistency.

Spacetown:
@Ram-Z In the test the options `--xml` and `--txt` are used together and both outputs are produced.

PR #387, opened 2020-06-03
Add option for separate text output #347

Spacetown:
Add option requested in #347.


gcovr/configuration.py
```diff
@@ -583,6 +583,17 @@ def choose(_cls, choices, default=None):
              "For text and HTML report.",
         action="store_true",
     ),
+    GcovrConfigOption(
+        "txt", ["-t", "--txt"],
```
latk, line 587: Single-letter options are a non-renewable resource. I'm not sure text output is important enough to get one. Opinions?

gcovr/tests/test_args.py
```diff
@@ -86,6 +86,10 @@ def helper_test_non_existing_directory_output(capsys, option='--output'):
     assert c.exception.code != 0
 
 
+def test_non_existing_directory_txt(capsys):
+    helper_test_non_existing_directory_output(capsys, '--xml')
```
latk, line 90:
```suggestion
    helper_test_non_existing_directory_output(capsys, '--txt')
```

gcovr/tests/text-output/Makefile
```diff
@@ -0,0 +1,17 @@
+all:
+	$(CXX) -fprofile-arcs -ftest-coverage -fPIC main.cpp -o testcase
+
+run: txt xml
+
+txt:
+	./testcase
+	$(GCOVR) -d --txt coverage.txt --xml coverage.xml
```
latk, line 8: At first I didn't notice that this would create both TXT and XML reports. Could this be clarified with a comment, or maybe made clearer by using a single Makefile target for both like `txt xml:`?

gcovr/configuration.py
chrta: I think you are right, `--html` also as no one letter option. I would vote for removing the `-t` here.

Spacetown: I added it because of the `--xml` but I'm not a friend of short options. If I don't know the command a long option is easier to understand. I'll remove the short option.

gcovr/tests/test_args.py
Spacetown: You're right.

gcovr/tests/text-output/Makefile
Spacetown: Sure. Changed to one target and added a comment.

Issue #381, opened 2020-05-30
Add color themes for html report


Spacetown:
From work I know CTC which uses red/blue, Jenkins default is also red/blue because of redgreen blind people. It would be nice to have such a feature for gcovr also.

latk:
Absolutely! Now that the HTML is mostly cleaned up, improving accessibility should be on the roadmap. I have mild color deficiency myself, and find gcovr's pinkish/magenta hue absolutely maddening.

Do you have a specific colour scheme in mind? Do you think we should change the default, or just implement a colour-blind mode?

In any case we should probably also add a textual indicator to uncovered lines, even if it's just a `-` hyphen.

Spacetown:
I've no schema in mind but I would change the default to a red/blue schema using lightblue instead of lightgreen. In this case the table background (SteelBlue and LightSteelBlue) should also be adapted to distinguish between the coverage colors.
Or is it better to use stronger colors?

Textual indicators will be good. What about the symbols from https://github.com/gcovr/gcovr/issues/360#issue-590526820?
I'm not interested in a dark mode.

BTW: Is this realy a Bug?

latk:
I've been experimenting a bit with colors, and think that more saturation than the current highlighting could be useful, at least if we want to keep the medium-gray left columns in the detail report. Here's a screenshot with red and blue hues, set to 60% saturation `#FF7D66` `#66D9FF`. Since the `&cross` symbol is no longer used for branch coverage, it could be repurposed for line coverage. 

![Screenshot_20200531_232604](https://user-images.githubusercontent.com/1782927/83363023-8d034100-a385-11ea-9c83-80c171edd160.png)

But perhaps it would be useful to have further contrast between covered and uncovered lines? If so, using something like `#99E6FF` to de-emphasize covered lines could be useful. Lightblue would be the same hue, just with even less saturation and a bit darker.

> BTW: Is this realy a Bug?

Accessibility problems are arguably a defect.

Spacetown:
What about using the red from the Codecov running fot the PRs and use a blue with the same saturation?
Do we really need to change the background of the whole line?
CTC only change the foreground and the color of the meter in the overview and uses different colors in the details. https://verifysoft.com/en_ctcpp.html

latk:
Good point, colors for bars and backgrounds can be different. CTC bars look good, but their line backgrounds have the same problems as current gcovr. Codecov backgrounds seem fine.

[...]

PR #388, opened 2020-06-04
Add color themes for html report #381

Spacetown:
Two themes are added:
- green: Current colors modified (more saturation).
- blue: Uses blue instead of green.

The colors are aproved by my wife.

The CSS is modified to use CSS variables for easier change colors (single place of configuration).

In the "Exec" column now contains a cross if the line is uncovered.

The test framework is update to compare all references first and use a single assertion. Getting all changes in a test instead of aborting at the first change. Also the option `--update_reference` updates all the references and not until the first change.


Spacetown:
Closed because of fourther modifications to the report.

Issue #344, opened 2019-12-03
Rounding of coverage results causes false positive with fail-under-* 100


gkovelman:
Hi,

My use-case is 100% code coverage in my project, which is validated with gcovr 4.1.

I'm seeing that, due to the rounding in `calculate_coverage` method, coverage such as 2071 lines out of 2072 lines is rounded from 99.95 to 100% and does not trigger an exit error under `--fail-under-line` 100 --fail-under-branch 100 --html-high-threshold 100` arguments.

How do you propose to mitigate this issue?
Currently I'm parsing manually the summary `-s` result from gcovr.

Thanks,
Gregory

gkpln3:
Same here

latk:
Hmm, that's what we get for mixing business logic with presentational logic :)
Thank you for reporting this bug!

Ideally, the `calculate_coverage()` function would only be used to create human-readable values, not to make decisions. The `get_global_stats()` function should leave the percentage calculation (and subsequent rounding) to its users:

* the `fail_under()` function does not need rounding
* the summary report can apply rounding itself

If anyone would like to make a PR that fixes this, I'd be happy to merge.

As a side note, gcovr is slowly moving towards more machine-readable output. Gcovr 4.2 adds a full report in JSON format, but that's a dump of the entire internal data model. A summary JSON report is proposed in PR #302 but needs to be updated.

gkovelman:
Thanks for the attention.
My use-case is quite simple so just getting an exit code of non-zero would be best for me, instead of parsing a json.

I'll see if I can submit a PR with your suggested changes.

Spacetown:
Why did you close the request? Is there no problem anymore?

gkovelman:
I closed it by mistake.
Yet I see that there's a PR in progress, so that's great.

latk:
Since the PR has been merged, can you try it out to see if it resolves the issue? You can install the unreleased development version of gcovr with `pip install git+https://github.com/gcovr/gcovr.git` (or possibly `python3 -m pip install ...`, depending on how Python is set up on your system).

senier:
> Since the PR has been merged, can you try it out to see if it resolves the issue? You can install the unreleased development version of gcovr with `pip install git+https://github.com/gcovr/gcovr.git` (or possibly `python3 -m pip install ...`, depending on how Python is set up on your system).

@latk I had the same issue with 4.1. Just tried 5.0 and it solves the problem. Thanks!

latk:
[...]

PR #389, opened 2020-06-06
Only output 100.0% if really 100.0% #344

Spacetown:
Add test for issue #344 and fix code to only show 100.0% if all branches are covered.


Spacetown:
Test was created by a script bur the JSON file would remove the dependency to the compiler. I'll check this later.

PR #390, opened 2020-06-07
Extend test option --update_reference for all formats.

Spacetown:
Option `--update_reference` wasn't working for XML outputs. For HTML output after the first difference the rest wasn't compared and updated.


Spacetown:
Changes added to other PR.

PR #393, opened 2020-06-14
Add color themes for html report #381

Spacetown:
- Adapt HTML for same look in Firefox and Safari. The color of the meter was not correct in Safari.
- The CSS is modified to use CSS variables for easier change colors (single place of configuration).
- Two themes are added:
  green: Current colors modified (more saturation). This is the default.
  blue: Uses blue instead of green.

The "Exec" column in the details view now contains a cross if the line is uncovered.


gcovr/configuration.py
```diff
@@ -623,6 +623,15 @@ def choose(_cls, choices, default=None):
         default=None,
         const=OutputOrDefault(None),
     ),
+    GcovrConfigOption(
+        "html_theme", ["--html-theme"],
+        group="output_options",
+        type=str,
+        choices=['green', 'blue'],
```
chrta, line 630: Does it make sense to get this list from CssRenderer.Themes, like `list(CssRenderer.Themes.keys())`

Spacetown: That makes sense. But then you should also define the default in the CssRenderer. I prefer `CssRenderer.get_themes()` to get the list of themes and `CssRenderer.get_default_theme()` to get the default. Then the configuration is independent of the data storage in the `CssRenderer`.

chrta: That sounds even better

Spacetown:
@chrta can you take a look at the HTML output page (test nested now contains a CSS to do this) because @latk has a mild color deficiency. @latk can you check the html code?

gcovr/html_generator.py
```diff
@@ -71,13 +71,36 @@ def load_user_template(template):
...
+        },
+        'blue': {
+            'unknown_color': "LightGray",
+            'low_color': "#FF4040",
+            'medium_color': "#F9FD63",
+            'high_color': "#6675FF",
+            'covered_color': "dodgerblue",
+            'uncovered_color': "#FF8C8C",
+            'takenBranch_color': "Blue",
+            'notTakenBranch_color': "Red"
```
latk, line 93: So, regarding the colors of the blue theme: I think the chosen blue colors are a bit too dark to be legible when used as background. I'd suggest replacing `dodgerblue` (`#1e90ff`) and instead use `#66b4ff` for high_color and covered_color.

That color is close to the `LightSteelBlue` used as table backgrounds. But if the blue theme uses `lightgray` for tables instead, the accent colors have stronger effect.

Similarly to the blue backgrounds, we could consider lightening up low_color to `#FF6666` for both themes. This should make the legend a bit easier to read.

But all in all, this configurability is a great feature that will make the reports more accessible.

gcovr/templates/root_page.html
```diff
@@ -2,53 +2,55 @@
...
+      <td>{{info.get_directory()}}</td>
+    </tr>
+    <tr>
+      <th scope="row">Date:</th>
+      <td>{{info.date}}</td>
+    </tr>
+    <tr>
+      <th scope="row">Legend:</th>
+      <td class="legend">
+        <span class="coverage-low">&nbsplow: &gt;= 0%&nbsp</span>
```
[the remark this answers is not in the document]
Spacetown: I'll change the padding in the css.

gcovr/templates/root_page.html
```diff
@@ -74,7 +76,7 @@
       {% endif %}
     </th>
     <td>
-      <meter class="{{row.lines.class}}" min="0" max="100" value="{{row.lines.coverage}}" title="%">{{row.lines.coverage}}</meter>
+      <meter class="{{row.lines.class}}" min="0" max="100" value="{{row.lines.coverage}}" title="{{row.lines.coverage}}%">{{row.lines.coverage}}</meter>
```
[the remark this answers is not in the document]
Spacetown: The title is shown in the mouse over and a lonesome percent doesn't help. Then I prefer to remove the whole title.

gcovr/templates/source_page.html
```diff
@@ -68,7 +72,7 @@
         </details>
         {% endif %}
       </td>
-      <td class="linecount {{row.covclass}}">{{row.linecount}}</td>
+      <td class="linecount {{row.covclass}}">{% if ( row.covclass == 'uncoveredLine' ) %}&cross;{% endif %}{{row.linecount}}</td>
```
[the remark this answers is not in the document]
Spacetown: You're right.

gcovr/templates/style.css
```diff
@@ -289,53 +28,29 @@ h1
...
+    text-decoration: underline;
 }
 
-.file-list
-{
-    margin: 1em auto;
-    border: 0;
-    border-spacing: 1px;
-}
+{% include "style.summary.css" %}
```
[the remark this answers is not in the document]
Spacetown: That was my intention.

gcovr/templates/style.meter.css
```diff
@@ -0,0 +1,86 @@
...
+    box-shadow: 0 5px 5px -5px #333 inset;
+}
+
+/* Webkit */
+
+meter::-webkit-meter-bar {
+    /* Outer background for Webkit */
+    background: none;
+    background-color: whitesmoke;
+    box-shadow: 0 5px 5px -5px #333 inset;
```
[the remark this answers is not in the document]
Spacetown: I had the impression that the meters stand out better from the background.

gcovr/html_generator.py
Spacetown: Sounds good, I'll give it a try. But if we use in the blue theme the same for high_color and covered_color this should also be done in the green theme.

Spacetown: The lighter colors are better because of the contrast. As mentioned, the `nested` test is a good example because all colors are used and also the row highlighting is available.

gcovr/templates/style.meter.css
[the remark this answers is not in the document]
Spacetown: I've checked it without shadow and it also looks good. I'll remove the shadow.

gcovr/html_generator.py
latk: Thanks! What's you opinion on changing LightSteelBlue for lightgray in table backgrounds in the blue theme? Currently that color isn't yet handled by the theme system, but removing blueish background would improve visibility of blue meters.

gcovr/templates/root_page.html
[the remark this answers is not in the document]
latk: I don't have a strong opinion about this so I'll rely on your judgement.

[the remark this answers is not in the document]
Spacetown: I prefer to put also the value in the title.

gcovr/html_generator.py
Spacetown: I'll check this. In the blue theme I had problems with the background.

Spacetown: I think the LightSteelBlue is better because the background of the meter is whitesmoke. I've added a little margin at the top of the meter and the blue theme to the test `nested2`.

PR #401, opened 2020-07-28
Add linkable lines in HTML details.

Spacetown:
Add linkable lines as listed in #43.


Spacetown:
I was also thinking about this but didn't remove it to show the link to the user.
I'll remove the decoration here fot a better readable report.

Yes, the regeneration is very good for such changes. :-)

PR #402, opened 2020-07-29
Add option --html-details-with-highlighting.

Spacetown:
Add optional syntax highlighting as added in #43.


gcovr/html_generator.py
```diff
@@ -115,6 +119,41 @@ def render(options):
         )
 
 
+class CodeFormater():
+
+    formatter = None
+    lexer = None
```
latk, line 125: I think this class is too complicated:

1. It models two entirely separate cases: one where the `formatter` is set, and one where the formatter is None. This looks like a good candidate for the [Replace Conditionals with Polymorphism](https://refactoring.com/catalog/replaceConditionalWithPolymorphism.html) refactoring, or for a Null Object or something.
2. Introducing a stateful `lexer` variable is potentially less clear than directly returning a function that highlights the code.

As a rough draft I'd do something like this:

```python
class NullHighlighting:
  def css(self):
    return ''

  @staticmethod
  def highlighter_for_file(filename):
    return lambda code: code

class PygmentHighlighting:
  def __init__(self):
    self.formatter = HtmlFormatter()

  def css(self):
    return self.formatter.get_style_defs()

  def highlighter_for_file(self, filename):
    try:
      lexer = get_lexer_for_filename(fielname)
    except pygments.util.ClassNotFound:
      return NullHighlighting.highlighter_for_file(filename)
    return lambda code: pygments.highlight(code, lexer, self.formatter)
```

Used as:

```python
highlighting = PygmentHighlighting() if options.html_details_with_highlighting else NullHighlighting()
hl = highlighting.highlighter_for_file(filename)
code = hl(code).rstrip()
```

gcovr/html_generator.py
```diff
@@ -332,9 +375,10 @@ def print_html_report(covdata, output_file, options):
         os.chdir(options.root_dir)
         with io.open(data['filename'], 'r', encoding=options.source_encoding,
                      errors='replace') as source_file:
+            data['lexer_used'] = formatter.set_lexer(data['filename'])
             for ctr, line in enumerate(source_file, 1):
                 data['source_lines'].append(
-                    source_row(ctr, line.rstrip(), cdata.lines.get(ctr))
+                    source_row(ctr, formatter.format_line(line), cdata.lines.get(ctr))
```
latk, line 381: I think (but am not sure) that Pygments should be used to highlight the entire file contents in one go. We can split the lines afterwards (Pygment's HTML formatter makes sure that no elements cross a line). This would matter for C++ raw strings or preprocessor directives with line continuation.

gcovr/configuration.py
```diff
@@ -635,6 +635,17 @@ def choose(_cls, choices, default=None):
...
+        "html_details_with_highlighting", ["--html-details-with-highlighting"],
+        group="output_options",
+        help="Add syntax highlighting to details page. "
+             "Implies --html-details. "
+             "OUTPUT is optional and defaults to --output.",
+        nargs='?',
+        type=OutputOrDefault,
+        default=None,
+        const=OutputOrDefault(None),
+    ),
```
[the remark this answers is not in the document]
Spacetown: OK, I'll change it to a boolean.

gcovr/html_generator.py
```diff
@@ -12,6 +12,10 @@
 import hashlib
 import io
 
+import pygments
+from pygments.lexers import get_lexer_for_filename
+from pygments.formatters.html import HtmlFormatter
```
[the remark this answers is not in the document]
Spacetown: I had this also in mind but I didn't recognize a slower test. I'll  also use lazy loading like Jinja.

gcovr/templates/source_page.html
```diff
@@ -73,7 +73,7 @@
         {% endif %}
       </td>
       <td class="linecount {{row.covclass}}">{% if ( row.covclass == 'uncoveredLine' ) %}&cross;{% else %}{{row.linecount}}{% endif %}</td>
-      <td class="src {{row.covclass}}">{{row.source}}</td>
+      <td class="src {{row.covclass}}">{% if lexer_used %}{% autoescape false %}{{row.source}}{% endautoescape %}{% else %}{{row.source}}{% endif %}</td>
```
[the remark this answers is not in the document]
Spacetown: Turn off auto escape was the solution I found. The Markup from Jinja looks easier.

gcovr/html_generator.py
Spacetown: I'll change this together with the lazy loading and wrap the string into the Markup

gcovr/html_generator.py
Spacetown: You're right, I've fotgot multi-line definitions.

Spacetown:
I've also a con:
- For large projects the generation time can increase significant.
But if there is a negotiation, we can set it as default.
I'll also add the dependency and a runtime fallback if the module is,'t installed.

latk:
The point that highlighting slows the report down is very good :thinking: 

The question is then what case should be optimized for: small to medium projects and users that are using gcovr for the first time, or larger projects. In one case there's non-obvious knowledge to make gcovr more useful, in the other non-obvious knowledge to make gcovr fast.

I think that, within reason, the median project and the newer user should be prioritized over performance:
* this likely benefits more people
* gcovr's design decisions already try to favor usability over speed (like the bruce-force heuristics to figure out where gcov should be invoked)
  * which is why fastcov exists (https://github.com/RPGillespie6/fastcov)
* there are better opportunities to gain performance than disabling syntax highlighting
  * like fixing the gcov invocation heuristics, or using the gcov-json format where possible
* people who want to speed up gcovr are more likely to read the full docs

Perhaps a compromise to ensure discoverability is to time how long HTML report generation takes and write a Clippy-style suggestion to stderr when it takes too long? “Tip: you can speed up your HTML reports by disabling syntax highlighting with --no-html-details-highlighting”. In any case, that would be out of scope for this PR.

gcovr/html_generator.py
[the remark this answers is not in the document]
Spacetown: Why is a decorator Lazy is used for Jinja? The lazy loading should also be working without this.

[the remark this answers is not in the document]
Spacetown: Ok, I got it. The function is only called once and the return value is stored inside the decorator for the next call.

Spacetown:
The hilighting looks good in default and in blue theme.

chrta:
Is there at least one test for both options, with and without syntax highlighting? Just to ensure that both settings do not break in the future.

Spacetown:
Good point @chrta, I'll update one test to use `--html-details-no-sysntaxhilighting`.

Spacetown:
> Good point @chrta, I'll update one test to use `--html-details-no-sysntaxhilighting`.

While adding the tests I found out, that leading and trailing newlines where missing in the output of pygments. The default is to strip them. See option `stripnl`in https://pygments-doc.readthedocs.io/en/latest/lexer.html. I'm updating the test data.

gcovr/configuration.py
```diff
@@ -641,6 +641,13 @@ def choose(_cls, choices, default=None):
         default=None,
         const=OutputOrDefault(None),
     ),
+    GcovrConfigOption(
+        "html_details_syntax_highlighting", ["--html-details-no-syntax-highlighting"],
+        group="output_options",
+        help="Disable syntax highlighting in HTML details page. "
+             "Default is activ syntax highlighting.",
```
chrta, line 648: `activ` -> `active`

chrta (changes requested):
Just one minor nit. Otherwise after trying it locally, it seems to work fine.

gcovr/configuration.py
Spacetown: Sure.

chrta (approved):
As far as i can tell, looks good.

gcovr/configuration.py
```diff
@@ -641,6 +641,13 @@ def choose(_cls, choices, default=None):
         default=None,
         const=OutputOrDefault(None),
     ),
+    GcovrConfigOption(
+        "html_details_syntax_highlighting", ["--html-details-no-syntax-highlighting"],
+        group="output_options",
+        help="Disable syntax highlighting in HTML details page. "
+             "Default is active syntax highlighting.",
+        action="store_false",
+    ),
```
latk, line 650: In the future, it's worth considering if we want to name negative options as `--NAMESPACE-no-FEATURE` or `--no-NAMESPACE-FEATURE`. I'm in the second camp because this allows positive/negative pairs to be autogenerated. 

For example:

```python
GcovrConfigOption(
  "html_details_syntax_highlighting", ["--html-details-syntax-highlighting"],
  group="output_options",
  help="Use syntax highlighting in HTML details page. "
       "Enabled by default.",
  action="store_const",
  default=True,
  const=True,
  const_negate=False,  # autogenerates --no-NAME with action const=False
)
```

Until now, the only use of this feature is the `--html-self-contained`/`--no-html-self-contained` option.

There are also concerns regarding configuration file support. A config key is derived for all options, unless explicitly disabled with `config=False`. It is not directly apparent what `html-details-no-syntax-highlighting = yes` would mean. Actually, I think there's a bug in the config file handling because `config-key = no` would reset to the default, which would make the above option impossible to disable. :thinking:

Spacetown: This is a Gold point. I haven't seen the existing negative Option. I also Luke the autogenerared - - no-... I'll Chance this in a separate PR.

PR #415, opened 2020-09-13
Use negative option to turn of syntax highlighting

Spacetown:
Change option as mentioned in #402

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,7 +34,7 @@ Improvements and new features:
    - Create details html filename independent from OS. (:issue:`375`)
    - Add :option:`--html-theme` to change the color theme. (:issue:`381`)
    - Add linkable lines in HTML details. (:issue:`401`)
-   - Add syntax highlighting in the details HTML report. This is configured with :option:`--html-details-syntax-highlighting`. (:issue:`402`, :issue:`415`)
+   - Add syntax highlighting in the details HTML report. This can be turned off with :option:`--no-html-details-syntax-highlighting <--html-details-syntax-highlighting>`. (:issue:`402`, :issue:`415`)
 
 Documentation:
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.rst
```diff
@@ -34,7 +34,7 @@ Improvements and new features:
    - Create details html filename independent from OS. (:issue:`375`)
    - Add :option:`--html-theme` to change the color theme. (:issue:`381`)
    - Add linkable lines in HTML details. (:issue:`401`)
-   - Add syntax highlighting in the details HTML report. This can be turned of with :option:`--html-details-no-syntax-highlighting`. (:issue:`402`)
+   - Add syntax highlighting in the details HTML report. This is configured with :option:`--html-details-syntax-highlighting`. (:issue:`402`, :issue:`415`)
```
Spacetown, line 37: The negative option can't be used in the documentation.

latk: Can be fixed by using an explicit target in the option link:

```suggestion
   - Add syntax highlighting in the details HTML report. This can be turned off with :option:`--no-html-details-syntax-highlighting <--html-details-syntax-highlighting>`. (:issue:`402`, :issue:`415`)
```

Background: our GcovrConfigOption generates Argparse options, but Argparse doesn't know about negations so it's an entirely separate option. The help message for the negation is therefore suppressed. The CLI docs are generated from the Argparse data structures with the Autoprogram Sphinx plugin. The long term solution would be to fork Autoprogram to operate on GcovrConfigOption directly, without relying on the lossy Argparse conversion. That way we could output docs for “--foo, --no-foo” as one item, which would immediately make this linkable (the `:option:` role relies on regexes to extract names from the `.. option::` directive declaration). Or we could go one level deeper and create an alternative to [that](https://github.com/sphinx-doc/sphinx/blob/v3.2.1/sphinx/domains/std.py#L181-L253) Sphinx directive. But that's too much yak-shaving even for me. Using an explicit link target is probably more appropriate.

latk (approved):

CHANGELOG.rst
Spacetown: I can also live with the explizite target.

PR #419, opened 2020-09-20
Remove Travis CI

Spacetown:
As mentioned in #404


doc/source/guide.rst
```diff
@@ -14,7 +14,7 @@ Gcovr User Guide
 The `Gcovr Home Page <http://gcovr.com>`__ is
 `<http://gcovr.com>`__.
 Automated test results are available through
-`Travis CI <https://travis-ci.org/gcovr/gcovr>`__ and
+`GitHub Actions <https://github.com/gcovr/gcovr/actions>`__ and
```
latk, line 17: We can add a filter to hide all the results from pull requests:

```suggestion
`GitHub Actions <https://github.com/gcovr/gcovr/actions?query=branch:master>`__ and
```

.github/workflows/deploy.yml
```diff
@@ -19,9 +19,21 @@ jobs:
...
+        sudo apt update
+        sudo apt-get install -y make ${{ matrix.gcc }} $(echo ${{ matrix.gcc }} | sed -e 's/gcc/g\+\+/')
+        sudo apt-get clean
     - name: Install dependencies
       run: |
         make setup-dev PYTHON=python
+    - name: Lint with flake8
+      run: |
+        make lint PYTHON=python
+        admin/release_checklist --no-verify-tags --no-verify-docs-next-version 4.2
```
latk, line 33: For the release, all items in the checklist should be enabled:

```suggestion
        admin/release_checklist 4.2
```

This will require a corresponding update in the checklist itself.

latk (commented):
Thanks for updating the docs and everything! Just some minor adjustments to how the release_checklist is used.

As a future direction, it might be interesting to figure out whether the workflow could be triggered manually, then take a pre-release version, test it, publish the release, and finally upload artefacts. Right now, a release is created before the workflow runs. However, investigating this has extremely low priority.

.github/workflows/deploy.yml
Spacetown: I copied this from Travis but you're fight.

Spacetown: I've added the checklist also to the test.yml (with the two options) and removed the options from the deploy.yml. checklist is also updated to scan both YML files.

latk (approved):

PR #420, opened 2020-09-21
Unify absolute path resolution

### LINKED ISSUE #365: Python 3.8 breaks network drive setups
Note: Due to my issue as described in https://github.com/gcovr/gcovr/issues/364, I could not test it with 4.2, only with 4.1, as the relevant code paths aren't triggered, but AFAICT the issue still remains in the current release.

My project uses a setup where on Windows a directory on the C: drive is mapped to the X: (see https://support.microsoft.com/en-us/help/4026635/windows-map-a-network-drive how to do that). gcovr works with Python 3.6 without problems. But when trying to update to Python 3.8 only (and no other relevant changes), running gcovr results in an exception:

> Traceback (most recent call last):
>   File "C:\Program Files\Python38\lib\runpy.py", line 193, in _run_module_as_main
>     return _run_code(code, main_globals, None,
>   File "C:\Program Files\Python38\lib\runpy.py", line 86, in _run_code
>     exec(code, run_globals)
>   File "X:\VuC_Build_PythonVenv\lib\site-packages\gcovr\__main__.py", line 600, in <module>
>     main()
>   File "X:\VuC_Build_PythonVenv\lib\site-packages\gcovr\__main__.py", line 588, in main
>     print_html_report(covdata, options)
>   File "X:\VuC_Build_PythonVenv\lib\site-packages\gcovr\html_generator.py", line 220, in print_html_report
>     filename=os.path.relpath(
>   File "C:\Program Files\Python38\lib\ntpath.py", line 703, in relpath
>     raise ValueError("path is on mount %r, start on mount %r" % (
> ValueError: path is on mount 'C:', start on mount 'X:'

This can be fixed by changing

```
            filename=os.path.relpath(
                os.path.realpath(cdata._filename), data['DIRECTORY']),

```

to

```
            filename=os.path.relpath(
                os.path.realpath(cdata._filename), os.path.realpath(data['DIRECTORY'])),

```
But then I get another exception later:

> Traceback (most recent call last):
>   File "C:\Program Files\Python38\lib\runpy.py", line 193, in _run_module_as_main
>     return _run_code(code, main_globals, None,
>   File "C:\Program Files\Python38\lib\runpy.py", line 86, in _run_code
>     exec(code, run_globals)
>   File "X:\VuC_Build_PythonVenv\lib\site-packages\gcovr\__main__.py", line 600, in <module>
>     main()
>   File "X:\VuC_Build_PythonVenv\lib\site-packages\gcovr\__main__.py", line 588, in main
>     print_html_report(covdata, options)
>   File "X:\VuC_Build_PythonVenv\lib\site-packages\gcovr\html_generator.py", line 181, in print_html_report
>     commondir = commonpath(files)
>   File "X:\VuC_Build_PythonVenv\lib\site-packages\gcovr\utils.py", line 185, in commonpath
>     return os.path.join(os.path.relpath(os.path.sep.join(common_dirs)), '')
>   File "C:\Program Files\Python38\lib\ntpath.py", line 703, in relpath
>     raise ValueError("path is on mount %r, start on mount %r" % (
> ValueError: path is on mount 'C:', start on mount 'X:'

I couldn't figure out how to solve this, so I don't know if there any other issues hidden.

dbywalec:
gcovr resolves relative paths to absolute in numerous places by using
os.path.abspath() & os.path.realpath() methods in inconsistent manner.

These methods are used in arbitrarily mixed manner here and there.
In build system with symlinks or drive substitution (Windows) it causes issues.

There is a substantial difference in returned result:
 * os.path.realpath() dereferences and resolves symlinks, whereas
 * os.path.abspath() does not resolves symlinks.

This commit unifies absolute path resolution by using os.path.abspath().

Fixes #365

### DIFF
--- a/gcovr/gcov.py
+++ b/gcovr/gcov.py
@@ -167,14 +167,14 @@ def guess_source_file_name(
 
 def guess_source_file_name_via_aliases(gcovname, currdir, data_fname):
     common_dir = commonpath([data_fname, currdir])
-    fname = os.path.realpath(os.path.join(common_dir, gcovname))
+    fname = os.path.abspath(os.path.join(common_dir, gcovname))
     if os.path.exists(fname):
         return fname
 
     initial_fname = fname
 
     data_fname_dir = os.path.dirname(data_fname)
-    fname = os.path.realpath(os.path.join(data_fname_dir, gcovname))
+    fname = os.path.abspath(os.path.join(data_fname_dir, gcovname))
     if os.path.exists(fname):
         return fname
 
--- a/gcovr/html_generator.py
+++ b/gcovr/html_generator.py
@@ -244,7 +244,7 @@ class RootInfo:
         }
 
         display_filename = (
-            os.path.relpath(os.path.realpath(cdata_fname), self.directory)
+            os.path.relpath(os.path.abspath(cdata_fname), self.directory)
             .replace('\\', '/'))
 
         if link_report is not None:
--- a/gcovr/utils.py
+++ b/gcovr/utils.py
@@ -78,9 +78,9 @@ def commonpath(files):
         return ''
 
     if len(files) == 1:
-        prefix_path = os.path.dirname(os.path.realpath(files[0]))
+        prefix_path = os.path.dirname(os.path.abspath(files[0]))
     else:
-        split_paths = [os.path.realpath(path).split(os.path.sep)
+        split_paths = [os.path.abspath(path).split(os.path.sep)
                        for path in files]
         # We only have to compare the lexicographically minimum and maximum
         # paths to find the common prefix of all, e.g.:
@@ -196,7 +196,7 @@ class Filter(object):
 
 class AbsoluteFilter(Filter):
     def match(self, path):
-        abspath = os.path.realpath(path)
+        abspath = os.path.abspath(path)
         return super(AbsoluteFilter, self).match(abspath)
 
 
@@ -206,7 +206,7 @@ class RelativeFilter(Filter):
         self.root = root
 
     def match(self, path):
-        abspath = os.path.realpath(path)
+        abspath = os.path.abspath(path)
         relpath = os.path.relpath(abspath, self.root)
         return super(RelativeFilter, self).match(relpath)
 
@@ -225,13 +225,13 @@ class AlwaysMatchFilter(Filter):
 
 class DirectoryPrefixFilter(Filter):
     def __init__(self, directory):
-        abspath = os.path.realpath(directory)
+        abspath = os.path.abspath(directory)
         os_independent_dir = abspath.replace(os.path.sep, '/')
         pattern = re.escape(os_independent_dir + '/')
         super(DirectoryPrefixFilter, self).__init__(pattern)
 
     def match(self, path):
-        normpath = os.path.normpath(path)
+        normpath = os.path.abspath(path)
         return super(DirectoryPrefixFilter, self).match(normpath)
 
 


latk:
Thank you for trying to tackle this long-standing issue!

This is really tricky. I understand that the path handling is somewhat inconsistent. However, changing everything to `abspath()` does not work, as evidenced by the various test failures under Linux. The tests that involve symlinks are merely disabled under Windows because we didn't get them to work there (and I don't have access to a Windows system for attempting to debug this myself).

In  particular: it is a now-irreversible design decision that filters shall apply against the real path, with symlinks resolved. Using abspath() will break stuff for Linux users, even if it might get gcovr to work under Windows. Windows support is nice, but Unix-like systems have priority since we're so closely coupled to GCC.

Your [comment](https://github.com/gcovr/gcovr/issues/365#issuecomment-696044964) has greatly improved my understanding of the Windows situation. I think it might be better to either use realpath() in more places, or perhaps to create a custom realpath() implementation that leaves drive letters alone – using the standard library is not always best. What do you think?

Spacetown:
What about using a custom function which gets the current function Name as an argument? Inside the function the function given as Argument is used if obtion isn't definex (e.g. `--use-abspath`).

latk:
> What about using a custom function which gets the current function Name as an argument? Inside the function the function given as Argument is used if obtion isn't definex (e.g. `--use-abspath`).

So a function like this?

```python
def resolve_path(path, options):

**Comment by latk** [comment]
> What about using a custom function which gets the current function Name as an argument? Inside the function the function given as Argument is used if obtion isn't definex (e.g. `--use-abspath`).

So a function somewhat like this?

```python
def resolve_path(path: str, options: Options) -> str:
  if options.use_abspath:
    return os.path.abspath(path)
  else:
    return os.path.realpath(path)

# in main:
if is_windows:
  options.use_abspath = True
```

That might work, but would mean that we have another dimension of path handling cases to consider. Users will not generally know whether they should enable/disable abspath mode. Selecting automatically based on OS might work, but would make gcovr behavior non-portable.

Personally, I think that Windows support can only go so far. That MinGW exists is amazing and I want gcovr to work in that context, but WSL will likely be a much nicer environment for cross-platform development. If reaching feature parity on native Windows means substantial sacrifices of maintainability or usability on other platforms, I don't want to do it.

dbywalec:
Hi,
Thank you for your remarks. It broadened my view on this topic too.
I understand the rationale regarding Linux/Unix platform (incl. Docker) to be primary focus and target. I fully support this approach.
Nevertheless I think Windows platform can be supported as well.
In my project I happen to use Windows as primary development and CI platform. The way we use the gcovr script provokes certain mishandling of paths.
I think it is about inconsistency of paths handling, conversion throughout the gcovr script, rather than platform specific issues.
You can think of paths using symlinks equally well on Linux. If you start using symbolic links in your build environment for whatever reason, you may likely stumble against the same problem. Windows and its drive substitution is only a one example.

According to your statement I will try to stick to conversions to `os.path.realpath()`.

I like the approach with providing dedicated methods for paths resolution. I do not think a new command line option is necessary. I would rather provide these methods for uniform conversion of paths throughout the project, so that developers are not tempted to use some another approach and things are consistent. This also paves a way towards consistent extensions in future.

That includes three aspects:
- path conversion from relative to absolute (real path)
- path conversion from absolute (real path) to relative based on current directory (`os.getcwd()`)
- getting current directory path, `os.getcwd()` returns absolute path rather than real

Sample methods:
```
def resolve_relative_path_to_absolute(path: str) -> str:
    return os.path.realpath(path)

def resolve_absolute_path_to_relative(path: str, start: str) -> str:
    return os.path.relpath(path, os.path.realpath(start))

def resolve_current_directory() -> str:
    return os.path.realpath(os.getcwd())
```

I will try to play with it a little bit and propose new solution.

dbywalec:
In order to reproduce and debug these tests I have to create a local setup on Windows. I am not so fluent in Docker and this GitHub CI setup. It takes a little bit of time. As a first test I made an attempt to run all existing tests on master branch on Windows machine. I managed to get running like 97% test on Windows. There are still 9 test failing or yielding error. I will share the current state of this work in another pull request for review.

Spacetown:
Duplicated by #535.

PR #421, opened 2020-09-21
Gcovr#365 test

dbywalec:


gcovr/json_generator.py
```diff
@@ -67,7 +67,7 @@ def print_json_summary_report(covdata, output_file, options):
 
     json_dict = {}
 
-    json_dict['root'] = os.path.relpath(options.root, output_file)
+    json_dict['root'] = resolve_path_abs_to_rel(options.root, output_file)
```
latk, line 70: One of the test failures points to this line. It seems that `output_file` can be `None`, which is the default for the `start` parameter in `relpath`. In contrast, your `resolve_path_abs_to_rel()` invokes `realpath(start)` without further checking, which will crash.

Spacetown:
@dbywalec Are you working on this or can this be closed?

Spacetown:
Closed because of no progress.

PR #422, opened 2020-09-24
Delete .#Makefile

dbywalec:
This file cannot be successfully checked out on Windows system. Git working directory always remains dirty. This is a symbolic link which is not a cross platform portable file. Furthermore it seems that this file is not used at all.


latk (approved):
Oops, it seems that file was committed by accident a long time ago. It's a tempfile without any functionality. Thanks for removing it!

PR #423, opened 2020-09-26
Extend test suite with gcc-6 ang gcc-8.

Spacetown:
Reference data for gcc-5 and gcc-8 ar kept in different folders. The folder to use is decided in `test_gcovr.py`.
Add matrix with gcc version to docker jobs.


Spacetown:
gcc-5 is not available on Windows and ubuntu-20-04. Is the test matrix OK or shall  we exclude/include combinations?
On indows I've problems with some test, any hints?
If all tests are running, we can remove AppVeyor (using gcc5).

dbywalec:
@Spacetown, I just wanted to mention that GCC 5 is available on Windows. I have recently used MinGW-W64-builds-4.3.0 with gcc-5.4.0 included for running gcovr tests under Windows. [https://sourceforge.net/projects/mingw-w64/files/](url)
Nevertheless upgrading to most recent GCC version is very desirable.

Spacetown:
> @Spacetown, I just wanted to mention that GCC 5 is available on Windows. I have recently used MinGW-W64-builds-4.3.0 with gcc-5.4.0 included for running gcovr tests under Windows. [https://sourceforge.net/projects/mingw-w64/files/](url)
> Nevertheless upgrading to most recent GCC version is very desirable.

Yes, of course. But I'm talking about the runners available on GitHub. They have a pre-installed gcc-8.

gcovr/tests/shared_lib/Makefile
```diff
@@ -40,9 +40,9 @@ run: txt xml html sonarqube coveralls
 
 coverage.json:
 ifeq ($(BASE_OS),MSYS_NT)
-	PATH="`pwd`/lib:${PATH}" testApp/test/a.out
+	PATH="$(dir $(lastword $(MAKEFILE_LIST)))lib:${PATH}" testApp/test/a.out
```
latk, line 43: Why doesn't `pwd` work here? Is this because of weird behaviour under Windows?

Using Makefile builtins seems more fragile, in particular because this here might result in relative paths like `PATH=lib:${PATH}`. Could perhaps be clarified as `$(dir $(abspath $(lastword $(MAKEFILE_LIST))))`?

Now that I see it, I'm not sure if `${PATH}` is correct (expands the variable in make) or whether `$$PATH` would be better (expands the variable in shell).

Spacetown: This was a try because the dll isn't found in Windows. I rhought that pwd might not work on GitHub with windows.
Two other tests also fail. The problem in CMake I'll discuss tomorrow with a colleague of mine.

gcovr/tests/test_gcovr.py
```diff
@@ -182,7 +184,13 @@ def pytest_generate_tests(metafunc):
...
                 pytest.mark.xfail(
                     name == 'rounding' and is_windows,
-                    reason="branch coverage seem to be platform-dependent")
+                    reason="branch coverage seem to be platform-dependent"),
+                pytest.mark.xfail(
+                    name == 'html-source-encoding-cp1252' and (os.getenv('CC') == 'gcc-8'),
+                    reason="g++-8 does not support command line option '--encoding'"),
+                pytest.mark.xfail(
+                    name == 'html-source-encoding-cp1252' and (os.getenv('TEST_OPTS') is not None),  # Environment is set in Docker build
+                    reason="g++-5 in Docker does not support command line option '--encoding'")
```
latk, line 193: I'm not sure why `--encoding` ever worked. According to the [GCC docs](https://gcc.gnu.org/onlinedocs/gcc-5.5.0/gcc/Preprocessor-Options.html#Preprocessor-Options) the correct option should probably be `-finput-charset=CHARSET`.

In any case, I don't think `TEST_OPTS` is an indicator of a Docker environment? It could be set in any build that uses the Makefile.

Spacetown: I've searched how to detect the docher container but there is no easy way. I'll remove the xfail and change the option in the tests.

latk:
I've tried to skim through the diff. While most of it are automated changes, it is *a lot*. So I'll have to trust you on this. Some questions:

* are there any relevant differences between the results for gcc-5 and gcc-8?
* what is your vision for how the test suite should evolve after this PR?

As to the test matrix: it seems fine, it could potentially even be thinned out more (we don't need to test all compiler × Python version combinations). I'd probably want to move to this matrix (potentially after this PR has landed):

```python
# test each compiler on 18.04
yield 'ubuntu-18.04', 'gcc-5', 'python-3.8'
yield 'ubuntu-18.04', 'gcc-8', 'python-3.8'
yield 'ubuntu-20.04', 'gcc-8', 'pypy3'  # include one test for pypy
# test all python versions on 20.04 and Windows
for python in 'python-3.6 python-3.7 python-3.8'.split():
  yield 'ubuntu-20.04', 'gcc-8', python
  yield 'windows-2019', 'gcc-8', python
```

What joyous coincidence that 3.5 is finally EOL this month :)

That the Checks are currently marked as pending is because I've configured “required checks” for the master branch, but these merely match on the name of the check, which has changed due to the parametrization.

gcovr/tests/test_gcovr.py
Spacetown: Done.

Spacetown:
> I've tried to skim through the diff. While most of it are automated changes, it is _a lot_. So I'll have to trust you on this. Some questions:
> 
>     * are there any relevant differences between the results for gcc-5 and gcc-8?

Yes, first I try to get the tests running, than I'll check the reference. It seems that the branches are counted different.
 
>     * what is your vision for how the test suite should evolve after this PR?

I think all test should be successful to merge the PR. As mentioned in #424 all possible checks shall be added. 
Also common parts of 'test.yml' and 'deploy.yml' shall be moved to a action which is used in both workflows.

Spacetown:
Replaced gcc-8 with gcc-6 in this PR. Version 8 results an unexpected changes in reference which isn't part of this PR.

Spacetown:
>     * are there any relevant differences between the results for gcc-5 and gcc-8?

Some branches are counted different between gcc-6, gcc-6 and gcc-8

> As to the test matrix: it seems fine, it could potentially even be thinned out more (we don't need to test all compiler × Python version combinations). I'd probably want to move to this matrix (potentially after this PR has landed):
> 
> ```python
> # test each compiler on 18.04
> yield 'ubuntu-18.04', 'gcc-5', 'python-3.8'
> yield 'ubuntu-18.04', 'gcc-8', 'python-3.8'
> yield 'ubuntu-20.04', 'gcc-8', 'pypy3'  # include one test for pypy
> # test all python versions on 20.04 and Windows
> for python in 'python-3.6 python-3.7 python-3.8'.split():
>   yield 'ubuntu-20.04', 'gcc-8', python
>   yield 'windows-2019', 'gcc-8', python
> ```

I've updated the matrix according to your suggestion. I've also added gcc-6.

> That the Checks are currently marked as pending is because I've configured “required checks” for the master branch, but these merely match on the name of the check, which has changed due to the parametrization.

Can you update the configuration for the required build.

Spacetown:
@latk, @dbywalec With tis PR the test is running under Ubuntu , Windows with sh and Windows witch cmd.

Spacetown:
@latk Can you review this PR and change the merge check configuration?

latk (approved):
I'm not sure I understand everything that's going on here, but being able to test with multiple compilers is a HUGE step forward. Thank you so much for this tedious work.

I have (temporarily) removed any required status checks in the admin settings. Once merged, a reasonable subset of the new build checks can be enforced.

That this PR was lingering for so long is due to my limited bandwidth. @Spacetown, it may be worth considering whether to call a vote to give you admin access so that you can ensure progress without waiting for me.

Spacetown:
@latk Shall I raise an issue for the admin rights?

PR #425, opened 2020-09-28
Adopt tests to run under Windows

dbywalec:
I made an attempt to run all existing tests on master branch on Windows machine. I managed to get running like 97% test on Windows. There are still 9 test failing or yielding error. Before investing more effort I would like get feedback whether this is the right direction. Many of the existing tests marked as xFAIL, can be successfully ran on Windows, incl. symbolic links.

The following tests still fail on my computer:

```
=============================================================== short test summary info ================================================================
FAILED test_gcovr.py::test_build[rounding-txt] - AssertionError: Diff output:
FAILED test_gcovr.py::test_build[rounding-html] - AssertionError: Diff output:
FAILED test_gcovr.py::test_build[simple1-txt] - AssertionError: assert False
FAILED test_gcovr.py::test_build[simple1-xml] - AssertionError: Diff output:
ERROR test_gcovr.py::test_build[shared_lib-txt] - AssertionError: assert False
ERROR test_gcovr.py::test_build[shared_lib-xml] - AssertionError: assert False
ERROR test_gcovr.py::test_build[shared_lib-html] - AssertionError: assert False
ERROR test_gcovr.py::test_build[shared_lib-sonarqube] - AssertionError: assert False
ERROR test_gcovr.py::test_build[shared_lib-coveralls] - AssertionError: assert False
====================================== 4 failed, 267 passed, 1 xfailed, 1 warning, 5 errors in 143.05s (0:02:23) =======================================
```

### DIFF
--- a/gcovr/tests/excl-line/Makefile
+++ b/gcovr/tests/excl-line/Makefile
@@ -23,6 +23,14 @@ coveralls: coverage.json
 	$(GCOVR) -a $< --coveralls coveralls.json
 
 clean:
+ifeq ($(OS),Windows_NT)
+	-cmd /C del /F /Q testcase*
+	-cmd /C del /F /Q *.gc*"
+	-cmd /C del /F /Q coverage*.*
+	-cmd /C del /F /Q sonarqube*.*
+	-cmd /C del /F /Q coveralls.json
+else
 	rm -f testcase
 	rm -f *.gc*
 	rm -f coverage*.* sonarqube*.* coveralls.json
+endif
--- a/gcovr/tests/exclude-directories-relative/Makefile
+++ b/gcovr/tests/exclude-directories-relative/Makefile
@@ -1,7 +1,12 @@
 CFLAGS= -fprofile-arcs -ftest-coverage -fPIC
 
 all:
+ifeq ($(OS),Windows_NT)
+	-mkdir build\a
+	-mkdir build\b
+else
 	mkdir -p build/a build/b
+endif
 	$(CXX) $(CFLAGS) -c a/file1.cpp -o build/a/file1.o
 	$(CXX) $(CFLAGS) -c b/main.cpp -o build/b/main.o
 	$(CXX) $(CFLAGS) build/b/main.o build/a/file1.o -o testcase
@@ -28,6 +33,14 @@ coveralls: coverage.json
 	$(GCOVR) -a $< --coveralls coveralls.json
 
 clean:
+ifeq ($(OS),Windows_NT)
+	-cmd /C del /F /Q testcase*
+	-cmd /C del /F /Q /S build
+	-cmd /C del /F /Q coverage*.*
+	-cmd /C del /F /Q sonarqube*.*
+	-cmd /C del /F /Q coveralls.json
+else
 	rm -f testcase
 	rm -rf build  # remove *.gc* *.o
 	rm -f coverage*.* sonarqube*.* coveralls.json
+endif
--- a/gcovr/tests/linked/Makefile
+++ b/gcovr/tests/linked/Makefile
@@ -33,10 +33,27 @@ coveralls: coverage.json
 	$(GCOVR) -a $< --coveralls coveralls.json
 
 clean:
+ifeq ($(OS),Windows_NT)
+	-cmd /C del /F /Q /S ..\\nested\\subdir\\*.gc*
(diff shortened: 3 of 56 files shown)


gcovr/tests/excl-line/Makefile
```diff
@@ -23,6 +23,14 @@ coveralls: coverage.json
 	$(GCOVR) -a $< --coveralls coveralls.json
 
 clean:
+ifeq ($(OS),Windows_NT)
+	-cmd /C del /F /Q testcase*
+	-cmd /C del /F /Q *.gc*"
```
latk, line 28: this `"` can be removed (sh thinks this is an un-terminated word)

gcovr/tests/exclude-directories-relative/Makefile
```diff
@@ -1,7 +1,12 @@
 CFLAGS= -fprofile-arcs -ftest-coverage -fPIC
 
 all:
+ifeq ($(OS),Windows_NT)
+	-mkdir build\a
+	-mkdir build\b
+else
```
latk, line 7: This leads  to error messages like this in Appveyor:

```
make[1]: Entering directory '/c/projects/gcovr-0p8sb/gcovr/tests/exclude-directories-relative'
mkdir build\a
mkdir: cannot create directory 'build\\a': No such file or directory
```

However, because of the leading `-` in the recipe line the error is ignored and the test fails at a later stage.

Is this because our Appveyor setup uses MSYS?

gcovr/tests/linked/Makefile
```diff
@@ -33,10 +33,27 @@ coveralls: coverage.json
...
+endif
 
 links:
+ifeq ($(OS),Windows_NT)
+	-mkdir subdir
+	cmd /C mklink /d subdir\B $(shell cd)\..\nested\subdir\B
+	-mkdir subdir\m
+	cmd /C mklink /d subdir\m\n $(shell cd)\..\nested\subdir\A
+	cmd /C mklink /d subdir\A $(shell cd)\subdir\m\n
+	cmd /C mklink /d subdir\loop $(shell cd)\subdir)
```
latk, line 55: extra trailing `)` causing failures

gcovr/tests/simple1/Makefile
```diff
@@ -43,6 +53,20 @@ coveralls:
 	$(GCOVR) -d --coveralls coveralls.json
 
 clean:
+ifeq ($(OS),Windows_NT)
+	-cmd /C del /F /Q testcase*
```
latk, line 57: It seems glob expansion fails because you've set `SHELL=cmd` above?

latk (commented):
Yes, this looks like a promising direction! I'm not totally enthusiastic that we'll end up with a wild mix of Posix sh and Windows cmd, but it that's what's needed to get this to work, then great! But could you perhaps add an explanation (here on GH, not necessarily in the Makefiles) what these different flags to cmd and del do?

I have the assumption that the necessary changes will be less invasive if glob patterns are expanded by Make before passing the resulting files to the shell, e.g. `CLEANFILES = $(wildcard coverage*.html)`, then substituting `$(CLEANFILES)` in the sh and cmd commands.

I've looked at the Appveyor test results and added some comments below. Interestingly, shared_lib passes on Appveyor. Is this because the test system uses some weird MSYS configuration that's different from your local setup?

gcovr/tests/linked/Makefile
latk: Also, what on earth does `cd` do here? Does it print the current directory or something, like pwd on Posix?

Spacetown:
It seems that the same test as in #423 are failing.

gcovr/tests/linked/Makefile
dbywalec: Another typo mistke with `)`

`$(shell cd)` is a trick to replace `pwd` which is not available on Windows. I let `make` resolve the current path.

gcovr/tests/excl-line/Makefile
dbywalec: It is a typo mistake, indeed.

Spacetown:
Is this PR still needed? On GitHub Windows is running with gcc-8.

Spacetown:
@dbywalec Is this PR still needed or can we close it?

dbywalec:
Hi, I think this issue can be closed. I ended up in another project and will not be able to work on this topic in foreseeable future.

PR #437, opened 2020-11-20
Always properly close INPUT in process_gcov_data

intosi:
When running gcovr with filters, a warning about an unclosed file was
reported for filtered or excluded file. By wrapping the io.read result
in a with statement, proper closure of the file is ensured regardless of
the path taken through the routine.

Test-information:

Checked that gcovr runs that use the -e or -f flags no longer output a
lot of warnings.

### DIFF
--- a/gcovr/gcov.py
+++ b/gcovr/gcov.py
@@ -89,44 +89,42 @@ def is_non_code(code):
 #
 def process_gcov_data(data_fname, covdata, source_fname, options, currdir=None):
     logger = Logger(options.verbose)
-    INPUT = io.open(data_fname, "r", encoding=options.source_encoding,
-                    errors='replace')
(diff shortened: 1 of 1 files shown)


Spacetown:
I've fixed flake8 error but now the github action fails with following message:
```
Error: Unable to process command '::set-env name=USE_COVERAGE::false' successfully.
Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```
I've created seperate PR #438 which should be merged first.

Spacetown (approved):

PR #443, opened 2020-12-10
Move the generators to a sub directory

Spacetown:
Instead of using the postfix _generator for each generator a separate sub directory is used.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -207,7 +207,7 @@ Path                    Description
 The program entrypoint and command line interface is in ``gcovr/__main__.py``.
 The coverage data is parsed in the ``gcovr.gcov`` module.
 The HTML, XML, text, and summary reports
-are in ``gcovr.html_generator`` and respective modules.
+are in ``gcovr.generator.html`` and respective modules.
 
 .. _test suite:
 
--- a/gcovr/__main__.py
+++ b/gcovr/__main__.py
@@ -7,7 +7,7 @@
 #
 # _____________________________________________________________________________
 #
-# Copyright (c) 2013-2020 the gcovr authors
+# Copyright (c) 2013-2021 the gcovr authors
 # Copyright (c) 2013 Sandia Corporation.
 # This software is distributed under the BSD License.
 # Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
@@ -38,15 +38,15 @@ from .version import __version__
 from .workers import Workers
 
 # generators
-from .generator.json import (gcovr_json_files_to_coverage)
-from .generator.cobertura_xml import print_xml_report
-from .generator.html import print_html_report
-from .generator.json import print_json_report, print_json_summary_report
-from .generator.txt import print_text_report
-from .generator.csv import print_csv_report
-from .generator.summary import print_summary
-from .generator.sonarqube import print_sonarqube_report
-from .generator.coveralls import print_coveralls_report
+from .writer.json import (gcovr_json_files_to_coverage)
+from .writer.cobertura import print_xml_report
+from .writer.html import print_html_report
+from .writer.json import print_json_report, print_json_summary_report
+from .writer.txt import print_text_report
+from .writer.csv import print_csv_report
+from .writer.summary import print_summary
+from .writer.sonarqube import print_sonarqube_report
+from .writer.coveralls import print_coveralls_report
 
 
 #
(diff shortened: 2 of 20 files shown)


latk:
I'm generally in favour of this, but there are interactions to consider.

* some generators also include other code, e.g. the JSON input format. In a sense, JSON is privileged because it corresponds closely to our internal coverage model.
* many editors display the base name of a file as the tab name. Clearer file names help keep an overview.
* would it make sense to move some of the utility functions to a generator-only utility module? (not necessarily in this PR)

Thoughts on this?

Potential future directions:

* if the configuration to html dependency can be broken, it would be possible to do lazy loading of generators! We are already doing some lazy loading e.g. for the template system, but the other generators also pull in heavy-ish dependencies. Might not matter much for end users, but might make the test suite a bit faster.

* There's currently rather convoluted code in the main function for loading and configuring the various generators. Ideally, some of those responsibilities could be moved into gcovr.generators. My dream is that at some point, generators expose an uniform interface so that the main function is simplified to something like `for gen in generators.configure(options): gen.generate()`

* The mirror of this PR would be to also clarify the data flow for input formats. We currently have three inputs: conversion from gcda files with gcov (possibly in parallel), using existing gcov files, and the --add-tracefile feature.

Spacetown:
> I'm generally in favour of this, but there are interactions to consider.
> 
>     * some generators also include other code, e.g. the JSON input format. In a sense, JSON is privileged because it corresponds closely to our internal coverage model.

You're right here, I didn't saw this. This generator is also a input reader.

>     * many editors display the base name of a file as the tab name. Clearer file names help keep an overview.

We can also change the postfix `_generator` to a prefix `generator_` to group the generators. This was my main goal.

>     * would it make sense to move some of the utility functions to a generator-only utility module? (not necessarily in this PR)

This make sense.

> Potential future directions:
> 
>     * if the configuration to html dependency can be broken, it would be possible to do lazy loading of generators! We are already doing some lazy loading e.g. for the template system, but the other generators also pull in heavy-ish dependencies. Might not matter much for end users, but might make the test suite a bit faster.
> 
>     * There's currently rather convoluted code in the main function for loading and configuring the various generators. Ideally, some of those responsibilities could be moved into gcovr.generators. My dream is that at some point, generators expose an uniform interface so that the main function is simplified to something like `for gen in generators.configure(options): gen.generate()`

I've started in this direction with moving the defaults for the HTML theme to the HTML generator. I think we should go in this direction with a base class for a generator with a clear interface.
 
>     * The mirror of this PR would be to also clarify the project structure for input formats. We currently have three inputs: conversion from gcda files with gcov (possibly in parallel), using existing gcov files, and the --add-tracefile feature.

A struture with input "filters" to read the data from different sources and output "filters" to write the internal data to a output file will be easier to maintain and extend.

chrta (approved):
lgtm I think it is a good direction to split up files into directories

Spacetown:
@latk, @chrta How should I proceed? The editor tab can be a point against the change. I use VS Code which uses the basename in the tab but I prefer smaller tabs to reduce scrolling in the tab bar. Also VS Code selects the focused file in the workspace tree. If you have two files with the same name from different directories opened the tabs for this files are extended with the directories.
I agree that it can be confusing in other editors but only if you are working on two files with the same name in different directories. For the test you have already the problem, there are many main.cpp and Makefiles.

I propose the following procedure:

- [ ]  Proceed with this PR with the sub directories.
- [ ] Add a base class for generators to simplify the load/call of the generators in the main function. This base class can also contain the common utility functions.
- [ ] Move the command line options to the generators.
- [ ] Do the same stuff for the input files.
- [ ] Add a input reader for the JSON gcov format. 😃

chrta:
I would prefer if you proceed with this pr. In my opinion the visualization in an IDE/editor should not prevent us from doing this change.
I would also prefer to keep the short filenames, as they are now. I would be against adding a "generator" to every filename in that directory, since this is just redundant in my opinion.

Spacetown:
>  * The mirror of this PR would be to also clarify the project structure for input formats. We currently have three inputs: conversion from gcda files with gcov (possibly in parallel), using existing gcov files, and the --add-tracefile feature.

I'm thinking about changing "generator" because of the naming of the input "filters". What about reader/writer or input/output?

Spacetown:
@chrta @latk Should we wait with this PR until the next release?

latk:
I see no reason to wait. Incremental improvement is better then waiting forever until we find a perfect design. And since this is a purely internal change, it is reversible if we change our minds later. Very low risk of just doing it.

PR #444, opened 2020-12-10
Add black formatter

Spacetown:
- Move configuration for coverage from setup.cfg to pyproject.toml.
- Switch from flake8 to flake9 (Fork which supports pytoml.xml) and update documentation.
- Add configuration for [black](https://black.readthedocs.io/en/stable/the_black_code_style.html) to format code.
  - Pro:
    - I used it in another project and the formatted code looks good for me.
    - Formatting of code is independand of the developer and looks the same in the whole project.
    - No need of configuration.
 - Contra:
    - No configuration possible.


latk:
After thinking about this for a bit, I'm not quite sure this is a step in the right direction.

* creating a `pyproject.toml`? Probably fine, though I don't see the value yet. Yes, Toml is a better format than Ini, but just moving configuration around doesn't necessarily improve much.
* using flake9? That looks like an incompatible fork that could cause problems if a first-time contributor wants to work on gcovr without a virtualenv, and who already has flake8 installed. If we want to move flake8 configuration into a config file, then setup.cfg would seem like the appropriate target.
* the black formatter is totally fine, as long as it won't change the current code (otherwise, a commit that re-formats the code would be appropriate). If this tool is supposed to be part of the development workflow, then adding it as a dependency would be appropriate.
* I don't quite see why requirements files are a problem. They are widely supported in the PIP ecosystem and easy to use. Our dependency lists don't have to be generated by code.

Of course I'm willing to be convinced otherwise. There are two very good aspects in this PR that I'd like to keep though: specifying a code formatter, and simplifying invocation of flake8 by using a config file.

Spacetown:
> * creating a `pyproject.toml`? Probably fine, though I don't see the value yet. Yes, Toml is a better format than Ini, but just moving configuration around doesn't necessarily improve much.
>
> * using flake9? That looks like an incompatible fork that could cause problems if a first-time contributor wants to work on gcovr without a virtualenv, and who already has flake8 installed. If we want to move flake8 configuration into a config file, then setup.cfg would seem like the appropriate target.

Flake9 only because flake8 doesn't support pyproject.toml. I thought one configuration to configure all tools would be fine.  I think it will be better to document onla that flake8 is used and the options are fixed in a configuration (maybe setup.py).

> * the black formatter is totally fine, as long as it won't change the current code (otherwise, a commit that re-formats the code would be appropriate). If this tool is supposed to be part of the development workflow, then adding it as a dependency would be appropriate.

If we start to use black it should be added to the tests and a reformat of all files is needed one time.

>     * I don't quite see why requirements files are a problem. They are widely supported in the PIP ecosystem and easy to use. Our dependency lists don't have to be generated by code.

The runtime requirements are configured in `setup.py`, the development requirements in `requirements.txt` and the documentation requirements in `doc/requirements.txt`. I think, a single place to specify all requirements is better.
 
> Of course I'm willing to be convinced otherwise. There are two very good aspects in this PR that I'd like to keep though: specifying a code formatter, and simplifying invocation of flake8 by using a config file.

The simplifying of the invocation is a big advantage.

Spacetown:
@latk I've revert the pyproject.toml, moved the configuration of flake8 to setup.cfg and introduced black.
The errorlevel of black is ignored at the moment because I don't want to reformat all the source files. This can be done if the file is changed.

.dockerignore
```diff
@@ -12,6 +12,7 @@
 !*.txt
 !*.cfg
 !**/requirements.txt
+!gcovr/version.py
```
latk, line 15: A few lines above there is a `!gcovr/**.py` pattern, so I think adding this ought to be redundant?

CHANGELOG.rst
```diff
@@ -52,10 +52,11 @@ Internal changes:
 
  - Add makefile + dockerfile for simpler testing.
  - Add .gitbugtraq to link comments to issue tracker in GUIs. (:issue:`429`)
- - Add GitHup actions to test PRs and master branch. (:issue:`404`)
+ - Add GitHub actions to test PRs and master branch. (:issue:`404`)
  - Remove Travis CI. (:issue:`419`)
- - Remove Appveyor CI and udloade coverage report from Windows and Ubuntu from the GitHub actions. (:issue:`455`)
+ - Remove Appveyor CI and udload coverage report from Windows and Ubuntu from the GitHub actions. (:issue:`455`)
```
latk, line 57: Typo.

```suggestion
 - Remove Appveyor CI and upload coverage report from Windows and Ubuntu from the GitHub actions. (:issue:`455`)
```

latk (approved):
Looks good! Just two minor inline comments below.

For future reference, updating the title might also be a good idea.

Unfortunately, reformatting all code will lead to substantial merge conflicts. So we should probably delay that step for a bit until fewer PRs are still pending. So it's good that the CI failure is ignored (for now).

.dockerignore
Spacetown: You're right.

PR #445, opened 2020-12-11
Run as limited user inside docker

Spacetown:
Extend test with read only output directory.

### DIFF
--- a/admin/Dockerfile.qa
+++ b/admin/Dockerfile.qa
@@ -1,11 +1,12 @@
 FROM ubuntu:18.04
 
+ARG USERID
 ARG CC
 ARG CXX
 
 RUN \
   apt update && \
-  apt-get install -y make $CC $CXX python3 python3-pip ninja-build && \
+  apt-get install -y sudo make $CC $CXX python3 python3-pip ninja-build && \
   apt-get clean
 
 WORKDIR /gcovr
@@ -18,4 +19,13 @@ RUN \
   $PYTHON -m pip install --no-cache-dir --upgrade pip && \
   $PYTHON -m pip install --no-cache-dir -r requirements.txt -r doc/requirements.txt
 
-CMD $PYTHON -m pip install -e . && make qa
+# Create new user "docker" and set password to "docker"
+RUN addgroup docker
+RUN useradd -d $HOME -u $USERID -g docker -g sudo -s /bin/bash docker
+RUN echo "docker:docker" | chpasswd
+
+USER docker:docker
+
+CMD ( echo docker | sudo -S -p "Running pip with sudo" $PYTHON -m pip install -e . ) && \
+  ( echo docker | sudo -S -p "Running chmod with sudo" chown -R docker:docker /gcovr ) && \
+  make qa
--- a/gcovr/tests/test_args.py
+++ b/gcovr/tests/test_args.py
@@ -19,6 +19,8 @@
 from ..__main__ import main
 from ..version import __version__
 
+import pytest
+import os
 import re
 import sys
 
@@ -98,11 +100,15 @@ def test_invalid_objdir(capsys):
     assert c.exception.code == 1
 
 
-def helper_test_non_existing_directory_output(capsys, option='--output'):
+def helper_test_non_existing_directory_output(capsys, option):
     c = capture(capsys, [option, 'not-existing-dir/file.txt'])
     assert c.out == ''
     assert 'Could not create output file \'not-existing-dir/file.txt\': ' in c.err
     assert c.exception.code != 0
+
+
+def test_non_existing_directory_output(capsys):
+    helper_test_non_existing_directory_output(capsys, '--output')
 
 
 def test_non_existing_directory_txt(capsys):
@@ -133,11 +139,15 @@ def test_non_existing_directory_csv(capsys):
     helper_test_non_existing_directory_output(capsys, '--csv')
 
 
-def helper_test_non_existing_directory_2_output(capsys, option='--output'):
+def helper_test_non_existing_directory_2_output(capsys, option):
     c = capture(capsys, [option, 'not-existing-dir/subdir/'])
     assert c.out == ''
     assert 'Could not create output directory \'not-existing-dir/subdir/\': ' in c.err
     assert c.exception.code != 0
+
+
+def test_non_existing_directory_2_output(capsys):
+    helper_test_non_existing_directory_2_output(capsys, '--output')
 
 
 def test_non_existing_directory_2_txt(capsys):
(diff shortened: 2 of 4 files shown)


Spacetown:
From a colleague I got the information how to use a different user in docker.
Now also a read only output directory is checked under ubuntu and inside the docker container.

gcovr/tests/test_args.py
```diff
@@ -150,6 +159,53 @@ def test_non_existing_directory_2_csv(capsys):
...
+
+
+@pytest.mark.skipif(sys.platform == 'win32', reason="Not for Windows")
+def test_non_writable_directory_output(capsys):
+    helper_test_non_writable_directory_output(capsys, '--output')
+
+
+@pytest.mark.skipif(sys.platform == 'win32', reason="Not for Windows")
+def test_non_writable_directory_txt(capsys):
+    helper_test_non_writable_directory_output(capsys, '--txt')
```
latk, line 176: It seems like these tests would be the perfect use case for `@pytest.mark.parametrize()`, but I recall there were some problems last time we tried this.

admin/Dockerfile.qa
```diff
@@ -18,4 +18,13 @@ RUN \
...
-CMD $PYTHON -m pip install -e . && make qa
+# Create new user "docker" and set password to "docker"
+RUN addgroup docker
+RUN useradd -d $HOME -g docker -g sudo -s /bin/bash docker
+RUN echo "docker:docker" | chpasswd
+
+USER docker:docker
+
+CMD ( echo docker | sudo -S -p "Running pip with sudo" $PYTHON -m pip install -e . ) && \
+  ( echo docker | sudo -S -p "Running chmod with sudo" chown -R docker:docker /gcovr ) && \
```
latk, line 29: I am afraid the `chmod` will affect the user ID of the file on the host system, potentially making the project inaccessible for the user. Since you create a group+user, they will get the UID=1000, GID=1000. This happens to be the matching UID/GID for the *first* user on most Linux systems. This will be the completely wrong user ID for other users, especially for people working on shared servers (not that unusual in academia). This could render the gcovr project directory inaccessible and undeletable for such users after running a test. Of course being able to run Docker is effective root access so they could restore access with e.g. `docker run --rm -v "$(pwd)":/gcovr ubuntu chmod -R "$(id -u):$(id -g)" /gcovr`, but that is rather non-obvious.

Idea: can we inject the appropriate numeric IDs at runtime through environment variables? Or would Docker's usermap feature help? I've never used it.

gcovr/tests/test_args.py
```diff
@@ -150,6 +159,53 @@ def test_non_existing_directory_2_csv(capsys):
     helper_test_non_existing_directory_2_output(capsys, '--csv')
 
 
+def helper_test_non_writable_directory_output(capsys, option):
+    c = capture(capsys, [option, '/file.txt'])
```
latk, line 163: Hmm, this test is correct within the Docker container, but might have surprising results on a real system. Can we somehow guard this to only execute in an expected environment?

latk (commented):
This sounds like a great change, especially for the extra tests this enables. However, these changes do have the potential to damage a developer's system, so extra discussion is needed.

gcovr/tests/test_args.py
Spacetown: Yes, `parametrize` didn't work together with the capsys argument.

admin/Dockerfile.qa
Spacetown: Good point, I forgot this. The `chmod` was needed because of the option `--archive_artifacts` which is used on GitHub actions. I`ll check this.

gcovr/tests/test_args.py
Spacetown: We can run this check only inside docker. You mean the effect if someone develops gcov with an account having access to `/`? In windows we can use the environment `SystemRoot` which should also be protected for normal users.
Maybe we can try to create a file in the exepected directory and if this fails, we run the real test, else we skip the test at runtime.

admin/Dockerfile.qa
Spacetown: I've checked this point. If I run following command inside docker I get the owner docker:
```shell
$ touch xy
$ dir -l xy
-rw-r--r-- 1 docker docker 0 Dec 14 20:00 xy
```

Outside docker, the owner is the user running the docker instance:
```shell
Snoopy@Admins-Air gcovr % ls -l xy
-rw-r--r--+ 1 Snoopy  wheel  0 Dec 14 21:00 xy
```

Also the file mode isn't changed in the host system.

It seems, that the chmod doesn't change the

latk: Could you try showing the file with `ls -ln` to show the numerical user/group IDs? I ran tests on Ubuntu host + container, and the two systems always showed the same numerical IDs in the file permissions. Of course the user name was different, but that's just due to different /etc/passwd contents. So when running as user 1000 in the container host would see user 1000. You might therefore be able to provoke breakage by choosing a different uid in the Docker container, e.g. with `useradd ... -u 1234`.

It could be that things work differently on Macos, since it doesn't support Docker natively but requires a virtualization layer. I would have expected to already see such breakage on your system, since Macos should start numbering UIDs at 501, not at 1000.

This article suggests telling the container about the UID via an environment variable, and creating an entrypoint that dynamically creates a user with appropriate UID: https://denibertovic.com/posts/handling-permissions-with-docker-volumes/ – this might actually work.

\<rant> The real problem is that on Linux, Docker spawns container processes from the Docker Daemon which runs as root, and doesn't really know about the user who “started” the container. The docker command is just a client for dockerd. I understand why that's necessary, but the entire architecture is insane and leads to endless problems. I looked at the user remap / userns feature and it's totally unsuitable for us, since people wanting to run the container would have to reconfigure their Docker Daemon. \</rant>

gcovr/tests/test_args.py
latk: Yes, but it's a matter of principle: don't touch random files, only files in the current directory or in a tempdir. Even though no sane person should run tests as root or Admin.

My preferred solution would be to make sure that this test won't execute outside of the testing container, e.g. by setting an env variable in the container and checking for its existence in the test. E.g. `ENV GCOVR_ISOLATED_TEST=zkQEVaBpXF1i` with some randomly selected but static value to prevent false positives.

Spacetown: Ok. I'll do it this way. The test is skipped if the environment variable has another value.

admin/Dockerfile.qa
Spacetown: Inside docker I get following output:
```
ls -l test.xy
-rw-r--r-- 1 docker docker 0 Dec 17 20:37 test.xy
ls -ln test.xy
-rw-r--r-- 1 1000 1000 0 Dec 17 20:37 test.xy
```

and inside MacOs:
```
Snoopy@Admins-Air gcovr % ls -l test.xy      
-rw-r--r--@ 1 Snoopy  wheel  0 Dec 17 21:37 test.xy
Snoopy@Admins-Air gcovr % ls -ln test.xy
-rw-r--r--@ 1 503  0  0 Dec 17 21:37 test.xy
```

It seems, that on MacOs the files from docker are mapped to the user running the container.

I added the effective user and group id to the container. Inside the container the group and user are mapped to this ids. At the moment I've added the ls commands to the makefile to get the data from the actions.

Spacetown: My gid from the host is already used inside docker, therefore I removed the gid again.
Because of the issue https://github.com/actions/setup-python/issues/171 I've switched to the last commit to run tests with pypy3.

gcovr/tests/test_args.py
Spacetown: I took your random ID and it's working.

latk (approved):
I've tested on my system and it seems to work fine. Thank you!

PR #447, opened 2020-12-16
Fix warning in console if not -o is used.

Spacetown:
Fix warning from test.
```shell
/usr/bin/python3 -m gcovr -d --json main.json
(WARNING) --output=None option was provided but not used.
```


latk (approved):
Looks good! A bit unfortunate about the missing test coverage, but adding a test to cover this case has too little value compared to the effort. (We can't just do an argument level test because this check happens *after* any output generation – but that might be refactored in the future.)

Spacetown:
Maybe we can extend the test suite to scan the test output and assert that there is no warning or error.

PR #450, opened 2021-01-06
Add script for copyright header.

Spacetown:
Inspired from another project I added a script which writes the copyright info to each python file. The comment contains the contributors sorted by the contribution date and afterwards by name.

The version in the copiright is used from `__Version__.py`.  For the copyright in the documtation the current year is used instead of the fixed 2018.

One idea is to create also the author list from the git log or to extend the copyright comment to other file types.

### DIFF
--- a/admin/add_copyright.py
+++ b/admin/add_copyright.py
@@ -9,8 +9,7 @@
 #
 # _____________________________________________________________________________
 #
-# Copyright (c) 2021 Spacetown <michael.foerderer@gmx.de>
-# and possibly others.
+# Copyright (c) 2021 the gcovr authors
 # Copyright (c) 2013 Sandia Corporation.
 # This software is distributed under the BSD License.
 # Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
@@ -32,6 +31,11 @@ HEADER_END = (
 
 
 def getLicenseSection(filename, comment_char="#"):
+    date = subprocess.check_output(
+        ["git", "log", "-1", "--format=format:%ad", "--date=short", "--", filename],
+        universal_newlines=True,
+    )
+    year = date[:4]
     yield comment_char + "  ************************** Copyrights and license ***************************"
     yield comment_char
     yield comment_char + " This file is part of gcovr {}, a parsing and reporting tool for gcov.".format(
@@ -41,9 +45,7 @@ def getLicenseSection(filename, comment_char="#"):
     yield comment_char
     yield comment_char + " _____________________________________________________________________________"
     yield comment_char
-    for name, year in getContributors(filename):
-        yield comment_char + " Copyright (c) " + year + " " + name
-    yield comment_char + " and possibly others."
+    yield comment_char + " Copyright (c) " + year + " the gcovr authors"
     yield comment_char + " Copyright (c) 2013 Sandia Corporation."
     yield comment_char + " This software is distributed under the BSD License."
     yield comment_char + " Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,"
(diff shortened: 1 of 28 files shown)


chrta (approved):
LGTM

latk:
Having an automated tool to update copyright notices is very good, but adding all authors that touched the files isn't necessarily helpful.

My rationale for introducing “Copyright YEAR the gcovr authors” + an AUTHORS file was the following.

* Having copyright notices is good, but they should not be misleading. We cannot accurately know the full range of rightsholders, in particular if individuals contribute on behalf of their employers or forward someone else's contributions. So the committer might not be the rightsholder that should be attributed. As code is refactored, someone's contributions might be moved to a different file. While a phrase such as “the gcovr authors” is maximally meaningless, it is also minimally misleading.
* I try to be very respectful of contributor's privacy. Of course name + email will be visible in the git log, but at least it's not directly scrapeable from the web. Adding their name to the AUTHORS file is entirely voluntary, and lets them explicitly select the name or alias they would like to use publicly. Many people are not fully aware of the information in their git config. This change would place PII and contact information on all systems where gcovr is installed, which raises GDPR questions that I'm not excited to think about.
* The effort of tracking precise contributor information is fairly high: there would have to be O(files × contributors) updates. Decoupling contributor tracking from the copyright notices reduces this to O(contributors + files × years). Technically this automated script reduces the effort, but would still introduce some degree of noise in the git log.

If we do this, I'd like to first hear an argument for the advantages of adding more detailed copyright notices, and in particular a strong argument for adding emails. We should probably also start using a mailmap file to allow people to select their preferred name + address.

I also noticed that the script has wiped out some comment blocks, though some of them were so outdated that it's probably a good idea to remove them!

Spacetown:
@latk Privacy is a good point. My intention was to have the same header in each file and also updated the year and the version by script because some headers where outdated or had different header or none copyright header. I'll replace the authors with your string and the original company.

Spacetown:
I've compared the files and in `doc/source/conf.py` I forgeot to move the old comment. The other comments were outdated.

admin/add_copyright.py
```diff
@@ -0,0 +1,124 @@
...
+
+def getLicenseSection(filename, comment_char="#"):
+    date = subprocess.check_output(
+        ["git", "log", "-1", "--format=format:%ad", "--date=short", "--", filename],
+        universal_newlines=True,
+    )
+    year = date[:4]
+    yield comment_char + "  ************************** Copyrights and license ***************************"
+    yield comment_char
+    yield comment_char + " This file is part of gcovr {}, a parsing and reporting tool for gcov.".format(
```
latk, line 41: If we drop support for Python 3.5 (which is EOL since last year), we could use f-strings: `f"foo {variable} bar"`

admin/add_copyright.py
```diff
@@ -0,0 +1,124 @@
...
+        version.__version__
+    )
+    yield comment_char + " https://gcovr.com/en/stable"
+    yield comment_char
+    yield comment_char + " _____________________________________________________________________________"
+    yield comment_char
+    yield comment_char + " Copyright (c) " + year + " the gcovr authors"
```
latk, line 48: I'd probably keep a year range to indicate that there were changes in between 2013 and now. Most files include some of that old code, though this is not quite correct for the newer generators. On the other hand, `git log --follow` isn't particularly accurate either.

```suggestion
    yield comment_char + f" Copyright (c) 2013-{year} the gcovr authors"
```

admin/add_copyright.py
```diff
@@ -0,0 +1,124 @@
...
+        if headerEndReached:
+            if skipEmptyLines:
+                if len(line) == 0:
+                    continue
+                skipEmptyLines = False
+                newLines.append("")
+            newLines.append(line)
+
+    if not headerEndReached:
+        newLines.extend(lines)
```
latk, line 95: This kind of `headerEndReached` and `skipEmptyLines` state can be replaced by control flow and iterators. Whether to do that is a stylistic decision. I find control flow easier to follow than state machines.

```python
lines = iter(lines)

# skip lines until header end marker
for line in lines:
  if len(line) > 0 and line == '#' + HEADER_END:
    break

# skip empty lines, keeping the first non-empty line
for line in lines:
  if line:
    newlines.append("")
    newlines.append(line)
    break

# keep all other lines
newlines.extend(lines)
```

latk (approved):
I like this!

Some tentative suggestions below.

admin/add_copyright.py
Spacetown: In the test it's already dropped but I think for the next release we shouldn't use this. After the release we can change every code places to format strings.

admin/add_copyright.py
Spacetown: I'll change this to `"".format()` and after the next release to `f"..."`.

admin/add_copyright.py
Spacetown: I've never used an iterator explicite but this is a good approach. The only missing thing is if the header isn't found. I'll use this.

admin/add_copyright.py
Spacetown: Cahanged to 2013-<Year of last change>.

PR #451, opened 2021-01-07
Fix additional GitHub language statistics

Spacetown:
Also remove Makefile and CSS files.


CezaryGapinski (approved):
Oh, you are right. Python language should be on the first place.

PR #452, opened 2021-01-09
Fix overzealous substitution of root folder

jkuebart:
If the root folder occurs somewhere in the file name, it is replaced by the empty string by `root_filter.sub('', filename)`, resulting in an invalid file name. Consider for example the transformation
```
/src/libs/libjhtml/src/libjhtml/DataUri.cc -> libs/libjhtmllibjhtml/DataUri.cc
```
when `--root /src` is given.

This PR fixes the problem by only substituting the root folder at the start of the file name.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -24,6 +24,7 @@ The following developers contributed to gcovr (ordered alphabetically):
     James Reynolds,
     Jeremy Fixemer,
     Jessica Levine,
+    Joachim Kuebart,
     Joel Klinghed,
     John Siirola,
     Jörg Kreuzberger,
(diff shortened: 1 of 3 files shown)


latk (approved):
Ouch, it's a bit embarrassing that a bug this obvious stayed there for so long - all the more thanks for fixing it!

If you want, you could also add your name to the Authors file and add a Changelog entry about this fix.

jkuebart:
@latk I guess most of the time roots are too long to appear by accident – this was in a docker container with a very short root path. I was quite perplexed by what was going on until I looked at it in detail ;-)

Thanks for the advice, I'll update the files – let me know if there's anything else to do (add a test? haven't looked into those yet).

Spacetown (approved):
LGFM

Spacetown:
Thanks for the fix.

PR #455, opened 2021-01-19
Upload coverage data from GitHub actions.

Spacetown:
@latk Can we now remove appveyor tests? GCC5 is checked in Docker and Windows is checked with GCC-8.


Spacetown:
Differences in coverage are because the report is know generated under ubuntu.

latk:
Would it be possible to collect coverage in one Windows build as well? Codecov merges multiple uploaded reports, which is helpful because we have some Windows- and some Linux-specific tests.

I'll delete Appveyor ass soon as you say so, or when this PR is merged.

latk (approved):
Looks good!

I think the Appveyor config can be deleted as well.

Spacetown:
@latk It seems that with the GitHub actions the data is visible as comment but the test is missing. I'll try to call the commen `codecov -X gcov search` used in appveyor to upload the data. Codecov.io support user defined flags for the upload, I'll try to flag with the OS.

Spacetown:
@latk The requred appveyor test can be removed.

latk:
ok done, Appveyor is deleted. Thank you very much for making this possible!

Spacetown:
And what about the missing needed test? Was there a special configuration on appveyor?

latk:
I think the codecov website is just unreliable in the diff view, but is actually processing the correct data. So I don't think anything has to be fixed regarding coverage collection (at least on our part).

Spacetown:
Strange, now the codecov is listed again. ???
Maybe this was an issue with the required appveyor test.

Issue #424, opened 2020-09-28
Prepare gcovr 5.0 release


latk:
Towards the next gcovr release, the following steps have to be done:

* [x] ensure all relevant changes (fixed issues, merged PRs) are noted in the changelog
* [x] check that the docs are roughly up to date, organize the changelog
* [x] prepare a draft GitHub release (same content as changelog, but rephrased/reformatted in a way that makes sense as an email)
* [x] temporary feature freeze, last update to the docs
* [x] bump the version number, including in documentation and screenshots
* [x] publish the GH release, which will create a tag and trigger the deployment workflow, uploading a wheel to PyPI

Updating the changelog is gargantuan task, fortunately it can be done in steps and parallelized across people with **write access**. To this end, I've created a new 4.3 milestone.

**If you assign a completed issue/PR to the 4.3 milestone**, this indicates either
* that the issue is already mentioned in the changelog,
* or a commitment that you will submit a PR that adds the issue to the changelog,
* or that the issue will be part of the release but is too minor to be mentioned in the changelog.

To limit merge conflicts, do not reorganize the changelog, but add new entries as a single block (can be organized into sections later). 

It could make sense to also create milestones for past releases, so that it's easier to find issues that have yet to be looked at. Issues tagged Question or wontfix can likely be skipped.

This issue can be used for further release planning. General process discussion should be done at #400, where the idea to use the “milestone” feature was suggested.

Spacetown:
Have all PRs listed in the changelog?
E.g. #407 is only a fix of #328 which is mentioned in the changelog. I prefer not to add it to the changelog.

latk:
The changelog is for humans, so it makes sense to only mention changes that are of interest to anyone who doesn't want to trawl through the commit history – so primarily for end users, but to some degree also for other gcovr developers.

So I wouldn't mention #407. Docs changes can be mentioned if there's new material that someone would be interested to read or re-read. #405 is a bit too minor I think. At most something like “clarified how filters work in the --help message”.

Spacetown:
But we link e.g. #407 and #405 to see that that the are already checked, ok?

We can also add a milestone for the next UpcomingRelease where we can do this work as part of the PR review.

latk:
Linking the issues/PRs sounds sensible, I've updated the top comment. (Later, the experience from this thread can be condensed as docs on the release process).

[...]

PR #457, opened 2021-01-21
Check if issue is mentioned in changelog.

Spacetown:
Test was mentioned in #424. The check can be deactivated by adding the text "[no changelog]" to a single line in the comment.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,9 +11,23 @@ on:
 
 jobs:
 
-  build:
+  release-check:
+    runs-on: ubuntu-20.04
 
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup environment
+      run: |
+        echo "GITHUB_PR_NUMBER=${{ github.event.pull_request.number }}" >> $GITHUB_ENV
+      shell: bash
+    - name: Run release_checklist
+      run: |
+        admin/release_checklist --no-verify-tags --no-verify-docs-next-version 4.2
+ 
+  build:
     runs-on: ${{ matrix.os }}
+    needs: release-check
+
     strategy:
       fail-fast: false
       matrix:
@@ -76,9 +90,6 @@ jobs:
     - name: Install dependencies
       run: |
         make setup-dev
-    - name: Run release_checklist
-      run: |
-        admin/release_checklist --no-verify-tags --no-verify-docs-next-version 4.2
     - name: Lint with flake8
       run: |
         make lint
@@ -100,9 +111,9 @@ jobs:
         make doc
   
   run-docker:
-
     runs-on: ubuntu-18.04
-  
+    needs: release-check
+
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,6 +11,39 @@ on:
 
 jobs:
 
+  changelog-check:
+    runs-on: ubuntu-20.04
+    if: ${{ github.event_name == 'pull_request' }}
+
+    env:
+      PR_BODY: ${{ github.event.pull_request.body }}
+      CHANGELOG_ISSUE: ":issue:`${{ github.event.pull_request.number }}`"
+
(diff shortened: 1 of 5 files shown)


chrta:
Code looks good. I have 2 questions:
1. Should the check be mentioned in https://gcovr.com/en/stable/contributing.html#how-to-submit-a-pull-request? Also the fact that adding `[no changelog]` can be added there for minor changes.
2. Do i understand correctly that adding `[no changelog]` to any commit message skips the check? So it is not necessary to add this to all commits in a pr, only to one commit.

latk:
As an alternative to looking at commit messages, we could consider looking at the pull request comment via the Github API. This would likely mean that the check would be performed only in the Action, not in the release checklist. In a shell script, the PR comment can be extracted like:

``` bash
curl https://api.github.com/repos/gcovr/gcovr/pulls/457 | jq -r .body
```

However, I have no strong preference regarding PR comment vs commit message as the source for these tags. My concern with commit messages would be that a PR with multiple commits could include both commits with changes that don't have to be mentioned in the Changelog, and other commits with changes that should be mentioned nevertheless. A PR message would be easier to change afterwards during review. The advantage of commit messages is that they can be checked locally and without network connectivity.

But I do like that this PR introduces more structure into the CI pipelines!

Spacetown:
>     1. Should the check be mentioned in https://gcovr.com/en/stable/contributing.html#how-to-submit-a-pull-request? Also the fact that adding `[no changelog]` can be added there for minor changes.

Of course.

>     2. Do i understand correctly that adding `[no changelog]` to any commit message skips the check? So it is not necessary to add this to all commits in a pr, only to one commit.

This was the intention but I find this idea better:
 

> As an alternative to looking at commit messages, we could consider looking at the pull request comment via the Github API. This would likely mean that the check would be performed only in the Action, not in the release checklist. 

Instead of using curl I'll try to use the available PR info in the pipeline and directly write the script there. I think the local check isn't realy needed because it is only needed for PR's and not for playing in the local clone.
With every approach we have the problem that the PR number is realy known after creating the PR.

.github/workflows/test.yml
```diff
@@ -11,9 +11,49 @@ on:
...
+  changelog-check:
+    runs-on: ubuntu-20.04
+    if: ${{ github.event_name == 'pull_request' }}
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run check
+      run: |
+        if [ ! -z "${{ github.event.pull_request.number }}" ]; then
+          if echo """${{ github.event.pull_request.body }}""" | grep -qE '^\[no\ changelog\]$'; then
```
latk, line 23: I think we can escape from escaping hell by using environment variables to pass data into the shell script.

Adding this to the step:

```yaml
env:
  PR_BODY: ${{ github.event.pull_request.body }}
```

should allow us to simplify this line to:

```bash
if grep -qE '^\[no\ changelog\]$' <<<"$PR_BODY"; then
```

(Whether we use `echo ... | grep` or a here-string `<<<WORD` is irrelevant though)

Spacetown: That's realy better. My first intention was to use a label instead of the PR body but the labels can only be added by members and not by contributors.

latk (approved):
LGTM :)

Thanks for implementing this! Automating such checks will help _future_ releases to be ready more quickly.

chrta (approved):

Spacetown:
@chrta @latk: For the main branch the test is turned of. I'll create a new PR in the afternoon.

gobater:
@Spacetown : please update documentation on this topic in https://gcovr.com/en/stable/contributing.html#how-to-submit-a-pull-request

Spacetown:
> @Spacetown : please update documentation on this topic in https://gcovr.com/en/stable/contributing.html#how-to-submit-a-pull-request

That's the documentation of the stable version. Please take a look at the latest source https://github.com/gcovr/gcovr/blob/master/CONTRIBUTING.rst#how-to-submit-a-pull-request

PR #458, opened 2021-02-01
Fix check on master (no PR context)

Spacetown:
Fix #457 on master branch.
Without a PR the changelog check is skipped and the dependent checks aren't executed.

[no changelog]

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,6 @@ jobs:
 
   changelog-check:
     runs-on: ubuntu-20.04
-    if: ${{ github.event_name == 'pull_request' }}
 
     env:
       PR_BODY: ${{ github.event.pull_request.body }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -13,17 +13,18 @@ jobs:
 
   changelog-check:
     runs-on: ubuntu-20.04
-    if: ${{ github.event_name == 'pull_request' }}
 
     env:
-      PR_BODY: ${{ github.event.pull_request.body }}
+      PR_BODY: ${{ github.event.pull_re_quest.body }}
```
latk, line 18: this change looks like a typo?

latk (approved):

.github/workflows/test.yml
Spacetown: This was my test if the pipeline aborts on unknown structure references because on master the pull_request isn't available.

Issue #325, opened 2019-07-25
How to get function coverage from .gcda ？


Toolazy2cruel:
We can add function coverage statistics in html.how to do?

latk:
Currently, gcovr does not support function coverage. In the meanwhile, you would have to use lcov for coverage reports, which does support this.

The underlying data from GCC/gcov does include per-function statistics, but right now gcovr just ignores those lines. A pull request that parses this data (without directly rendering it in reports) would be most welcome.

A similar features was requested in #68.

Toolazy2cruel:
Thank you very much for your answer, I will try to solve this problem and hope that interested partners can work together.

latk:
Thanks! If you'd like to try doing this, look at the [contribution guidelines](https://gcovr.com/en/master/contributing.html) for how to set up a development environment etc.

A couple of pointers:

* we'd have to add the -f/--function-summaries and -m/--demangled-names option when calling gcov ([around here](https://github.com/gcovr/gcovr/blob/1f5af01d1a347c917939df34082254544d85a5f1/gcovr/gcov.py#L636-L640))

* the function parsing code goes [here](https://github.com/gcovr/gcovr/blob/1f5af01d1a347c917939df34082254544d85a5f1/gcovr/gcov.py#L355-L356)

* A function line in the coverage data looks like this:

      function main called 1 returned 1 blocks executed 75%

  Note that the function name can look more complex, e.g. `Foo<int>::inc()`

* a FileCoverage object [only stores](https://github.com/gcovr/gcovr/blob/1f5af01d1a347c917939df34082254544d85a5f1/gcovr/coverage.py#L122-L128) LineCoverage. We'd have to add a new FunctionCoverage class similar to the LineCoverage and BranchCoverage objects. The FunctionCoverage would need to keep track of the function name, and the number of calls.

* the most up to date gcov documentation is here: https://gcc.gnu.org/onlinedocs/gcc/Gcov.html

k10blogger:
In older version of gcov tool(4.3.3) the `-m/ --demangled-names` option doesn't exists wouldnt it be better to pass these flags externally?

Toolazy2cruel:
> In older version of gcov tool(4.3.3) the `-m/ --demangled-names` option doesn't exists wouldnt it be better to pass these flags externally?

[...]

PR #459, opened 2021-02-02
Update docs and test.

Spacetown:
- Update the docs for running tests as mentioned in #325.
- Rename make target `format` to `check-format` and add it to the docs.

[no changelog]

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -142,7 +142,8 @@ How to set up a development environment
 
 For working on gcovr, you will need a supported version of Python 3,
 GCC version 5, 6 or 8 (other GCC versions are supported by gcovr,
-but will cause spurious test failures), ``make`` and ``cmake``.
+but will cause spurious test failures), ``make``, ``cmake`` and
+``ninja``.
 Please make sure that the tools are in the system ``PATH``.
 On **Windows**, you will need to install a GCC toolchain as the
 tests expect a Unix-like environment. You can use MinGW-W64 or MinGW.
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -141,8 +141,33 @@ How to set up a development environment
 ---------------------------------------
 
 For working on gcovr, you will need a supported version of Python 3,
-and GCC version 5. Other GCC versions are supported by gcovr,
-but will cause spurious test failures.
+GCC version 5, 6 or 8 (other GCC versions are supported by gcovr,
+but will cause spurious test failures), ``make`` and ``cmake``.
+Please make sure that the tools are in the system ``PATH``.
+On **Windows**, you will need to install a GCC toolchain as the
+tests expect a Unix-like environment. You can use MinGW-W64 or MinGW.
+An easier way is to :ref:`run tests with Docker <docker tests>`,
+on **Windows** a Pro license or the WSL (Windows subsystem for Linux)
+is needed.
+
+-   Check your GCC installation, the binary directory must be added to
+    the PATH environment. If the command gcc-5/g++-5/gcov-5,
+    gcc-6/g++-6/gcov-6 or gcc-8/g++-8/gcov-8 are available everything
+    is OK. For gcc-6 and gcc-8 you should use the option ``CC=...```
+    see :ref:`run and filter tests <run tests>`. If this isn't OK you
+    have to create symlinks for the gcc executables with the following steps.
+    You can check the GCC version with gcc --version. If the output says
+    version 8, you should also be able to run gcc-8 --version. Your Linux
+    distribution should have set all of this up already.
+    If you don't have an alias like gcc-8, perform the following steps to
+    create an alias for gcc, this should also work in the MSYS shell under Windows:
+
+    1. Create a directory somewhere, e.g. in your home directory: mkdir ~/bin
+    2. Create a symlink in that directory which points to GCC:ln -s $(which gcc) ~/bin/gcc-8
+    3. Add this directory to your PATH: export PATH="$HOME/bin:$PATH"
+    4. Re-test gcc-8 --version to ensure everything worked.
+    5. Create additional symlinks for g++ -> g++-8 and gcov -> gcov-8.
+
 
 -   (Optional) Fork the project on GitHub.
 
@@ -170,7 +195,7 @@ but will cause spurious test failures.
 
     See ``doc/README.txt`` for details on working with the documentation.
 
--   (Optional) Activate GitHub and Appveyor for your forked GitHub repository,
+-   (Optional) Activate GitHub for your forked GitHub repository,
     so that the cross-platform compatibility tests get run
     whenever you push your work to your repository.
     These tests will also be run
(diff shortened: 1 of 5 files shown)


Spacetown:
@gobater Is this OK?

CONTRIBUTING.rst
```diff
@@ -141,13 +141,17 @@ How to set up a development environment
...
+and GCC version 5, 6 or 8. Other GCC versions are supported by gcovr,
 but will cause spurious test failures.
 
 -   (Optional) Fork the project on GitHub.
 
 -   Clone the git repository.
 
+-   (Optional) If your compiler executable is gcc have to create a
+    symbolic link or copy with the version of the GCC. E.g. if it is GCC
+    version 8 you have to create a gcc-8.
```
latk, line 153: It is not clear to me what this step is instructing. Presumably the goal is that there's an executable called `gcc-8` in the PATH, correct? I have no idea how to do that on Windows, but we could at least add simple Linux/MacOS instructions. Something like:

> If you're on Linux or MacOS, you can check the GCC version with `gcc --version`. If the output says version 8, you should also be able to run `gcc-8 --version`. Your Linux distribution should have set all of this up already.
>
> If you don't have an alias like `gcc-8`, perform the following steps to create an alias for `gcc`:
>
> 1. Create a directory somewhere, e.g. in your home directory: `mkdir ~/bin`
> 2. Create a symlink in that directory which points to GCC: `ln -s $(which gcc) ~/bin/gcc-8`
> 3. Add this directory to your PATH: `export PATH="$HOME/bin:$PATH"`
> 4. Re-test `gcc-8 --version` to ensure everything worked.
> 5. Create additional symlinks for `g++` :arrow_right: `g++-8` and `gcov` :arrow_right: `gcov-8`.

Alternatively, we could just promote the Docker-based workflow as the default environment, and essentially deprecate running the tests on the host. Anecdotally there can be some issues with getting Docker to run on Windows, but once that's out of the way the rest is much simpler.

Spacetown: Great instruction to add symliks. This should also work in the MSYS environment under Windows. I'll use this instruction.

gobater: Probably, this issue is a mingw/cygwin/msys2 only issue. This seems to be a problem only for a few tests. I'm wondering why the other tests work
The symlink instructions are not "complete" since in a normal linux pc, you have all gcc related binaries with the "dash version" naming, including:
* gcc
* gcc-nm
* gcc-ar
* g++
* gcov
* gcov-tool
and probably many I have forgotten...

so, is there any possibility to fix this in the two tests that are failing?

CONTRIBUTING.rst
```diff
@@ -179,14 +199,13 @@ but will cause spurious test failures.
...
--   for Windows: ``.github/workflows/test.yml`` and ``appveyor.yml``
+-   for Windows: ``.github/workflows/test.yml``
 
 On **Windows**, you will need to install a GCC toolchain
 as the tests expect a Unix-like environment.
 You can use MinGW-W64 or MinGW.
-To run the tests,
-please make sure that the ``make`` and ``cmake`` from your MinGW distribution
-are in the system ``PATH``.
+To run the tests, please make sure that the ``make`` and ``cmake`` from
```
gobater, line 207: do we also need cmake?

gobater:
For my PR (https://github.com/gcovr/gcovr/pull/362) the references need to be updated (for all GCC versions) since a new feature is added, that affects almost all report formats.

Is there any official procedure to do that?

CONTRIBUTING.rst
Spacetown: Yes, there are also some tests for CMake.

gobater: ninja is missing as a requirement...

CONTRIBUTING.rst
```diff
@@ -141,8 +141,34 @@ How to set up a development environment
...
+    version 8, you should also be able to run gcc-8 --version. Your Linux
+    distribution should have set all of this up already.
+    If you don't have an alias like gcc-8, perform the following steps to
+    create an alias for gcc, this should also work in the MSYS shell under Windows:
+
+    1. Create a directory somewhere, e.g. in your home directory: mkdir ~/bin
+    2. Create a symlink in that directory which points to GCC:ln -s $(which gcc) ~/bin/gcc-8
+    3. Add this directory to your PATH: export PATH="$HOME/bin:$PATH"
+    4. Re-test gcc-8 --version to ensure everything worked.
+    5. Create additional symlinks for g++ -> g++-8 and gcov -> gcov-8.
```
latk, line 170: A bit of code highlighting should make the resulting document more readable.

```suggestion
    1. Create a directory somewhere, e.g. in your home directory: ``mkdir ~/bin``
    2. Create a symlink in that directory which points to GCC: ``ln -s $(which gcc) ~/bin/gcc-8``
    3. Add this directory to your PATH: ``export PATH="$HOME/bin:$PATH"``
    4. Re-test ``gcc-8 --version`` to ensure everything worked.
    5. Create additional symlinks for g++ -> g++-8 and gcov -> gcov-8.
```

CONTRIBUTING.rst
```diff
@@ -170,7 +196,7 @@ but will cause spurious test failures.
 
     See ``doc/README.txt`` for details on working with the documentation.
 
--   (Optional) Activate GitHub and Appveyor for your forked GitHub repository,
+-   (Optional) Activate GitHub for your forked GitHub repository,
     so that the cross-platform compatibility tests get run
     whenever you push your work to your repository.
     These tests will also be run
```
latk, line 202: This item can be deleted, or replaced by a mention that the tests WILL be run automatically when using GitHub, no activation needed. Well, this might require #471 to be merged first :)

Makefile
```diff
@@ -69,12 +74,12 @@ endif
 endif
 	$(GCOV) --version
 
-qa: lint black test doc
+qa: doc #lint check-format test doc
```
latk, line 77: Is it intended to comment out QA tasks other than doc?

```suggestion
qa: doc lint check-format test doc
```

CONTRIBUTING.rst
```diff
@@ -225,6 +240,8 @@ The QA process (``make qa``) consists of multiple parts:
 
 - linting (``make lint``)
 
+- checking format (``make format``)
```
latk, line 243: Was renamed.

```suggestion
- checking format (``make check-format``)
```

And we should perhaps mention that this step won't yet succeed…

latk (approved):
I think we can merge this soonishly, but please see some smaller remarks below.

latk:
@gobater The `--update_references` mechanism is mentioned in the section on running the tests: <https://gcovr.com/en/master/contributing.html#run-and-filter-tests>. But perhaps it could be made more explicit in the “how to submit a pull request” section that updated files will be available for download from the CI system.

CONTRIBUTING.rst
Spacetown: Now I can remove it. :-)

Spacetown: I'll remove also the AppVeyor CI batch and all other references to AppVeyor from doc.

PR #460, opened 2021-02-02
Check if milestone is added to PR.

Spacetown:
Extend check for selected milestone to remind project member to add PR to the current milestone.

[no changelog]

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,19 +8,39 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+    types: [opened, synchronize, reopened, edited]
 
 jobs:
+
+  milestone-check:
+    runs-on: ubuntu-20.04
+
+    continue-on-error: true
+
+    env:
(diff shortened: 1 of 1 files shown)


chrta (approved):
LGTM, but i am no expert.

latk:
Should we really make the test matrix conditional on these quick checks like milestone, release-checklist, changelog? That way, the PR submitter will not receive CI feedback until those aspects are fixed. But the milestone can't be selected by external contributors, which could lead to delays.

Spacetown:
The missing milestone doesn't skip the test, it's only showing as failed but the rest is executed. 
The release-check was also mandatory in the past and the change log check can be fixed by the contributor.  We save a lot of computing time if we fail early (7min vs. 15sec).

latk:
Ok, I see from the first Actions run on this PR (https://github.com/gcovr/gcovr/actions/runs/531822835) that the milestone check will not block the other tests. I was not sure how the `continue-on-error: true` option would interact with `needs: [milestone-check, ...]`.

Issue #391, opened 2020-06-07
Vote: giving everyone* write access


latk:
For the last two years of gcovr development there have been many great contributions from the community, but I haven't always managed to stay atop of things and keep things flowing smoothly. I am therefore looking for ways to become less of a bottleneck, in particular by giving out write access more actively.

Proposal:
* Starting 2020-06-15, I will begin inviting previous contributors to the @gcovr organization, which will give them **write access** to the gcovr/gcovr repository.
* Instead of a heavyweight process, I'll act at my discretion to decide whether giving someone write access is appropriate. E.g. I expect that nearly all pull requests indicate that the contributor would use write access for good.
* This policy will primarily trigger on future contributions, but I also expect to look at some existing contributors.

Per the voting procedure suggested in #176, all **existing members** (@gcovr/developers, @gcovr/owners) are asked to **cast a vote** on this proprosal by writing a comment. Voting closes after one week, at 2020-06-15 12:00 UTC.

Details / Background / Safeguards:

* People with write access (“developers”) can manage issues, review pull requests, merge them, and cut releases.

* The master branch is a protected branch, meaning that developers cannot push to it directly. Pull requests can only be merged if a different developer approves the PR. This serves as a safeguard to prevent abuse of write access.

* The gcovr project has yet to adopt a formal Code of Conduct. In the meanwhile:
  * Everyone (whether developer or other) is expected to act respectfully.
  * Developers act on behalf of the project and must hold themselves to a higher standard.
  * Trolling and rule-lawyering is not welcome in this context.
  * I pledge to investigate and sanction unacceptable behaviour with due care, and can be reached at opensource at lukasatkinson.de (please mention `[CoC]` in the subject).
  * This is not subject to the vote because I take this for granted.

* There previously was a [process](https://gcovr.com/en/4.2/contributing.html#become-a-gcovr-developer) to get developer access, but it was never invoked. If this proposal is accepted, the old process will be retired.

Spacetown:
I agree with this. What I'm missing is how new releases are planned.

jsiirola:
+1

I agree with @Spacetown that there should be a written release process (mostly to document who can initialize the process and who has actual rights / authority to “push” the release), but I am not entirely sure that that needs to be handled here.

latk:
The vote passed unanimously on Jun 15.

[...]

PR #461, opened 2021-02-02
Update issue templates

Spacetown:
Add issue templates

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,7 +2,7 @@
 name: Bug report
 about: Create a report to help us improve
 title: ''
-labels: ''
+labels: 'Type: Bug'
 assignees: ''
 
 ---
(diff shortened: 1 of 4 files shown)


latk (approved):
Thank you for adding these, this should be very helpful to users.

Looking at the labels we're using, it's also fairly common to get *questions* about using gcovr – those are not a feature request, though they sometimes turn into one. So we could perhaps add another template for that. Aside from OS/version the question template should also ask for a project directory layout, which is critical for understanding most questions related to filters, roots, object-directory, and so on.

Spacetown:
@latk It's not possible to extend the PR with the Web interface. I hope the added template for general questions is working.

latk:
Yup, the template looks like it should work.

I will not select a specific CoC document without a vote. I've outlined an interim CoC in #391. I haven't previously advanced such a vote because of the “who watches the watchers” problem – who can handle complaints when I am the subject? This problem should resolve itself in a week when we have two active admins.

chrta (approved):

PR #418, opened 2020-09-16
Change Version numbers to strings

Jugst3r:
Quick fix, tests should be clean

PR #463, opened 2021-02-03
Change JSON format version to "0.2"

### LINKED ISSUE #462: Assertion failure: "gcovr_json_data['gcovr/format_version'] == JSON_FORMAT_VERSION"
After installing the latest gcovr from git, I cannot run gcovr like I could with the installed 4.2 package installed by pip.
`pip3 install --upgrade --user git+https://github.com/gcovr/gcovr.git`
After installing, I get this during usage:
```
+ gcovr -r /code --add-tracefile full_json_output.json --html --html-details -o report/index.html
  File "/usr/local/bin/gcovr", line 8, in <module>
    sys.exit(main())
  File "/home/user/.local/lib/python3.6/site-packages/gcovr/__main__.py", line 273, in main
    collect_coverage_from_tracefiles(covdata, options, logger)
  File "/home/user/.local/lib/python3.6/site-packages/gcovr/__main__.py", line 306, in collect_coverage_from_tracefiles
    gcovr_json_files_to_coverage(datafiles, covdata, options)
  File "/home/user/.local/lib/python3.6/site-packages/gcovr/json_generator.py", line 135, in gcovr_json_files_to_coverage
    assert gcovr_json_data['gcovr/format_version'] == JSON_FORMAT_VERSION
```

Spacetown:
Fix for #462, the old format uses a 0.1 and the new format a version "0.1" which is incompatible in an assert.

Closes #462.


latk:
Hmm, we should probably define compatibility guarantees for the JSON format. I've never written it down, but my expectation used to be that the JSON is stable-ish for downstream consumers, but that gcovr will only read merge JSON files that were created by the same version – otherwise we'd have to maintain substantial backwards compatibility features in the future.

In that sense, an alternative solution to this PR would be to bump the format version to 0.2 just because the version field representation changed.

Spacetown:
@latk The compatibility should be mentioned in the docs but here the only change is the version representation and my suggestion is to keep the version.
I think we should also extend the assert with an explanation.
What about replacing the manual changed version string with the gcovr version if the user shouldn't use trace files generated by the old version?

CezaryGapinski:
@Spacetown 
> What about replacing the manual changed version string with the gcovr version if the user shouldn't use trace files generated by the old version?

Do you think the JSON format will be changed so frequently? I think it wouldn't be so frequent so we can be keep it separate to the gcovr version.

IMHO the version can be changed to 0.2. Previous representation of `gcovr/format_version` is in the current release and change from #418 is only on current master branch, so now is a chance to change it to explicit show this small difference with string representation in the next release.

Spacetown:
Ok. I'll change the version and add an explanation to the assert. The test can also be removed.

CezaryGapinski (approved):

PR #465, opened 2021-02-04
Use setuptools.find_packages() in setup.py (fixes #464)

### LINKED ISSUE #464: gcovr.writer module missing after installation
Latest master is non-functional after installation with setuptools/pip (due to the recent change in https://github.com/gcovr/gcovr/pull/443):

```
ModuleNotFoundError: No module named 'gcovr.writer'
```

To reproduce simply install gcovr from the current master using pip and the resulting installation in `site-packages` will be missing the `writer` package:
```
$ ls -1 /usr/local/lib/python3.6/dist-packages/gcovr/
configuration.py
coverage.py
gcov.py
__init__.py
__main__.py
__pycache__
templates
utils.py
version.py
workers.py
```

A simple patch that replaces fixed `packages` value in `setup.py` with `find_packages()` call from setuptools does resolve the problem:
```diff
diff --git a/setup.py b/setup.py
index 943356e..c73db7c 100644
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ Script to generate the installer for gcovr.
 """
 
 from runpy import run_path
-from setuptools import setup
+from setuptools import setup, find_packages
 
 
 version = run_path('./gcovr/version.py')['__version__']
@@ -30,7 +30,7 @@ setup(name='gcovr',
       version=version,
       platforms=["any"],
       python_requires='>=3.5',
-      packages=['gcovr'],
+      packages=find_packages(),
       install_requires=[
           'jinja2',
           'lxml',
```

avtolstoy:
Given recent refactoring done in #443, use `setuptools.find_packages()` to generate the list of packages for installation.
Fixes #464

[no changelog]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ Script to generate the installer for gcovr.
 """
 
 from runpy import run_path
-from setuptools import setup
+from setuptools import setup, find_packages
 
 
 version = run_path('./gcovr/version.py')['__version__']
@@ -30,7 +30,7 @@ setup(name='gcovr',
       version=version,
       platforms=["any"],
       python_requires='>=3.5',
-      packages=['gcovr'],
+      packages=find_packages(include=['gcovr*'], exclude=['gcovr.tests']),
       install_requires=[
           'jinja2',
           'lxml',


latk (approved):

latk:
Github Actions seems to be having a bad day, but given that this is a high priority / low risk change I'm merging anyway.

Thanks for fixing this!

avtolstoy:
Thanks for a quick turnaround on this as well!

Spacetown:
@latk Can we test this without deploying?

latk:
@Spacetown I think so: build a wheel in one Action job, save the wheel as an artifact, then try installing the wheel in another job and do a quick smoke test (e.g. that an empty JSON file can be turned into a text report, or that the help message works).

PR #471, opened 2021-02-08
Make CI run successfully on all branches

latk:
This PR contains two changes.

First, the **milestone check** only makes sense on pull requests, so it's now limited to events of that type. Previously, this would cause the CI to fail on the master branch.

Second, running the CI is also useful on **branches other than master**. In particular, this lets contributors run CI on their forked repositories.

Since this is a purely internal change relating to #460:
[no changelog]

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,9 +5,7 @@ name: Test
 
 on:
   push:
-    branches: [ master ]
   pull_request:
-    branches: [ master ]
     types: [opened, synchronize, reopened, edited]
 
 jobs:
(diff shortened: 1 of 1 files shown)


Spacetown (approved):
LGFM. Running on every branch is good for private testing in the fork without creating a PR in the main repository.
One question: If you have a PR in the main branch and update the branch the CI is triggered in the main repository for the PR and in the fork for the branch (if configured). Is it possible to supress the CI in the fork?

latk:
> If you have a PR in the main branch and update the branch the CI is triggered in the main repository for the PR and in the fork for the branch (if configured). Is it possible to supress the CI in the fork?

Hmm, I see that running the CI on the branch is unnecessary when already running CI as part of the PR. But I don't think GH is smart enough to understand this, especially since the push and pull request actions could be completely unrelated.

Unfortunately, this seems like one of those cases where it's so much easier to just burn the electricity than figuring out a good solution.

PR #473, opened 2021-02-11
Fix documentation generation for Python 3.6.

Spacetown:
Because of https://github.com/sphinx-contrib/autoprogram/issues/42 version is pinned for Python 3.6.

[no changelog]

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,3 +1,4 @@
 sphinx
 sphinx_rtd_theme
-sphinxcontrib-autoprogram>=0.1.5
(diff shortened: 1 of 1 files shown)


latk (approved):
A most sensible fix until upstream issues are sorted out.

Should upstream drop support for older Python versions, that won't be a drama – we could limit documentation builds to newer Pythons only.

I have checked which Python version is used on ReadTheDocs. It is not possible to select a more specific version than “3.x”, but the build logs show 3.7. So things should be fine there.

Spacetown:
Skip the documentation in 3.6 was my first thought but I think pinning the version is better. I was wondering because the version of Sphinx was the same as in the old build.

PR #474, opened 2021-02-11
Add abstract interface for reader/writer.

Spacetown:
Ass discussed in #443 a object oriented interface is added.

- The writer `xml.py` is renamed to `coveralls.py` and a option was added. I think we should deprecate the option `--xml` because `sonarcube.py` also produce a XML file.
- `add_copyright.py` was changed to generate black conform files.
- Move generator utils to a seperate file.
- Add reader for JSON files.
- Move the check functions and `OutputOrDefault` to `GcovrConfigOption` to access them eays in the readers and writers.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,9 @@ Documentation:
 
 Internal changes:
 
+ - Move generators/writer to sub directory. (:issue:`443`)
+ - Add base class for reader and writers. Move the format specific stuff (CLI options, checks, ...) to the
+   derived class. (:issue:`474`)
  - Clean GCC environment variables in test suite. (:issue:`493`)
  - Fix problems from deployment of release 5.0. (:issue:`494`)
  - Use yaxmldiff for XML diffing in tests. (:issue:`495`)
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ endif
 endif
 	$(GCOV) --version
 
-qa: doc lint check-format test doc
+qa: lint check-format test doc
 
 lint:
 	$(PYTHON) -m flake8 doc gcovr
--- a/admin/add_copyright.py
+++ b/admin/add_copyright.py
@@ -89,8 +89,10 @@ def addCopyrightToPythonFile(filename, lines):
     if headerEndReached:
         for line in iterLines:
             if line != "":
-                # Use one empty line
+                # Use one empty line or two for classes
                 newLines.append("")
+                if line.startswith("class"):
+                    newLines.append("")
                 newLines.append(line)
                 break
         # keep all other lines
--- a/gcovr/__main__.py
+++ b/gcovr/__main__.py
@@ -25,28 +25,22 @@ from argparse import ArgumentParser
 from os.path import normpath
 from tempfile import mkdtemp
 from shutil import rmtree
-from glob import glob
 
 from .configuration import (
     argument_parser_setup, merge_options_and_set_defaults,
-    parse_config_file, parse_config_into_dict, OutputOrDefault)
+    parse_config_file, parse_config_into_dict)
 from .gcov import (find_existing_gcov_files, find_datafiles,
                    process_existing_gcov_file, process_datafile)
-from .utils import (get_global_stats, AlwaysMatchFilter,
+from .utils import (AlwaysMatchFilter,
                     DirectoryPrefixFilter, Logger)
 from .version import __version__
 from .workers import Workers
 
-# generators
-from .writer.json import (gcovr_json_files_to_coverage)
-from .writer.cobertura import print_xml_report
-from .writer.html import print_html_report
-from .writer.json import print_json_report, print_json_summary_report
-from .writer.txt import print_text_report
-from .writer.csv import print_csv_report
-from .writer.summary import print_summary
-from .writer.sonarqube import print_sonarqube_report
-from .writer.coveralls import print_coveralls_report
+# Readers
+from .reader import Readers
+# Writers
+from .writer import Writers
+from .writer.utils import get_global_stats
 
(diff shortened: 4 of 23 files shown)


Spacetown:
@latk I hope this is a little step to get your dream come true.

latk:
I've looked through the code, and it has some interesting ideas – the main() is much better now! It was a bit hard to review though, because it all consists of a single commit that moved a lot of code around *and* reformatted it.

I want to discuss two high-level aspects before doing a more thorough review:

* consequences from dividing/grouping code one way or the other
* data and control flow

## The matrix problem, and potential consequences.

When dividing up functionality, there are different approaches with different consequences. I want us to think this through before committing to any alternative.

If we have N components with M concerns each, we somehow have to divide up these different aspects. I refer to that as a “matrix problem” since it's usually equally reasonable to separate aspects by row or by column:

|  | json | html | txt | … |
|---|---|---|---|---|
| arguments |||||
| validation |||||
| writing |||||

A general heuristic for solving this problem is: what changes together, stays together. This would suggest that the presented division by output format is correct, since arguments are likely to change together with the current writer (if at all).

Such a division might also enable pluggable writers, some way down the road. However, we are *far* from offering a sufficiently stable plugin interface, so I'm not particularly concerned about this.

Dividing by concern would enable something entirely different: loading writers only on demand. In the presented architecture, all writers must be loaded up front in order to discover available arguments. A throwaway experiment I tried out some time ago suggested opportunity for noticeably faster startup when lazy-loading writers (this wasn't previously possible due to utilities from the HTML writer being used from other components). I don't want to give up this potential performance win unless we are sure that it's worth it.

## Data flow and control flow

In the proposed design, argument validation logic is moved to the various writer modules. There are two aspects that make me a bit uncomfortable.

First, all writers operate on the shared `options` object which has long been a problem due to unclear interface boundaries. But now, the other options/accesses are less visible because they are in different files. I would tend to leave the validation as it is until a clear interface boundary can be introduced, e.g. if each writer has a constructor that extracts the required arguments into member fields.

Secondly, control flow. The validation code and some of the writer code contain application-level control flow. There should not be a `sys.exit()` outside the main – that is a CLI concern, not a writer concern. If we want to move this validation code, we should rather think about suitable exceptions.

## Conclusion

All in all, I'm really thankful for this PR because it enables us to have a conversation about future directions. But it's a lot of change with little discussion so I'm hesitant to say “yes” outright. On the other hand, just moving code around is a fairly low risk change.

What do you think? Which future directions would you prioritize with the “matrix problem”? Which changes can we focus on in this PR, what should be deferred to another iteration?

Spacetown:
> I've looked through the code, and it has some interesting ideas – the main() is much better now! It was a bit hard to review though, because it all consists of a single commit that moved a lot of code around _and_ reformatted it.

Mea culpa! Next time I'll try to create several commits.

> A general heuristic for solving this problem is: what changes together, stays together. This would suggest that the presented division by output format is correct, since arguments are likely to change together with the current writer (if at all).

That's also my prefered way. I don't want to add the options in file A, add the option handling in file B and the code to write the output in file C. Another pro for this way is that if you only change the HTML writer you can be sure that the other formats aren't affected. 
 
> Such a division might also enable pluggable writers, some way down the road. However, we are _far_ from offering a sufficiently stable plugin interface, so I'm not particularly concerned about this.

I'm with you, this can be done in the far future.

> Dividing by concern would enable something entirely different: loading writers only on demand. In the presented architecture, all writers must be loaded up front in order to discover available arguments. A throwaway experiment I tried out some time ago suggested opportunity for noticeably faster startup when lazy-loading writers (this wasn't previously possible due to utilities from the HTML writer being used from other components). I don't want to give up this potential performance win unless we are sure that it's worth it.

A improvement can be to do lazy loading of the parts needed for writing like it's done in the HTML writer.

> In the proposed design, argument validation logic is moved to the various writer modules. There are two aspects that make me a bit uncomfortable.
> 
> First, all writers operate on the shared `options` object which has long been a problem due to unclear interface boundaries. But now, the other options/accesses are less visible because they are in different files. I would tend to leave the validation as it is until a clear interface boundary can be introduced, e.g. if each writer has a constructor that extracts the required arguments into member fields.

Do you mean the `check_options` of the reader/writeer?
 
> Secondly, control flow. The validation code and some of the writer code contain application-level control flow. There should not be a `sys.exit()` outside the main – that is a CLI concern, not a writer concern. If we want to move this validation code, we should rather think about suitable exceptions.

I was also not so happy with this but I didn't want to change the code as much. I can use an ArgumentException here.

> All in all, I'm really thankful for this PR because it enables us to have a conversation about future directions. But it's a lot of change with little discussion so I'm hesitant to say “yes” outright. On the other hand, just moving code around is a fairly low risk change.
>
> What do you think? Which future directions would you prioritize with the “matrix problem”? Which changes can we focus on in this PR, what should be deferred to another iteration?

I prefer to keep all code related to a format in one file. For a new writer you only have to implement the functions of the base class and add it to the global list.
I think follwing points should be part of this PR:
- Replace sys.exit by exception handling in all readers/writers.
- Add a "namespace" to all option variables, e.g. option `--html-absolute-paths` should set `html_relative_anchors` instead of `relative_anchors`.

Separate PR:
- Change the gcov stuff to a reader.
- Implement lazy loading to improve startup performance.
- Maybe add an option group for each writer writer options except the option activating the writer, e.g. `--html`, `--html-details`and `--html-details-syntax-highlighting` will stay in group `output_options` all other options are moved to group `html_options`.

Spacetown:
@latk What do you think about the proposal? Should I add lazy loading for the readers with this PR?

Spacetown:
This PR has has to many conflicts and should be started from scratch.

PR #480, opened 2021-03-24
Update CSS for meter in WebKit (Crome).

### LINKED ISSUE #475: No longer colored columns
I'm testing out the new version of gcovr, last night's @master revision, and I find the HTML tables are missing all the low/med/high coloring. Also, the progress-bars don't look like progress bars. 

Here's a cropped up comparison.

image

Spacetown:
Fix #475 
[no changelog]

### DIFF
--- a/gcovr/templates/style.meter.css
+++ b/gcovr/templates/style.meter.css
@@ -7,7 +7,7 @@ meter {
     width: 30vw;
     min-width: 4em;
     max-width: 15em;
-    height: 12px;
+    height: 0.75em;
     padding: 0;
     vertical-align: baseline;
     margin-top: 3px;
@@ -22,7 +22,8 @@ meter::-webkit-meter-bar {
     /* Outer background for Webkit */
     background: none;
     background-color: whitesmoke;
-    height: 12px;
+    height: 0.75em;
+    border-radius: 0px;
 }
 
 meter::-webkit-meter-optimum-value,
--- a/gcovr/templates/style.meter.css
+++ b/gcovr/templates/style.meter.css
@@ -3,7 +3,6 @@
 /* Common */
 meter {
     -moz-appearance: none;
-    -webkit-appearance: none;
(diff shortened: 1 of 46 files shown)


chrta:
I can verify that it improves rendering on my machine (ubuntu 20.04, chrome 90.0.4430.212). I just opened the checked in file gcovr/tests/html-default/reference/gcc-5/coverage.html

Master branch without this fix:
chrome-master

File from this branch:
chrome-css-fix

Spacetown:
@chrta I'm still not happy with the result in Chrome compared to Firefox or Safari. In my opinion the hight of the green part is to small.

chrta:
@Spacetown You are right, it does not look very good.
When i add this via developer tool to the css:
```
::-webkit-meter-bar {
    height: 10px;
}
```
i can modify the height of the bar. Since i do not know css, i do not know if this would we a good idea to add this into our css.
Result:
chrome-css-modify

Spacetown:
@chrta Great, the normal height didn't work.

Spacetown:
There is only a difference in the corners.

Firefox:
grafik

Safari:
grafik

Chrome:
grafik

chrta:
Yes, it looks the same way in my chrome, too. I managed to modify the edges by adding border-radius property:
```
meter::-webkit-meter-bar {
    /* Outer background for Webkit */
    background: none;
    background-color: whitesmoke;
    height: 12px;
    border-radius: 0px;
}
```

Then it looks the same in chrome as it does in firefox.

Spacetown:
😖 I've checked this but at the wrong element in the debugger.
Now it looks the same in all browsers and if the user uses a smaller or bigger font it resizes it's height. On my system the default font is 16px and the 0.75em gives the previous height of 12px.

chrta (approved):
Now it looks the same for me in firefox and chrome. :+1:

Issue #134, opened 2016-08-05
Add tests for llvm


whart222:

Spacetown:
Test environment extended in #484

PR #484, opened 2021-05-02
Extend tests with clang-10

Spacetown:
Extend test suite with clang-10 inside docker container and add reference data for clang-10 based on gcc-8.
As far as I can see, gcc adds a branch to closing braces which is missing in clang-10.


Makefile
```diff
@@ -90,6 +90,10 @@ test: export CXXFLAGS := --this_flag_does_not_exist # Env removed in text_gcovr.
 test: export GCOV := $(GCOV)
 
 test:
+ifneq ($(wildcard /Users/Shared/Informatik/GitHub/gcovr/gcovr/tests/linked/subdir/loop),)
+	# Remove directory because test is not running if it exists
+	rm -Rf /Users/Shared/Informatik/GitHub/gcovr/gcovr/tests/linked/subdir/loop
```
chrta, line 95: Should this block really be checked in?

chrta (commented):
Just one question

Spacetown:
Closes #134

gcovr/gcov.py
```diff
@@ -29,13 +29,17 @@
...
 
 exclude_line_flag = "_EXCL_"
 exclude_line_pattern = re.compile(r'([GL]COVR?)_EXCL_(START|STOP)')
 
 c_style_comment_pattern = re.compile(r'/\*.*?\*/')
 cpp_style_comment_pattern = re.compile(r'//.*?$')
 
+gcov_cmd = None
+gcov_options = None
+
```
chrta, line 42: I would prefer if this could be implemented without global variables.
But i am not sure where to put these cached variables.

This is a non-blocking comment, if there is no easy/better way to do tis.

chrta (changes requested):
In general this looks good to me. Just two minor questions/remarks.

gcovr/gcov.py
Spacetown: I think for this the command (this function) must be wrapped in a class with static members. I can take a deeper look after my holiday.

gcovr/gcov.py
```diff
@@ -750,6 +762,7 @@ def run_gcov_and_process_files(
             stderr=subprocess.PIPE).communicate()
         out = out.decode('utf-8')
         err = err.decode('utf-8')
+        print(out)
```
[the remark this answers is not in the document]
Spacetown: I'll remove the print.

Makefile
```diff
@@ -90,6 +90,10 @@ test: export CXXFLAGS := --this_flag_does_not_exist # Env removed in text_gcovr.
 test: export GCOV := $(GCOV)
 
 test:
+ifneq ($(wildcard $(dir $(word 1,$(abspath $(MAKEFILE_LIST))))gcovr/gcovr/tests/linked/subdir/loop),)
+	# Remove directory because test is not running if it exists
+	rm -Rf $(dir $(word 1,$(abspath $(MAKEFILE_LIST))))gcovr/gcovr/tests/linked/subdir/loop
+endif
```
latk, line 96: Why does this have to be done in the top-level makefile? Can't this be moved to the Makefile for that specific test?

doc/examples/example_cmake.sh
```diff
@@ -26,7 +26,7 @@ exec >&2  # redirect output to STDERR
 
 #BEGIN cmake_build
 cd $BLD_DIR
-cmake $SRC_DIR
+cmake -DCMAKE_BUILD_TYPE=PROFILE $SRC_DIR
```
latk, line 29: What does this do? This snippet is included in the documentation (cookbook section), so any unusual options should be explained there.

gcovr/gcov.py
```diff
@@ -604,6 +605,73 @@ def check_unrecognized_lines(self, ignore_parse_errors):
         sys.exit(1)
 
 
+class gcovr_executer:
```
latk, line 608: Other classes use CamelCase.
```suggestion
class GcovrExecuter:
```

gcovr/gcov.py
```diff
@@ -604,6 +605,73 @@ def check_unrecognized_lines(self, ignore_parse_errors):
...
+    gcov_cmd = None
+    gcov_options = None
+
+    def __init__(self, options):
+        # If the first element of cmd - the executable name - has embedded spaces
+        # (other than within quotes), it probably includes extra arguments.
+        if gcovr_executer.gcov_cmd is None:
+            gcovr_executer.gcov_cmd = shlex.split(options.gcov_cmd)
+            gcovr_executer.gcov_options = [
```
latk, line 617: Do these really have to be class variables (globals)? The code looks like assigning to `self` fields would work.

gcovr/tests/html-source-encoding-cp1252/Makefile
```diff
@@ -1,5 +1,5 @@
 all:
-	$(CXX) -fprofile-arcs -ftest-coverage -fPIC -finput-charset=cp1252 main.cpp -o testcase
+	$(CXX) -fprofile-arcs -ftest-coverage -fPIC $(if $(patsubst clang++%,,$(CXX)),-finput-charset=cp1252,) main.cpp -o testcase
```
latk, line 2: If Clang just ignores alternative source encodings, it might be better to skip this test. After all, this test is more about our HTML rendering than the compiler or the gcov tool. Currently, test skipping has to be configured in the `test_gcovr` file.

gcovr/tests/simple1/Makefile
```diff
@@ -7,7 +7,7 @@ endif
 ifeq ($(BASE_OS),MSYS_NT)
   FAIL_UNDER_LINE := 76.9
 else
-  FAIL_UNDER_LINE := $(if $(filter $(CC),gcc-6 gcc-8),76.9,79.9)
+  FAIL_UNDER_LINE := $(if $(filter $(CC),clang-10),72.7,$(if $(filter $(CC),gcc-8),77,80))
```
latk, line 10: at some point, there has to be a better way than Make's Lispish conditionals. I think it might work to put the limits into files in the reference directory and then run `FAIL_UNDER_LINE := $(shell cat reference/$(CC)/FAIL_UNDER_LINE)`.

latk (approved):
Testing with Clang would be *really good*! Since everything seems to work, this could be merged as is.

But it seems this kind of work touches on lots of tangential concerns. I've added some inline comments below. It could also make sense to mention the Clang tests in `installation.rst`.

I have only looked at the code, not at the generated reference files. So I'm assuming/hoping that the coverage data is reasonable.

gcovr/gcov.py
Spacetown: First I used global variables to safe time because clang doesn't understand the `--demangle-names`.
Maybe it's better to revert this part and spend the time to build the commandline several times. I think the safed time is less in comparsion to the time gcov needs.

gcovr/tests/html-source-encoding-cp1252/Makefile
Spacetown: Ok, I'll skip the test.

gcovr/tests/simple1/Makefile
Spacetown: Good idea, I'll try this.

doc/examples/example_cmake.sh
Spacetown: In the existing example the build is always done with profiling:
https://github.com/gcovr/gcovr/blob/e0b7afef00123b7b6ce4f487a1c4cc9fc60528bc/doc/examples/CMakeLists.txt#L6

To build without profiling you need to change CMakeLists.txt. For this purpose CMake has predefined [build types](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) which can be extended by user defined ones. Now you can define additional options only for this build type (e.g. `-DCMAKE_BUILD_TYPE=PROFILE`):
https://github.com/gcovr/gcovr/blob/cb8ba6549639463c723246d7e090547ab3c9c1b7/doc/examples/CMakeLists.txt#L7-L8

If you have different toolchains for a cmake you can define the profiling flags in the toolchain file which can be reused in different projects and the project doesn't need to know the flags for each toolchain.

gcovr/gcov.py
latk: Rebuilding the command is essentially free, especially compared to I/O like launching a gcov process.

Using a class like this is probably a good idea, but only if it simplifies things. I find the data flow from `options` to the *global* `gcov_cmd` and `gcov_options` variables very confusing. Potentially, it might be possible to change the class variables to instance variables and to push the `executor` construction further outward, perhaps to the main(). This would have the same caching effect, clear data flows, but would involve extra function parameters to pass the executor around.

doc/examples/example_cmake.sh
latk: aah, OK. I had filtered out any .txt reports for reviewing, but this accidentally prevented me from seeing your change in CMakeLists.txt. Thanks!

gcovr/gcov.py
Spacetown: I think this is a bigger change and something for the future. E.G. to define a interface for different input formats. I'll revert the change here.

gcovr/gcov.py
Spacetown: This class will be removed, see next comment.

Makefile
Spacetown: Because pytest collection phase has an infinite loop if this directory exists from a previous run. I'll try to configure pytest to ignore the directory.

latk: Oh no! That is a *very good* reason to keep this code here. I didn't understand the context from the “test is not running” comment, so expanding the comment to mention pytest collection + infinite loops would be perfect.

I'm also wondering if the combined Make/shell could be simplified a bit. I'd guess a pure-shell command might be clearer:

```
@ cursed_directory=...; if test -d "$$cursed_directory'; then rm -Rf "$$cursed_directory"; fi
```

<sub>`@` to avoid this lengthy command from being printed out by Make, `$$` for escaping, not sure about the correct test flag (`-d`, `-L`, `-e`?)</sub>

Could you also explain why an absolute path is needed here? Makefiles are effectively guaranteed to be only called from the correct directory – whoever uses `make -f` has issues. <sub>(Disclaimer: our `gcovr/tests/Makefile` uses this flag, but it seems unnecessary. I didn't even know we had that file!</sub>

chrta (approved):
lgtm

Makefile
Spacetown: I think this was because the wildcard function needs an absolute path.
In the test directory there is also a makefile which calls the clean target for all tests. Maybe calling this before pytest is better. Also adding the mentioned comment why it is called.

Spacetown: I've added pytest options to setup.cfg and know the directory is ignored.

PR #488, opened 2021-06-08
Bump version 5.0

Spacetown:
Bump the version and update the screenshots.

[no changelog]


Spacetown:
@latk It seems that the upload of the coverage data doesn't work anymore. In the first run of https://github.com/gcovr/gcovr/actions/runs/919474024 the upload from Ubuntu was working but Windows fails. After a rerun of all jobs both are failing with the message:
```
{'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
```
I've checked the token but this is correct.

Spacetown:
Replaced by #489 because of problems with name of branch in coverage report.

PR #491, opened 2021-06-10
Fix deploy workflow.

Spacetown:
- Fix problems mentioned in https://github.com/gcovr/gcovr/issues/424#issuecomment-858091994.
- Fix CMake varning because minimum required version was 2.6.4.

The deploy workflow is also executed on `push` and `pull_request` but the publish is skipped. There is also the dist folder saved as artifact.
Now we can be sure that the build will succeed.

[no changelog]

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -59,8 +59,14 @@ jobs:
     - name: Build
       run: |
         python setup.py sdist bdist_wheel
+    - name: Upload distribution
+      if: ${{ success() }}
+      uses: actions/upload-artifact@v2
+      with:
+        name: dist
+        path: dist/**
     - name: Publish
-      if: ${{ github.event_name == 'release' }}
+      if: ${{ success() && (github.event_name == 'release') }}
       env:
         TWINE_USERNAME: __token__
         TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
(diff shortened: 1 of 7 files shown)


Spacetown:
I hope everything is prepared for tomorrow. Keep the fingers crossed.

latk:
Is there a way to add `--no-verify-tags` to the `release_checklist` invocation if `github.event_name != 'release'`? Or should we skip that entire step for non-release runs? After all, the check is also performed by the normal test workflow.

Spacetown:
@latk Yes, there is a way and I started with this. I removed it  because the test only fails after numbing the new version. If the release is published it will not fail anymore.

chrta (approved):
lgtm (but i am no specialist here).

I would probably have installed ninja via "apt-get install ninja-build", if we do not want the bleeding edge (in ubuntu 20.04 is version 1.10.1).

latk:
> @latk Yes, there is a way and I started with this. I removed it because the test only fails after numbing the new version. If the release is published it will not fail anymore.

Ok, that makes sense. It seems acceptable to have a short period between bumping the version and releasing it where the workflow doesn't quite work. Not absolutely perfect since this period is critical, but this is a case of “don't let perfect be the enemy of good”.

latk (approved):
:rocket:  Ship it :)

PR #493, opened 2021-06-11
Clear environment variables used by GCC compiler

Spacetown:
Clean GCC environment variables as mentioned in https://github.com/gcovr/gcovr/issues/492#issuecomment-859564833.


gcovr/tests/test_gcovr.py
```diff
@@ -32,42 +32,60 @@
...
+    "C_INCLUDE_PATH",
+    "CPLUS_INCLUDE_PATH",
+    "OBJC_INCLUDE_PATH",
+    "CFLAGS",
+    "CXXFLAGS",
+    "LDFLAGS",
+    "LANG"
+]:
+    if var in env:
+        env.pop(var)
```
latk, line 51: Instead of *deleting* LANG, it could make sense to force it to a standard locale such as `LANG=C` or `LANG=C.UTF-8`. We already do that in the Dockerfile since `black` seemed to have problems otherwise, cf #487. Either way, LGTM.

latk (approved):

gcovr/tests/test_gcovr.py
Spacetown: You're right. Will change this before merge.

PR #494, opened 2021-06-11
Fix problems from deployment of release 5.0

Spacetown:
- Fix README.rst for PyPI upload.
- Rename license_file to license_files because of deprecation warning.
- Add twine to the requirements.txt to run check mentioned in https://packaging.python.org/guides/making-a-pypi-friendly-readme/. Our problems are not found by the check.
- Remove the deployment trigger for the release because we want to use annotated tags. The workflow is triggered if a tag is pushed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ setup(
     name="gcovr",
     version=version,
     long_description=long_description,
-    long_description_content_type="text/markdown",
+    long_description_content_type="text/x-rst",
     platforms=["any"],
     python_requires=">=3.6",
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -22,26 +22,44 @@
...
+    r"``\1``",
+    long_description,
+    flags=re.MULTILINE
+)
+
+setup(
+    name="gcovr",
+    version=version,
+    long_description=long_description,
+    long_description_content_type="text/markdown",
```
latk, line 52: Readme is RST not Markdown

```suggestion
    long_description_content_type="text/x-rst",
```

latk (approved):

PR #495, opened 2021-06-13
replace test XML diffing: yaxmldiff instead of pyutilib

latk:
Pyutilib is a wonderful collection of Python helpers developed by the same people who created gcovr. But as gcovr is converging towards more standard Python practices, we've only been using it for XML diffing.

It turns out that XML diffing is a very popular problem, but most libraries do something that's too clever for our tests or don't produce readable output.

The `lxml.doctestcompare` module is very good but annotates differences *inline*, e.g. produces this when comparing `<a/>` with `<b/>`:

    Expected:
      <a></a>

    Got:
      <b></b>

    Diff:
      <a (got: b)></a (got: b)>

In large documents, this can be hard to spot.

There are additional issues, such as special `<any/>` tags, `any=""` attributes, and `...` ellipsis that are useful for doctests, but not useful for comparing two actual XML documents.

Another tool is xmldiff which is also very good, except that it outputs diff as instructions that are more machine-readable than human-accessible. For the same problem, it will output:

    [rename, /a[1], b]

So because I'm dissatisfied with available tools, I've written Yet Another XML Differ (yaxmldiff). It uses lxml and tries to approximate a unified diff:

    - <a/>
    + <b/>

On larger problems, it will try to collapse irrelevant sub-trees and attributes. For example, here is a diff between the gcc-5 and gcc-8 reference files for the simple1 test:

```diff
  <coverage branch-rate="0.5" complexity="0.0"
-   branches-covered="4"
+   branches-covered="2"
-   branches-valid="8"
+   branches-valid="4"
-   line-rate="0.8"
+   line-rate="0.7777777777777778"
-   lines-covered="8"
+   lines-covered="7"
-   lines-valid="10"
+   lines-valid="9"
-   timestamp=""
+   timestamp="1601411485"
-   version=""
+   version="gcovr 5.0"
  >
    <sources>
    ...
    </sources>
    <packages>
      <package branch-rate="0.5" complexity="0.0" name=""
-       line-rate="0.8"
+       line-rate="0.7777777777777778"
      >
        <classes>
          <class branch-rate="0.5" complexity="0.0" filename="..." name="..."
-           line-rate="0.8"
+           line-rate="0.7777777777777778"
          >
            <methods/>
            <lines>
              <line .../>
              <line ...>
              ...
              </line>
              <line .../>
              <line .../>
              <line .../>
              <line branch="..." hits="1"
-               number="17"
+               number="14"
              />
              <line ...>
              ...
              </line>
              <line branch="..." hits="0"
-               number="22"
+               number="19"
             />
              <line .../>
-             <line ...>...</line>
            </lines>
          </class>
        </classes>
      </package>
    </packages>
  </coverage>
```

This tool did spot a problem in the existing reference files. In the GCC-5 reference for the "shadow" test, Pyutilib considered the following two attributes equivalent, likely because it tries to compare things that look like numbers as numbers:

    condition-coverage="50% (1/2)"
    condition-coverage="50% (2/4)"

The reference file was updated manually.

This also hints at the drawback of yaxmldiff: it treats everything as text, and cannot apply rounding like the Pyutilib functions.

Links:

* yaxmldiff on PyPI: <https://pypi.org/project/yaxmldiff/>
* yaxmldiff on GitHub: <https://github.com/latk/yaxmldiff.py>

This PR contains two commits: one for the actual change, one to re-format the test_gcovr.py file. The reformatting will be redundant once #493 is merged.


gcovr/tests/test_gcovr.py
```diff
@@ -300,23 +352,35 @@ def test_build(compiled, format, available_targets, generate_reference, update_r
...
+                        ),
+                    )
 
     diff_is_empty = len(whole_diff_output) == 0
     assert diff_is_empty, "Diff output:\n" + "".join(whole_diff_output)
 
     # some tests require additional cleanup after each test
-    if 'clean-each' in available_targets:
-        assert run(['make', 'clean-each'])
+    if "clean-each" in available_targets:
```
Spacetown, line 383:
```suggestion
    if "clean-each" in available_targets:  # pragma: no cover
```

gcovr/tests/test_gcovr.py
```diff
@@ -155,74 +171,102 @@ def pytest_generate_tests(metafunc):
...
+                    name, unresolved_prereqs
+                )
+            )
 
         # check that all available known formats are also listed in the "run" target
-        unreferenced_formats = set(KNOWN_FORMATS).intersection(targets).difference(target_run)
+        unreferenced_formats = (
+            set(KNOWN_FORMATS).intersection(targets).difference(target_run)
+        )
         if unreferenced_formats:
```
Spacetown, line 206:
```suggestion
        if unreferenced_formats:  # pragma: no cover
```

gcovr/tests/test_gcovr.py
```diff
@@ -155,74 +171,102 @@ def pytest_generate_tests(metafunc):
...
-                name, unknown_formats))
+            raise ValueError(
+                "{}/Makefile target 'run' references unknown format {}".format(
+                    name, unknown_formats
+                )
+            )
 
         # check that all "run" targets are actually available
         unresolved_prereqs = target_run.difference(targets)
         if unresolved_prereqs:
```
Spacetown, line 195:
```suggestion
        if unresolved_prereqs:  # pragma: no cover
```

gcovr/tests/test_gcovr.py
```diff
@@ -155,74 +171,102 @@ def pytest_generate_tests(metafunc):
...
         targets = parse_makefile_for_available_targets(
-            os.path.join(basedir, name, 'Makefile'))
+            os.path.join(basedir, name, "Makefile")
+        )
 
         # check that the "run" target lists no unknown formats
-        target_run = targets.get('run', set())
+        target_run = targets.get("run", set())
         unknown_formats = target_run.difference(KNOWN_FORMATS)
         if unknown_formats:
```
Spacetown, line 186:
```suggestion
        if unknown_formats:  # pragma: no cover
```

Spacetown (approved):
LGFM. The suggestions are because this part shall never be reached in the CI..

PR #502, opened 2021-07-21
Add error to STDOUT --fail-under-line or --fail-under-branch.

### LINKED ISSUE #98: Coverage Threshold parameter

Spacetown:
Add message suggested in #98.

Closes #98

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,7 @@ Improvements and new features:
 
 Documentation:
 
- - Add error to STDOUT --fail-under-line or --fail-under-branch. (:issue:`502`)
+ - Add error to STDERR for --fail-under-line or --fail-under-branch. (:issue:`502`)
 
 Internal changes:
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,8 @@ Improvements and new features:
 
 Documentation:
 
+ - Add error to STDOUT --fail-under-line or --fail-under-branch. (:issue:`502`)
+
 Internal changes:
 
  - Clean GCC environment variables in test suite. (:issue:`493`)
--- a/gcovr/__main__.py
+++ b/gcovr/__main__.py
@@ -52,7 +52,7 @@ from .writer.coveralls import print_coveralls_report
 #
 # Exits with status 2 if below threshold
 #
-def fail_under(covdata, threshold_line, threshold_branch):
+def fail_under(covdata, threshold_line, threshold_branch, logger):
     (lines_total, lines_covered, percent,
         branches_total, branches_covered,
         percent_branches) = get_global_stats(covdata)
@@ -61,10 +61,25 @@ def fail_under(covdata, threshold_line, threshold_branch):
         percent_branches = 100.0
 
     if percent < threshold_line and percent_branches < threshold_branch:
+        logger.error(
+            "Failed, not enough line coverage (got: {}, expected min: {})",
+            percent,
+            threshold_line
+        )
         sys.exit(6)
     if percent < threshold_line:
+        logger.error(
+            "Failed, not enough line coverage (got: {}, expected min: {})",
+            percent,
+            threshold_line
+        )
         sys.exit(2)
     if percent_branches < threshold_branch:
+        logger.error(
+            "Failed, not enough branch coverage (got: {}, expected min: {})",
+            percent,
+            threshold_branch
+        )
         sys.exit(4)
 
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.rst
```diff
@@ -12,6 +12,8 @@ Improvements and new features:
 
 Documentation:
 
+ - Add error to STDOUT --fail-under-line or --fail-under-branch. (:issue:`502`)
```
latk, line 15:
```suggestion
 - Add error to STDERR for --fail-under-line or --fail-under-branch. (:issue:`502`)
```

The error messages are actually written to STDERR, as it should be. If messages were written to STDOUT, this would also conflict with reports that are printed to STDOUT, e.g. `gcovr -x > cov.xml`.

gcovr/__main__.py
```diff
@@ -61,10 +61,25 @@ def fail_under(covdata, threshold_line, threshold_branch):
         percent_branches = 100.0
 
     if percent < threshold_line and percent_branches < threshold_branch:
+        logger.error(
+            "Failed, not enough line coverage (got: {}, expected min: {})",
+            percent,
+            threshold_line
+        )
```
latk, line 68: Both the line and branch coverage failed the threshold, but only line coverage is mentioned. Is that intentional?

I'd probably write this code like this, so that independent error messages are shown for each criterion:

```python
exit_code = 0
if percent < threshold_line:
    exit_code |= 2
    logger.error("failed minimum line coverage (got {}%, minimum {}%)", percent, threshold_line)
if percent_branches < threshold_branch:
    exit_code |= 4
    logger.error("failed minimum branch coverage (got {}%, minimum {}%)", percent_branches, threshold_branch)
if exit_code:
    sys.exit(exit_code)
```

gcovr/__main__.py
```diff
@@ -61,10 +61,25 @@ def fail_under(covdata, threshold_line, threshold_branch):
...
         sys.exit(2)
     if percent_branches < threshold_branch:
+        logger.error(
+            "Failed, not enough branch coverage (got: {}, expected min: {})",
+            percent,
+            threshold_branch
```
latk, line 81: Should be `percent_branches`, not `branch`.

Could we add some simple tests that check that the messages contain the correct values? The makefile-based tests can't do this, but capturing STDERR in `test_args.py` should work (by providing coverage data via JSON in a tempfile).

latk (commented):
good feature, some comments.

gcovr/__main__.py
Spacetown: You'r right but the bitwise or for the exitcode is confusing for me. If you search for the sys.ext(6) you'll never find it. What about doing something like:
```python
line_nok = False
branch_nok = False
if percent < threshold_line:
    line_nok = True
    logger.error("failed minimum line coverage (got {}%, minimum {}%)", percent, threshold_line)
if percent_branches < threshold_branch:
    branch_nok = True
    logger.error("failed minimum branch coverage (got {}%, minimum {}%)", percent_branches, threshold_branch)
if line_nok and branch_nok:
    sys.exit(6)
if line_nok:
    sys.exit(2)
if branch_nok:
    sys.exit(4)
```

gcovr/__main__.py
Spacetown: I'll add a test.

gcovr/tests/simple1/Makefile
```diff
@@ -18,10 +18,16 @@ run: txt xml html sonarqube json json_summary coveralls
...
+	$(GCOVR) --fail-under-line 80.1 2>fail_under.stderr; test $$? -eq 2 || ( cat fail_under.stderr & exit 1 )
+	grep -F "failed minimum line coverage" fail_under.stderr
+	$(GCOVR) --fail-under-branch 50.1 2>fail_under.stderr; test $$? -eq 4 || ( cat fail_under.stderr & exit 1 )
+	grep -F "failed minimum branch coverage" fail_under.stderr
+	$(GCOVR) --fail-under-line 80.1 --fail-under-branch 50.1 2>fail_under.stderr; test $$? -eq 6 || ( cat fail_under.stderr & exit 1 )
+	grep -F "failed minimum line coverage" fail_under.stderr
+	grep -F "failed minimum branch coverage" fail_under.stderr
+	$(GCOVR) --fail-under-line $(FAIL_UNDER_LINE) --fail-under-branch 49.9 2>fail_under.stderr; test $$? -eq 0 || ( cat fail_under.stderr & exit 1 )
+	grep -F "failed minimum line coverage" fail_under.stderr; test $$? -eq 1
+	grep -F "failed minimum branch coverage" fail_under.stderr; test $$? -eq 1
```
latk, line 30: I think this can be simplified to `! grep ...`?

latk (approved):

gcovr/tests/simple1/Makefile
Spacetown: But then everithing not 0 will success, from `grep` manpage:
```
EXIT STATUS
     The grep utility exits with one of the following values:

     0     One or more lines were selected.
     1     No lines were selected.
     >1    An error occurred.
```

PR #503, opened 2021-07-21
Generate also a Report of Excluded Coverage

### LINKED ISSUE #409: Generate also a Report of Excluded Coverage
We support to exclude coverage from some regions via the GCOVR_EXCL_... markers. In my opinion these should only be used with care and if really necessary.  To judge the places where the markers are used it would be nice to include them in some way to the overall reports. I could imagine some ways (for e.g. the HTML report):

a) We have another column which shows how many lines are skipped. Then we can quickly judge how many lines are not considered?
b) We add an additional page which lists all places where we exclude coverage?
c) We mark the excluded regions with a special color in the detailed coverage report, currently they are just not marked at all.

Spacetown:
Add excluded flag to JSON output and mark lines in HTML report.

Shall the excluded lines also be listed for each file in the overview? 

Closes #409


CHANGELOG.rst
```diff
@@ -19,11 +19,10 @@ Improvements and new features:
 - Fix "root" path in JSON summary report. (:issue:`548`)
 - Solve problems with file name limitations and document the used options of ``gcov``. (:issue:`528`)
 - Add :option:`--no-markers` to ignore exclusion markers in code. (:issue:`361`)
+- Generate also a Report of Excluded Coverage. (:issue:`503`)
 
 Documentation:
 
- - Add error to STDERR for --fail-under-line or --fail-under-branch. (:issue:`502`)
-
```
latk, line 26: Removing this line seems to be a merge conflict artifact.

gcovr/coverage.py
```diff
@@ -140,7 +148,7 @@ def update(self, other):
         assert self.lineno == other.lineno
         self.count += other.count
         self.noncode &= other.noncode
-
+        self.excluded &= other.excluded
```
latk, line 151: Is it correct to exclude a line only if all coverage data sets being merged exclude this line? Or should a line be excluded if any data set excludes it?

Conflicting exclusions could appear e.g. if we're merging two JSON files, one of which was created with `--no-markers` or if they had different `--exclude-lines-by-pattern` settings.

I think that if one data set has the `excluded` flag set, then this means that data on that line should not be considered. I would write this as `self.excluded |= other.excluded`, but am open to different arguments. For example, using an `&` would be consistent with how file-level filtering works.

This is different from the `noncode` flag which we use to mark lines that gcov considers to be uncovered – if another coverage data set does include data for that line, the noncode flag should disappear.

If this line is changed, then the code for merging file coverage below must not set `excluded=True`.

latk (approved):
Looks good, I just have a question about merging coverage data (see below).

I wouldn't show statistics on excluded coverage in the overviews (e.g. HTML report, text report) since it might overload the report with information. Of course, this is something that can be added in the future if there is user interest. And maybe, summarizing coverage as covered/excluded/total might not be so unreadable after all.

gcovr/coverage.py
Spacetown: A good question. I think also that excluding the line one time will exclude it in the merged report is better. Else we would get lower coverage values. I'll change this.

latk:
Looks like the reference files need updating for the documentation examples, but then this is DONE! Thanks for working on this feature! (And sorry for my extremely late review…)

latk:
I'm currently studying the source-level exclusion logic in order to figure out whether it can be disentangled from the gcov-file parser … and noticed that this PR doesn't quite work:

* this PR adds information about excluded lines to some reports
* but nowhere is the `excluded=True` flag set.

There is no code path that could change this flag, unless one reads a manually prepared JSON file via the `--add-tracefile` option. Also, all our example reference files only include `"gcovr/excluded": false` entries, not even one `"gcovr/excluded": true`.

I missed this entirely during my review. But maybe, tracking exclusion status properly in the data model would actually be useful – in particular, it would be valuable to distinguish `noncode` from `excluded`. I'll have to think about this some more.

Spacetown:
You're right. The only place where a `LineCoverage` is added without preserving the excluded flag is:
https://github.com/gcovr/gcovr/blob/40496751280e6bf305ee242485372a749988d0cb/gcovr/gcov_parser.py#L450-L453
I've prepared a fix but the background color in the blue theme (#53BFFD) can't be distinguished from the covered color (#66B4FF).

Issue #271, opened 2018-07-12
Incorrect paths for header files


ZedThree:
Header files don't seem to be dealt with correctly:

```bash
$ gcovr -j8 -r . --html-details -o gcovr-report/coverage.html -dTraceback (most recent call last):
  File "/home/p/.local/bin/gcovr", line 11, in <module>
    sys.exit(main())
  File "/home/p/.local/lib/python3.6/site-packages/gcovr/__main__.py", line 588, in main
    print_html_report(covdata, options)
  File "/home/p/.local/lib/python3.6/site-packages/gcovr/html_generator.py", line 275, in print_html_report
    errors='replace')
FileNotFoundError: [Errno 2] No such file or directory: 'src/array.hxx'
```

While `--html-details` throws an error, just running `gcovr` counts each header separately for each directory containing a source file that includes it, so I get stuff like:

```
------------------ ...
File               ...
------------------ ...
src/dir1/array.hxx ...
src/dir2/array.hxx ...
...
```

If I keep the gcov files, I can see they have the correct names, i.e.:

```
./^#include#array.hxx.gcov
./src/dir1/^#^#include#array.hxx.gcov
./src/dir2/^#^#include#array.hxx.gcov
```

This is with gcovr version 4.1 and gcc 8.1

latk:
Thank you for reporting this problem, you're right that gcovr
behaviour seems to be completely bonkers in this case. I'll try to
take a look during the weekend.

It would help immensely if you could add the following:

1. the first few lines of all involved gcov files (they contain the
actual source file name; the gcov file name is not reliable)

2. a short hello-world style example to reproduce the issue so that I
can experiment on this myself. I.e. the necessary file contents,
directory layouts, and compiler flags.

latk:
This issue has not seen any progress in a while, so I am closing it for now. If the problem persists, please add further details so it can be re-opened. If someone else is experiencing a similar issue, please create a new issue.

ZedThree:
Sorry for the late reply, I went on holiday before I got round to updating this! Apologies for the following information dump...

Here is a zipfile containing a complete toy problem: 
[gcovr-271.zip](https://github.com/gcovr/gcovr/files/2222769/gcovr-271.zip)

Directory layout is as follows:

```
├── make.config
├── makefile
├── include
│   ├── makefile
│   ├── sys
│   │   └── foo.hxx
│   └── toy.hxx
├── src
│   ├── makefile
│   ├── sys
[...]

PR #505, opened 2021-07-26
Update html.py

zhajkun:
fix issue #271, make gcovr to search for the files in all directory.

### DIFF
--- a/gcovr/writer/html.py
+++ b/gcovr/writer/html.py
@@ -374,9 +374,21 @@ def print_html_report(covdata, output_file, options):
     # Generate an HTML file for every source file
     #
     error_occurred = False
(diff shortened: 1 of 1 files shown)


latk:
Hi zhajkun! Thank you for this suggestion. It seems your code is ensuring that a source file with a matching name exists before adding it to the HTML report, thus maybe preventing some errors.

However, I have severe doubts whether we can include this in gcovr:

* it assumes that the project is in a git repository
* it only looks at the basenames of files
* it scans through the entire project for each input file, which is going to be extremely slow on large projects
* when files are skipped it does so silently, thus possibly hiding errors and reporting wrong coverage statistics
* I don't think it actually fixes #271

So I am closing this pull request for now, but I am grateful for every attempt at fixing long-standing issues like #271 :)

Issue #511, opened 2021-09-08
Possible bug in function coverage: assertion fails in some cases


erikbs:
**Describe the bug**
The assertion on line 81 in coverage.py (i.e. `assert self.lineno == other.lineno`), introduced in commit 5b9cdec (30 July 2021), fails in some cases. We had issues with this in multiple projects before switching to release v5.0, but I only have details from one of them. In that project, the assertion would fail in the same file (a C++ header unique to this project), at the same line (discovered by adding debug printouts). The affected line was reached about 15 times, and the assertion failed 1-4 times per run (I assume that the thread-pool is the reason). When it failed, `self.lineno` and `other.lineno` were either 31 or 32, where line 32 contains a method definition and 31 is the empty line above it (line 30 is a destructor initialised to `default` and line 33 is another method definition). If the assertion failed multiple times in a run, `self.lineno` could be the same in all cases or it could be 31 in some cases and 32 in the rest.

The tracebacks (see below for examples) revealed that the assertion was reached from various different code paths. Unfortunately I was not able to strip down the project to create a minimal example (simple, irrelevant changes, such as deleting an unrelated source file or deleting the empty line, could make the problem go away), and I am not allowed to share the source code. I understand that with this little information it might be hard to debug the issue, but I am submitting a report anyway in case it contains enough information to put someone on the right track.

**To Reproduce**
1. Deleted the build directory
2. Built the project for coverage and ran tests that produced coverage data as usual
3. Ran a script that calls `gcovr` through Python, using the following command: `PYTHONPATH=${GCOVR_SRC} python3 -m gcovr --html --html-details -o ${BINDIR}/coverage/coverage.html --gcov-executable=${GCOV} -j 8 -s --root ${SOURCE_DIR} -- ${BINDIR}` (`GCOVR_SRC` is the root of the gcovr repo and the other environment variables are as one would expect).

**Expected behavior**
Preferably the assertion should not fail. If it is not possible to find the cause, I suggest that the report be produced without unhandled exceptions, e.g. by emitting a warning instead.

**Desktop (please complete the following information):**
 - OS: CentOS
 - GCC version 9, 10 and 11 (same results)
 - GCOVR version: master branch (per 30 July 2021)
 - Python: 3.6

**Additional context**
Traceback examples include (file paths anonymised):

```
Traceback (most recent call last):
  File "(...)/gcovr-src/gcovr/workers.py", line 76, in worker
    work(*args, **kwargs)
[...]

PR #512, opened 2021-09-10
new parser for the .gcov format

latk:
Summary: This commit introduces a completely rewritten parser for gcov's human-readable report format. The implementation pays closer attention to the upstream file format, e.g. with respect to the exact number formatting. Compared to the previous parser, the data flow has been greatly clarified, making this parser more obviously correct. The new parser will make it easier to react to future changes in the format, and to make use of improvements from newer GCC versions – such as the `Working directory` metadata.

## Background

The old parser had some problems that could be addressed by this rewrite.

-   The old parser was very complex due to mixing low level parsing and high-level decisions about coverage exclusion.

    Fixed in the new parser: now, parsing is split into multiple phases that clearly separate concerns. The low-level parsing phase produces a data model that is safe to use for the other phases.

-   The old parser used an object-oriented design where parsing state was stored in instance fields. This created a huge "state space" that was difficult to track, potentially leading to bugs. For example, it was not immediately clear in the investigation of <https://github.com/gcovr/gcovr/issues/511> whether the problem was just that the parser forgot to update its line number correctly.

    Fixed in the new parser: using a more functional/procedural design, the data flows are very clear. State is tracked explicitly. By separating parsing phases, the state space is much smaller.

-   To expand on the previous point: The old parser essentially consisted of multiple interleaved state machines. There was one state machine for processing coverage data, and an interleaved state machine for tracking active coverage exclusion markers. This interleaving made it difficult to understand whether the state transitions were correct.

    Fixed in the new parser: coverage exclusion patterns are collected in a separate phase, before the actual coverage data is processed.

-   The old parser made use of very fragile parsing strategies, such as using `str.split()` excessively. This gave rise to fragile assumptions about the exact format. For example, the IndexError in <https://github.com/gcovr/gcovr/issues/226> was an example of wrong assumptions.

    The new parser uses regular expressions to parse tag lines, and only uses `str.split()` for the more structured source code lines. This is more self-documenting. Treatment of numerical values was aligned with the routines in the gcov source code. Should the format deviate in the future, the regexes will fail to match, making it possible to detect and fix the errors. (Until then, `--gcov-ignore-parse-errors` can be used).

## Design of the new parser

The new parser is more complex in the sense that there is a lot more code. But there is a clearer separation of concerns, and the parser was closely informed by the gcov documentation and source code. As a result, I am confident that it handles far more edge cases correctly, in particular relating to the handling of numbers/percentages.

There are three items for external use:

**`parse_metadata(lines)`** creates a dict of values from the metadata lines. The old parser handled the very first line of the report separately to extract the filename. The new parser uses the same more robust parsing code for this metadata.

**`ParserFlags`** is a flag-enum that describes various boolean features. A single object with flags seems simpler to handle than multiple variables like `exclude_throw_branches`.

**`parse_coverage(lines, ...)`** is the main function for parsing the coverage. It performs multiple phases:

-   Each input line is parsed/tokenized into an internal data model. The data model is strongly typed. The various classes like `_SourceLine` are implemented as NamedTuples, which is both very convenient and very memory-efficient.

    Relevant items: `_parse_line()`, `_parse_tag_line()`, data model

-   Exclusion markers are extracted from source code lines and arranged into a data structure for later lookup.

    Relevant items: `_find_excluded_ranges()`, `_make_is_in_any_range()`

-   Parsed lines are processed to populate a `FileCoverage` model. At this stage, exclusions are applied. The state space is very small, with only four variables that have to be tracked.

    Relevant items: `_ParserState`, `_gather_coverage_from_line()`, `_line_noncode_and_count()`, `_function_can_be_excluded()`, `_branch_can_be_excluded()`, `_add_coverage_for_function()`

-   Warnings are reported, and any potential errors re-raised. This is equivalent to the previous parser.

    Relevant items: `_report_lines_with_errors()`

## Impact on tests

The new parser is almost completely bug-compatible with the old parser. This is e.g. visible in the potentially unintuitive handling of the `noncode` status. The new separation between low-level parsing and high-level decisions makes it more clear what is actually going on.

There was a significant change in the **Nautilus parser test**. The input file contains the following pattern:

    ------------------
        #####:   52:foo() ? bar():

Previously, this line 52 was not reported as uncovered. I consider that to be an error, and have updated the expected output correspondingly. This could indicate that the new parser is in fact more robust than the old parser when it comes to template specialization sections.

In the **excl-branch test**, gcovr will encounter gcov input as the following when using GCC-8 or later:

        #####:    9:    virtual ~Bar()
        #####:   10:    {} // ...
    ------------------
    Bar::~Bar():
    function Bar::~Bar() called 0 returned 0% blocks executed 0%
        #####:    9:    virtual ~Bar()
        #####:   10:    {} // ...
    call    0 never executed
    call    1 never executed
    ------------------
    Bar::~Bar():
    function Bar::~Bar() called 0 returned 0% blocks executed 0%
        #####:    9:    virtual ~Bar()
        #####:   10:    {} // ...
    ------------------

The old parser associated the `function` annotations with line 11. This was clearly incorrect. The test reference was updated to associate the destructor with line 9.

Other than that, the tests were only updated to account for the different parser APIs.

Internally, the new parser uses a lot of doctests.

## Future directions

The new parser extracts *all* available data, only to throw it away. It might now become feasible to make use of more of this data. In particular:

-   handling template specialization sections properly
-   collecting block-level coverage data
-   using the `working directory` metadata field

## Conflicts with other development efforts

-   <https://github.com/gcovr/gcovr/pull/503> report of excluded coverage

    Makes a small patch to the parser. The same effect can be achieved by adding a few lines in `_gather_coverage_from_line()`.

-   <https://github.com/gcovr/gcovr/pull/484> tests with clang-10

    Touches neighboring lines. Will be reported as a merge conflict by Git, but there's no semantic conflict.

-   <https://github.com/gcovr/gcovr/pull/474> abstract interface for reader/writer

    Small change in the parser code regarding `sys.exit(1)` (new parser: `raise SystemExit(1)`). It's worth noting that this is effectively unreachable code. Lines will only be reported if there was an exception, and if there was an exception it will be re-thrown.

-   <https://github.com/gcovr/gcovr/pull/361> --no-markers to ignore exclusion markers

    Touches the exclusion handling code. This is of course totally changed by the new parser. But the new parser would make it even easier to implement that functionality.

-   <https://github.com/gcovr/gcovr/pull/350> decision coverage

    Adds significant new parsing code, but most of it is outside of the gcov-parser. These changes could be ported with moderate effort to the new parser.


gcovr/gcov_parser.py
```diff
@@ -0,0 +1,994 @@
...
+    >>> parse_metadata('''
+    ...   -: 0:Foo:bar
+    ...   -: 0:Key:123
+    ... '''.splitlines())
+    {'foo': 'bar', 'key': '123'}
+    """
+    collected = dict()
+    for line in lines:
+        # IO.readlines() leaves trailing newlines...
+        line = line.rstrip("\n")
```
Spacetown, line 204: Could the rstrip in each for loop be moved to the place where the file is read?

gcovr/gcov_parser.py
```diff
@@ -0,0 +1,994 @@
...
+    segments = line.split(":", 2)
+    if len(segments) == 3:
+        count_str, lineno, source_code = segments
+        count_str = count_str.strip()
+        lineno = lineno.strip()
+
+        # METADATA (key, value)
+        if count_str == "-" and lineno == "0":
+            key, value = source_code.split(":", 1)
+            return _MetadataLine(key.lower(), value)
```
Spacetown, line 668: Why do you change the key to lowercase? I prefer keeping the original case.

gcovr/gcov_parser.py
```diff
@@ -0,0 +1,994 @@
...
+            # STOP flags remove a marker from the exclusion stack
+            for header, flag in _EXCLUDE_LINE_PATTERN.findall(code):
+
+                if flag == "START":
+                    exclusion_stack.append((header, lineno))
+                    continue
+
+                if flag == "STOP":
+                    if not exclusion_stack:
+                        logger.warn(
```
Spacetown, line 842: Can a test be added to cover this?

gcovr/gcov_parser.py
```diff
@@ -0,0 +1,994 @@
...
+                            "when processing {fname}",
+                            header=header,
+                            lineno=lineno,
+                            fname=filename,
+                        )
+                        continue
+
+                    start_header, start_lineno = exclusion_stack.pop()
+                    if header != start_header:
+                        logger.warn(
```
Spacetown, line 855: Dito

gcovr/gcov_parser.py
```diff
@@ -0,0 +1,994 @@
...
+                        )
+
+                    exclude_line_ranges.append((start_lineno, lineno))
+
+        if exclude_lines_by_pattern_regex:
+            if exclude_lines_by_pattern_regex.match(code):
+                exclude_line_ranges.append((lineno, lineno + 1))
+
+    for header, lineno in exclusion_stack:
+        logger.warn(
```
Spacetown, line 873: Dito.

Spacetown (commented):
Can you apply black to gcov.py and gcov_parser.py?

gcovr/gcov.py
```diff
@@ -99,57 +99,61 @@ def is_non_code(code):
...
+    if options.gcov_ignore_parse_errors:
+        parser_flags |= ParserFlags.IGNORE_PARSE_ERRORS
+    if options.exclude_function_lines:
+        parser_flags |= ParserFlags.EXCLUDE_FUNCTION_LINES
+    if options.exclude_internal_functions:
+        parser_flags |= ParserFlags.EXCLUDE_INTERNAL_FUNCTIONS
+    if options.exclude_unreachable_branches:
+        parser_flags |= ParserFlags.EXCLUDE_UNREACHABLE_BRANCHES
+    if options.exclude_throw_branches:
+        parser_flags |= ParserFlags.EXCLUDE_THROW_BRANCHES
```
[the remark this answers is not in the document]
latk: That sounds doable, but I'd like a very clear dependency graph. Right now, the dependencies related to the parser are:

    gcov -> gcov_parser -> {util, coverage}

To integrate these flags with the CLI/configuration, we'd need a `gcov_parser -> configuration` dependency which I'd really not like (the parser is a pure, isolated bit of code that doesn't even do I/O), or a `configuration -> gcov_parser` dependency (doable, but not a huge fan: I perceive the parser flags as an implementation detail that the config doesn't have to know about). Or the parser flags would have to be moved to third module, which would make the parser less self-contained.

Right now, the gcov module translates between the Options and the ParserFlags, which I like. Even if it's a couple of lines of boilerplate.

gcovr/gcov_parser.py
```diff
@@ -0,0 +1,994 @@
...
+Handle parsing of the textual ``.gcov`` file format.
+
+Other modules should only use the following items:
+`parse_metadata()`, `parse_coverage()`, `ParserFlags`.
+
+The behavior of this parser was informed by the following sources:
+
+* the old GcovParser class
+* the *Invoking Gcov* section in the GCC manual (version 11)
+* the ``gcov.c`` source code in GCC (especially for understanding the exact number format)
```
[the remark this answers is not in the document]
latk: Done, I extended the comment with links to the various sources.

gcovr/gcov_parser.py
latk: Great point! Now, the gcov module does `INPUT.read().splitlines()` instead of `INPUT.readlines()` which takes care of the problem at the source.

gcovr/gcov_parser.py
latk: I did that because the upstream gcov format is subjectively “wrong”, but you're right that keeping this consistent is better. Done!

gcovr/gcov_parser.py
latk: I added tests for 100% coverage of the gcov_parser module. It's a bit awkward to test the warnings because they have no consistent format, so I added an `_ExclusionRangeWarnings` class that encapsulates the exact messages.

latk:
The PR now has more tests, and the relevant files were re-formatted with black.

This PR closes #511, but I decided not to mention that issue in the changelog since it relates to a bug in an unreleased state. So this PR is half quality uplift, half fixups after #362 (function coverage metric).

gcovr/gcov.py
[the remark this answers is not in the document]
Spacetown: Ok, was only an idea.

Spacetown (approved):
LGFM

PR #514, opened 2021-09-16
Add option to run all compilers

Spacetown:
Also reduce the test matrix for GitHub actions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,9 @@ set_sensible_default = $(if $(filter undefined default,$(origin $(1))),$(2),$(va
 PYTHON := $(call set_sensible_default,PYTHON,python3)
 
 override AVAILABLE_CC := gcc-5 gcc-6 gcc-8 clang-10
+ifeq ($(MAKECMDGOALS),docker-qa)
+override AVAILABLE_CC += all
+endif
 
 # Setting CXX and GCOV depending on CC. Only CC has to be set to a specific version.
 # If using GitHub actions on Windows, gcc-8 is set but gcc is used, so we override it.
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -90,7 +95,7 @@ test: export CXXFLAGS := --this_flag_does_not_exist # Env removed in text_gcovr.
 test: export GCOV := $(GCOV)
 
 test:
-	cd gcovr/tests && make clean
+	cd gcovr/tests && make clean > /dev/null
```
latk, line 98: Or `make --quiet`? But the /dev/null Unix-ism seems to work on the Windows-based tests as well, so that seems fine as well.

Makefile
```diff
@@ -99,10 +104,16 @@ doc:
 docker-qa: export TEST_OPTS := $(TEST_OPTS)
 docker-qa: export GCOVR_ISOLATED_TEST := zkQEVaBpXF1i
 
+ifeq ($(CC),all)
+docker-qa:
+	$(foreach cc,$(filter-out all,$(AVAILABLE_CC)),make $@ CC=$(cc); )\
```
latk, line 109: Should probably use the `$(MAKE)` variable when calling itself recursively?

At this point I'm also fairly sure that we're stretching Make beyond what it can reasonably do. We should probably consider porting parts of the Makefile to a Python script, but maybe keep Make as a convenient frontend. Thoughts on this?

latk (approved):
This is a good idea!

Makefile
```diff
@@ -12,17 +12,22 @@ set_sensible_default = $(if $(filter undefined default,$(origin $(1))),$(2),$(va
 PYTHON := $(call set_sensible_default,PYTHON,python3)
 
 override AVAILABLE_CC := gcc-5 gcc-6 gcc-8 clang-10
+ifeq ($(MAKECMDGOALS),docker-qa)
```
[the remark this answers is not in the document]
Spacetown: I only use docker. 🤣 
Of course I can add the option for all targets.

Makefile
Spacetown: There is no option in make to suppress the output of the targets.

latk: `make -s` definitely works to suppress the usual output, and is even Posix-compliant. The `--quiet` or `--silent` options are GNU extensions to Make.

Makefile
Spacetown: Of course is`$(MAKE)` correct.
I'll adapt the make to be able to use all for every target.
I think we should change parts of the make to use tox or nox.

Makefile
Spacetown: I overlooked the `--quit` in the help output. I'll change this.

Makefile
```diff
@@ -54,6 +60,34 @@ help:
...
+test:
+	$(RUN_ALL_COMPILERS)
+
+doc:
+	$(RUN_ALL_COMPILERS)
+
+docker-qa:
+	$(RUN_ALL_COMPILERS)
+
+else
```
latk, line 89: Is it really meaningful to run targets other than qa/docker-qa for each compiler version? Extracting the `RUN_ALL_COMPILERS` pattern seems sensible, but I'd only apply it to those two targets.

Spacetown: Done.

PR #515, opened 2021-09-16
Html function outline

### LINKED ISSUE #68: feature: add side frame which shows functions outline for html output
when you click a foo.cpp to view its detailed coverage, you should keep scrolling down to the specified function you want to see.
just like that we can navigate to the specified function easily with a outline of functions in IDE like eclipse.
is it a reasonable feature?

Spacetown:
Add dropdown to jump to function in details view.

This PR also fixes the background of the function coverage in the file. The color was taken from the overall function coverage instead of the current file. See in the test `linked` where the coverage inside `subdir/B/main.cpp` was 100% but had the class `coverage-low` which is red.

Closes #68


gcovr/writer/html.py
```diff
@@ -405,6 +404,14 @@ def print_html_report(covdata, output_file, options):
 
         data['filename'] = cdata_fname[f]
 
+        # Only use demangled names (containing a brace)
+        data['function_names'] = sorted(
+            [
+                (f, cdata.functions[f].lineno)
+                for f in cdata.functions.keys() if ("(" in f) or (f == "main")
```
latk, line 411: I don't think demangled names have to contain a paren `()`. In particular, plain C function names will not contain one (since they aren't mangled to start with). So I would remove this filtering.

gcovr/templates/source_page.html
```diff
@@ -16,6 +16,31 @@
...
+        <script>
+          function gotoFunctionName () {
+            var menu = document.gotoFunctionName.selection;
+            if (menu.options[menu.selectedIndex].value != "") {
+              document.location.href = menu.options[menu.selectedIndex].value;
+              menu.selectedIndex = 0;
+            }
+          }
+          document.gotoFunctionName.selection.addEventListener('change', gotoFunctionName);
+        </script>
```
latk, line 40: We can't always use inline scripts, just like we can't always use inline CSS: the reports might be served from a HTTP server with an unsuitable Content Security Policy (CSP).

But I don't think we need any JavaScript – a list of links should be sufficient. The per-line branch coverage data uses a `<details>` element to create a popup without JS, maybe we can repurpose that here to save space.

So I think we could maybe replace this navigation tool by something like:

```html
<details>
  <summary>go to function</summary>
  <ul>
  {% for function_name in function_names %}
    <li><a href="#{{ function_name.1 }}">{{ function_name.0 }}</a></li>
  {% endfor %}
  </ul>
</details>
```

latk (commented):
Cool feature, but see comments.

Also, screenshots would really help to illustrate the feature :)

gcovr/templates/source_page.html
Spacetown: I'll check this.

gcovr/writer/html.py
Spacetown: Ok, now I see that in the test there is nearly all C++. I'll take a look into this. I added the filter because in clang we don't have the option `--demangle-names` and a name like e.g. `__Zx6foo5` isn't usable.

latk: I'd rather show a mangled name than accidentally hide valuable information. Separately from this PR, we can figure out a way to ensure as much demangling as possible. For example, using `_cxa_demangle()` function from libc++ if available, or piping through `c++filt` from GNU binutils or llvm-cxxfilt.

gcovr/templates/source_page.html
Spacetown: Done.

gcovr/templates/source_page.html
```diff
@@ -49,6 +49,17 @@
 {% endblock %}
 
 {% block content %}
+{% if function_names %}
+<details class="listOfFunctions">
+  <summary>List of functions</summary>
+  <ul>
+    {% for function_name in function_names %}
+    <li><a href="#{{ function_name.1 }}">{{ function_name.0 }}, line {{ function_name.1 }}</a></li>
```
latk, line 57: The link doesn't seem to work, it seems the line anchors are prefixed with a lowercase-L.

```suggestion
    <li><a href="#l{{ function_name.1 }}">{{ function_name.0 }}, line {{ function_name.1 }}</a></li>
```

gcovr/templates/style.details.css
```diff
@@ -82,6 +82,12 @@ pre
     margin-bottom: 0;
 }
 
+.listOfFunctions > ul > li > a
+{
+    text-decoration: none;
+    color: inherit;
```
latk, line 88: When removing the underline and characteristic blue text, the links need some other affordance to show that they are interactive elements. For example, the following adds a background colour on hover.

```css
.listOfFunctions a {
    text-decoration: none;
    color: inherit;
    transition: 0.3s;
    padding: 0 4px;
}

.listOfFunctions a:hover, .listOfFunctions a:focus {
    background: lightsteelblue;
}
```

Spacetown: I was also thinking about this. A background is even better than the underline. Should we add a icon via CSS after the link?

latk: what kind of icon would you suggest? Perhaps turning that list into a table that also shows the number of calls, and covered/uncovered status through the background color?

Spacetown: Something like this:
Bildschirmfoto 2021-09-19 um 12 11 40

Or a unicode character like her:
Bildschirmfoto 2021-09-19 um 12 13 09

latk: 🤔 any affordance would do, as long as it's clear that the function name is clickable. It makes sense to remove `text-decoration: underline` since it tends to conflict with underscores in function names, but one or more hints like blue text color, interactive background color or outline, or symbols are necessary. Of the suggested symbols I only dislike the “box with arrow pointing outward” since it suggests an external site. The “→” would optically clash with the bullet point from the list, but that can be fixed with a “list-style: none”.

Spacetown: I added a table with function level information. The rows have the same behavior as the toplevel filelist (alternating row colors and chanhing color on hover). What do you think about this?

Bildschirmfoto 2021-09-19 um 23 11 37

latk: Looks fine! There are some details I might change in a future PR, but I don't want to “bikeshed” the design so much that we never get around to releasing this improvement :)

It strikes me that this table would benefit greatly from linking call locations (perhaps as part of a separate PR) but I'm not sure how this can be done correctly as we can't know whether a function has internal linkage – there might be different functions with the same name in different compilation units. Call locations aren't yet represented in our coverage data model, but my new parser would make it easy to get this info.

Spacetown: Than  I'll update the test data.

latk (approved):
I didn't manage to fully review these changes, but I don't think the changes to the code will break anything. So I'd be happy to see this get merged.

PR #516, opened 2021-09-22
Add nox to run test and environment

Spacetown:
As suggested in #514 this PR replaces the makefile to run the quality test with nox.
[Nox](https://nox.thea.codes/en/stable/index.html) is configured with a `noxfile.py` where the tests are described. The needed modules for the test are installed in a test specific virtual environment. Because of this the requirement.txt files are removed, the user only need to install nox.

Also the mechanism if only a gcc instead of gcc-x is available (e.g. for GitHub actions running on Windows) is changed. Instead of copying the executable an additional environment variable `CC_REFERENCE` is introduced which is used in `test_gcovr.py` if it is defined.

The building of the wheel is also moved from the pipeline to the `noxfile.py` and the pipeline now uses nox for running `setup.py`. 

The `noxfile.py` also contains a list of black conform files. This files are checked with different options to abort if there are formatting errors. If a file is formated with black it should be added to the list. After all files are black conform the call should be changed.


noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+        "sphinxcontrib-autoprogram==0.1.5 ; python_version=='3.6'",
+        "sphinxcontrib-autoprogram>=0.1.5 ; python_version>='3.7'",
+    )
+    session.install("-e", ".")
+    session.run("bash", "-c", "cd doc && make html O=-W", external=True)
+
+
+@nox.session
+def tests(session: "nox.session") -> None:
+    session_id = "tests_version({})".format(GCC_VERSION2USE)
```
latk, line 99: All uses of `"".format()` can be replaced by `f""`-strings.

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+    session.install("-e", ".")
+    set_environment(session, version)
+    session.log("Print tool versions")
+    session.run("python", "--version")
+    session.run(session.env["CC"], "--version", external=True)
+    session.run(session.env["CXX"], "--version", external=True)
+    session.run(
+        session.env["GCOV"].replace(" gcov", "")
+        if "llvm-cov" in session.env["GCOV"]
+        else session.env["GCOV"],
```
latk, line 136: What does this do? I don't understand. Is this trying to un-shellquote the command? In `gcov.py`, we use `shlex.split()` for that purpose. Alternatively, we could yolo it and `run("sh", "-c", f"{session.env['GCOV']} --version")`

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+    session.run("twine", "check", "dist/*", external=True)
+
+
+@nox.session
+def upload_wheel(session: "nox.session") -> None:
+    session.install("twine")
+    session.run("twine", "upload", "dist/*", external=True)
+
+
+def docker_container_id(version: str) -> None:
```
latk, line 163:
```suggestion
def docker_container_id(version: str) -> str:
```

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+@nox.session
+def tests_all_versions(session: "nox.session") -> None:
+    for version in GCC_VERSIONS:
+        session_id = "tests_version({})".format(version)
+        session.log("Notify session {}".format(session_id))
+        session.notify(session_id)
+
+
+@nox.session
+@nox.parametrize("version", [nox.param(v, id=v) for v in GCC_VERSIONS])
```
latk, line 113: This Noxfile uses the concept of a “version” a lot. I think talking about a `compiler` might be less confusing than a `version`, since the version could also refer to the gcovr or Python version.

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+    session.run("twine", "upload", "dist/*", external=True)
+
+
+def docker_container_id(version: str) -> None:
+    return "gcovr-qa-{}".format(version)
+
+
+@nox.session
+def docker_qa_build(session: "nox.session") -> None:
+    session_id = "docker_qa__build_version({})".format(GCC_VERSION2USE)
```
latk, line 169: Typo:

```suggestion
    session_id = "docker_qa_build_version({})".format(GCC_VERSION2USE)
```

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+@nox.session
+@nox.parametrize("version", [nox.param(v, id=v) for v in GCC_VERSIONS])
+def docker_qa_build_version(session: "nox.session", version: str) -> None:
+    set_environment(session, version, False)
+    session.run(
+        "bash",
+        "-c",
+        " ".join(
```
latk, line 189: Doesn't need bash. The shell interpolation for the `CC` and `CXX` variables can be done in Python.

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+@nox.session
+@nox.parametrize("version", [nox.param(v, id=v) for v in GCC_VERSIONS])
+def docker_qa_run_version(session: "nox.session", version: str) -> None:
+    set_environment(session, version, False)
+    session.run(
+        "bash",
+        "-c",
+        " ".join(
```
latk, line 232: Doesn't need bash, see above.

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+            [
+                "docker",
+                "run",
+                "--rm",
+                "-e",
+                "TESTOPTS",
```
latk, line 238: No TESTOPTS variable exists. Instead, we should probably stick posargs into an environment variable, pass it to the container like this, and then manually pass it to Nox' posargs in the Dockerfile.

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+@nox.session
+def docker_qa(session: "nox.session") -> None:
+    session_id = "docker_qa_build_version({})".format(GCC_VERSION2USE)
+    session.log("Notify session {}".format(session_id))
+    session.notify(session_id)
+    session_id = "docker_qa_run_version({})".format(GCC_VERSION2USE)
+    session.log("Notify session {}".format(session_id))
+    session.notify(session_id)
```
latk, line 257: Could simplify to `session.notify("docker_qa_version({GCC_VERSION2USE})")`

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+
+@nox.session
+def docker_qa_all_versions(session: "nox.session") -> None:
+    for version in GCC_VERSIONS:
+        session_id = "docker_qa_build_version({})".format(version)
+        session.log("Notify session {}".format(session_id))
+        session.notify(session_id)
+        session_id = "docker_qa_run_version({})".format(version)
+        session.log("Notify session {}".format(session_id))
+        session.notify(session_id)
```
latk, line 268: Can replace loop contents with `session.notify("docker_qa_version({version})")`

requirements.txt
```diff
@@ -1,13 +0,0 @@
...
-coverage
-flake8
-jinja2
-lxml
-pygments == 2.7.4
-pytest
-pytest-cov
-twine
-wheel
-yaxmldiff
```
latk, line 13: I'd have a preference for keeping the requirements.txt file as it is machine-readable and may assist with other tools. E.g. I install library dependencies (jinja2, lxml, pygments, pytest, yaxmldiff) in a development virtualenv so that my editor can provide autocomplete. But all the tools (black, cmake, codecov, coverage, flake8, pytest-cov, twine, wheel) can be managed via Nox.

latk (commented):
This is a great improvement – doing string manipulation in Make was getting totally out of hand. But it's interesting how different Nox' push-style notifications are from Make's pull-style dependencies.

Anyway, I have lots of comments. Including some errors, particularly about the `TEST_OPTS` variable that has been removed (good!) but not completely.

At some point, the administration tasks like adding copyright headers or the release-check could also be Noxified/Pythonized like this.

.github/workflows/test.yml
```diff
@@ -130,31 +129,21 @@ jobs:
...
       run: |
-        make lint
+        python3 -m nox --non-interactive --session lint
     - name: Check format of files
       if: ${{ ( matrix.python-version != 'pypy3' ) }}
       # Currently fail of this job is OK.
       continue-on-error: true
       run: |
-        make check-format
+        python -m nox --non-interactive --session black
```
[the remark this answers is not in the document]
Spacetown: Yes here should python3 be used.
Why don't use nox directly? Which nox is used if python2.7 and python3 is installed?

CONTRIBUTING.rst
```diff
@@ -240,28 +230,29 @@ are in ``gcovr.generator.html`` and respective modules.
...
    - unit tests in ``gcovr/tests``
    - integration tests in ``gcovr/tests``
    - documentation examples in ``doc/examples``
 
-- documentation build (``make doc``)
+- documentation build (``python3 -m nox --session doc``)
 
 The tests are in the ``gcovr/tests`` directory.
-You can run the tests with ``make test`` or ``python3 -m pytest -- gcovr doc/examples``.
+You can run the tests with ``python3 -m nox --session tests_version($CC)``.
```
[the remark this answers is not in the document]
Spacetown: Yes, this was the intention but we can document `tests` also as the first choice.

.github/workflows/test.yml
```diff
@@ -105,8 +105,7 @@ jobs:
     - name: Setup environment
       run: |
         echo "USE_COVERAGE=${{ ( ( matrix.os == 'windows-2019' ) && ( matrix.python-version == '3.7' ) ) || ( ( matrix.os == 'ubuntu-18.04' ) && ( matrix.gcc == 'gcc-6' ) && ( matrix.python-version == '3.8' ) ) }}" >> $GITHUB_ENV
-        echo "CC=${{ matrix.gcc }}" >> $GITHUB_ENV
-        echo "PYTHON=python" >> $GITHUB_ENV
+        echo "TEST_OPTS=--archive_differences" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
Spacetown: The first implementation was only configurable by environment variables. I'll change this to positional arguments.

.github/workflows/test.yml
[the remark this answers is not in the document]
latk: We no longer install Python 2 in the GH Actions so that might not be a concern any more. But I'm not sure if we're using the `python3 -m` style invocation due to some Windows issues.

CONTRIBUTING.rst
```diff
@@ -308,34 +299,38 @@ Each Makefile contains the following targets:
...
+``TEST_OPTS``. Run all tests after each change is a bit slow, therefore you can
+limit the tests to a specific test file, example project, or output format.
 For example:
 
 .. code:: bash
 
     # run only XML tests
-    make test TEST_OPTS="-k 'xml'"
+    export TEST_OPTS="-k 'xml'"
+    python3 -m nox --session tests
```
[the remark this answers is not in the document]
Spacetown: Yes, I'll update the docs.

doc/requirements.txt
```diff
@@ -1,4 +0,0 @@
-sphinx
-sphinx_rtd_theme
-sphinxcontrib-autoprogram == 0.1.5 ; python_version == "3.6"
-sphinxcontrib-autoprogram >= 0.1.5 ; python_version >= "3.7"
```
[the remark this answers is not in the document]
Spacetown: Ok. I'll do it this way.

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+    "noxfile.py",
+    "gcovr/gcov.py",
+    "gcovr/gcov_parser.py",
+]
+
+nox.options.sessions = [
+    "qa",
+    "lint",
+    "doc",
+    "tests_version({})".format(GCC_VERSION2USE),
```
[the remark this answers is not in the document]
Spacetown: The qa was introduced because it was also available in the makefile but we can remove it here.

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+    session.env["GCOVR_TEST_SUITE"] = "1"
+    session.env["CC"] = cc
+    session.env["CFLAGS"] = "--this_flag_does_not_exist"
+    session.env["CXX"] = cxx
+    session.env["CXXFLAGS"] = "--this_flag_does_not_exist"
+    session.env["GCOV"] = gcov
+
+
+@nox.session
+def qa(session: "nox.session") -> None:
```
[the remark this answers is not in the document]
Spacetown: Ok.

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+    for session_id in nox.options.sessions:
+        session.log("Notify session {}".format(session_id))
+        session.notify(session_id)
+
+
+@nox.session
+def lint(session: "nox.session") -> None:
+    session.install("flake8")
+    if platform.python_implementation() == "CPython":
+        session.install("black")
```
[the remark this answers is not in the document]
Spacetown: I'll add a comment to the black issue.

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+            )
+            session.run("python", "-m", "black", "--diff", *DEFAULT_TEST_DIRECTORIES)
+    else:
+        session.log(
+            f"Skip black because of platform {platform.python_implementation()}."
+        )
+
+
+@nox.session
+def black(session: "nox.session") -> None:
```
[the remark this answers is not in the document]
Spacetown: Ok. I'll add a documentation (https://nox.thea.codes/en/stable/config.html#session-description).

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+    session.run("python", "-m", "black", *args)
+
+
+@nox.session
+def doc(session: "nox.session") -> None:
+    session.install(
+        "sphinx",
+        "sphinx_rtd_theme",
+        "sphinxcontrib-autoprogram==0.1.5 ; python_version=='3.6'",
+        "sphinxcontrib-autoprogram>=0.1.5 ; python_version>='3.7'",
```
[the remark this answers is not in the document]
Spacetown: Ok.

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+    )
+    session.install("-e", ".")
+    session.run("bash", "-c", "cd doc && make html O=-W", external=True)
```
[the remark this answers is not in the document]
Spacetown: Ok.

noxfile.py
Spacetown: I'll do this.

noxfile.py
Spacetown: Is the result of `llvm-cov gcov --version` the same as from `gcov --version`. If yes, than this isn't needed. I'll check this.

noxfile.py
Spacetown: Ok.

noxfile.py
Spacetown: Ok, I'll rename this.

noxfile.py
Spacetown: Ok.

noxfile.py
Spacetown: I'll change this.

noxfile.py
Spacetown: Dito.

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+@nox.session
+def upload_wheel(session: "nox.session") -> None:
+    session.install("twine")
+    session.run("twine", "upload", "dist/*", external=True)
+
+
+def docker_container_id(version: str) -> None:
+    return "gcovr-qa-{}".format(version)
```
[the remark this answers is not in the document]
Spacetown: This was removed by accident.

noxfile.py
Spacetown: As f-string?

latk: err, right, sorry.

noxfile.py
Spacetown: You're right.

requirements.txt
Spacetown: Ok. Should the dependencies in nox be installed via the requirements.txt?

latk: Hmm, on second thought jinja2, lxml, and pygments will already be pulled in by `pip install -e .` which is part of the developer setup instructions, and also part of the Nox session doing testing. So my argument would only affect pytest and yaxmldiff, the latter of which is practically irrelevant.

In that case, leaving everything as it currently is might be OK.

.github/workflows/test.yml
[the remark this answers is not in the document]
Spacetown: On my MAC there is also a python installed:
```
python --version
Python 2.7.18 
```
But I see that no pip is available for installing nox.
I'll first fix the python to python3 and can check to use nox directly afterwards.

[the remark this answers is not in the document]
Spacetown: Changed to nox.

.github/workflows/test.yml
```diff
@@ -130,31 +128,21 @@ jobs:
...
-    - name: Cache pip
-      uses: actions/cache@v2
-      with:
-        # This path is specific to Ubuntu
-        path: ~/.cache/pip
-        # Look to see if there is a cache hit for the corresponding requirements file
-        key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt', 'doc/requirements.txt') }}
-        restore-keys: |
-          ${{ runner.os }}-pip-
-          ${{ runner.os }}-
```
latk, line 142: Is it OK to stop caching the packages? Or is it no longer reasonably possible to store the cache since the packages are now partially controlled via the Noxfile?

CONTRIBUTING.rst
```diff
@@ -240,28 +230,32 @@ are in ``gcovr.generator.html`` and respective modules.
...
    - integration tests in ``gcovr/tests``
    - documentation examples in ``doc/examples``
 
-- documentation build (``make doc``)
+- documentation build (``python3 -m nox --session doc``)
 
 The tests are in the ``gcovr/tests`` directory.
-You can run the tests with ``make test`` or ``python3 -m pytest -- gcovr doc/examples``.
+You can run the tests with ``python3 -m nox --session tests``
+for the default GCC version (used from environment ``CC``iv availabele, else gcc-5.
```
latk, line 247: Typos. I think phrasing it like this is clearer:

```suggestion
for the default GCC version (specified via ``CC`` environment variable, defaults to gcc-5).
```

admin/Dockerfile.qa
```diff
@@ -30,6 +27,6 @@ USER docker:docker
 # Unicode is necessary for some tools like "black" to work.
 ENV LC_ALL=C.UTF-8 LANG=C.UTF-8
 
-CMD ( echo docker | sudo -S -p "Running pip with sudo" $PYTHON -m pip install -e . ) && \
-  ( echo docker | sudo -S -p "Running chmod with sudo" chown -R docker:docker /gcovr ) && \
-  make qa
+CMD ( echo docker | sudo -S -p "Running chmod with sudo..." chown -R docker:docker /gcovr/gcovr ) && \
+  echo "\ndone\nStarting test..." && \
+  python3 -m nox --envdir /tmp/envs --non-interactive
```
latk, line 32: This will prevent caching of the Nox environments – each time the container runs, it will get a new tmp, and force all dependencies to be downloaded anew. This is going to be especially noticeable for the sizeable packages like cmake which weighs about 20MB.

But I understand that there might be clashes if the envs have the same names as the hosts's envs. Could we avoid this by using `--envdir ./.nox-containerized` instead?

doc/README.txt
```diff
@@ -3,7 +3,7 @@ sphinx.
 
 The necessary python packages have to be installed, e.g. with pip:
 
-    pip install -r requirements.txt
+    pip install nox
```
latk, line 6: This README refers to building the docs. For that, installing that directory's `requirements.txt` is in fact required. However, we could add a note that most of the time calling the top-level `nox -s doc` will be sufficient.

noxfile.py
```diff
@@ -0,0 +1,279 @@
...
+        session.install("black")
+    if session.posargs:
+        args = session.posargs
+    else:
+        args = DEFAULT_TEST_DIRECTORIES
+    session.run("flake8", *args)
+
+    if platform.python_implementation() == "CPython":
+        if session.posargs:
+            session.run("python", "-m", "black", *session.posargs)
```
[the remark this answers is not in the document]
latk: I would not expect `nox -s lint -- doc` to apply any changes. I would expect that to restrict the linting checks to the doc directory. If I wanted to reformat the files in the doc directory, I'd run `nox -s black -- doc`.

noxfile.py
```diff
@@ -0,0 +1,280 @@
...
+        session.log(f"Notify session {session_id}")
+        session.notify(session_id)
+
+
+@nox.session(python=False)
+@nox.parametrize("version", [nox.param(v, id=v) for v in GCC_VERSIONS])
+def docker_qa_run_compiler(session: nox.session, version: str) -> None:
+    """Run the docker container for a specific GCC version."""
+    set_environment(session, version, False)
+    session.env["NOX_POSARGS"] = " ".join([repr(a) for a in session.posargs])
```
latk, line 238: This variable is later expanded via `shlex.split()`. Then, we should probably use `shlex.join(session.posargs)` here as well.

latk (approved):
I think this is now in a state that can be merged. But as always, various comments :)

noxfile.py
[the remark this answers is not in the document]
latk: I ran `mypy noxfile.py` as a type checker and it runs fine when using `nox.Session` (uppercase S), but complains about using `nox.session` (lowercase s).

.github/workflows/test.yml
Spacetown: I don't know how the cash works. The packages are installed by the noxfile. Maybe we can use `hashFile('noxfile.py')` for the cache.

admin/Dockerfile.qa
Spacetown: I was alos thinking about this. We can't use the default directory and we need to use a different directory for each docker cointainer. I'll change this.

noxfile.py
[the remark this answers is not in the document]
Spacetown: The decorator is a lowercase `nox.session` and the class should be uppercase `nox.Session`.

noxfile.py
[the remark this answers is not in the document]
Spacetown: I'll move the default args to the black session (needed for the GitHub Actions) and for the lint I'll notify the black session with the needed arguments.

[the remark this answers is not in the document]
Spacetown: In the Makefile lint only called flake8 and check-format was black. If lint also calls black we don't need the check-format step in the CI. What do you think?

doc/README.txt
Spacetown: Te requirements.txt was restored so I'll revert the change and add the nox -s doc.

noxfile.py
Spacetown: I'll change this.

Spacetown: `shlex.join` was introduced in python 3.8. I'll use this if available, else a copy of the function is used.

noxfile.py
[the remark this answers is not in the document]
latk: I think the idea of keeping these steps separate in the CI was to get better error messages about which step of the QA process failed. Right now, running Black isn't necessarily attractive because it produces huge diffs that fill up the terminal. But long-term, it makes sense to just run all linting tools together as a single Nox session so that developers don't have to manually select a particular step. Not sure what the correct tradeoffs are here, so for the time being I'd just leave everything as it is now.

Issue #508, opened 2021-08-06
Exception when searching for headers in src and sources in .libs directory


kcudnik:
**Describe the bug**
when headers are in different directories then sources, i get this error:

header exception:
```
(gcovr could not infer a working directory that resolved it.)
Traceback (most recent call last):
  File "/usr/bin/gcovr", line 11, in <module>
    load_entry_point('gcovr==4.2', 'console_scripts', 'gcovr')()
  File "/usr/lib/python3/dist-packages/gcovr/__main__.py", line 253, in main
    print_reports(covdata, options, logger)
  File "/usr/lib/python3/dist-packages/gcovr/__main__.py", line 368, in print_reports
    generator(covdata, output.value, options)
  File "/usr/lib/python3/dist-packages/gcovr/html_generator.py", line 248, in print_html_report
    with io.open(data['FILENAME'], 'r', encoding=options.source_encoding,
FileNotFoundError: [Errno 2] No such file or directory: 'lib/src/Notification.h'
```
**Notification.h is located in lib/inc/ not in lib/src**

source exception:
```
        (gcovr could not infer a working directory that resolved it.)
Traceback (most recent call last):
  File "/usr/bin/gcovr", line 11, in <module>
    load_entry_point('gcovr==4.2', 'console_scripts', 'gcovr')()
  File "/usr/lib/python3/dist-packages/gcovr/__main__.py", line 253, in main
    print_reports(covdata, options, logger)
  File "/usr/lib/python3/dist-packages/gcovr/__main__.py", line 368, in print_reports
    generator(covdata, output.value, options)
  File "/usr/lib/python3/dist-packages/gcovr/html_generator.py", line 248, in print_html_report
    with io.open(data['FILENAME'], 'r', encoding=options.source_encoding,
FileNotFoundError: [Errno 2] No such file or directory: 'lib/src/.libs/SaiInterface.cpp'
```
**SaiInterface.cpp is located in lib/src not lib/src/.libs/**

**To Reproduce**
1. clone sonic-swss-common (https://github.com/Azure/sonic-swss-common)
2. compile with CXXFLAGS="" LDFLAGS="" ./configure --enable-code-coverage
3. install swss-common deb package
4. enable unix socket on redis server: vim /etc/redis/redis.conf anddng 2 lines:
 - unixsocket /var/run/redis/redis-server.sock
 - nixsocketperm 777
5. cd /var/run/redis/
6. ln -s redis-server.sock redis.sock
[...]

PR #521, opened 2021-10-20
Add gcc abspath

Spacetown:
Add test for option mentioned in https://github.com/gcovr/gcovr/issues/508#issuecomment-944903456 and update docs.


gcovr/tests/gcc-abspath/Makefile
```diff
@@ -0,0 +1,15 @@
+all:
+	cd subfolder&& $(MAKE)
+
+run: html
+
+html:
+	cd subfolder&& ./testcase
+	mv -f --target-directory . subfolder/*.gc??
+	$(GCOVR) -d --html-details -o coverage.html
```
latk, line 9: Using the JSON output format might be easier to manage though.

latk (approved):
LGTM! But I would add a note to the changelog.

PR #525, opened 2021-11-08
Don't use temporary directory for running gcov

Spacetown:
As discussed in #524 on MacOs there is a problem with the temporary directory. The files generated by gcov are prefixed with the name of the data file (Option `--long-path-names`) to be able to run `gcov` in parallel.
The regular expression to catch the "Source not found" error is extended to match the message for a clang based gcov command.
The test of the example is extended to test all available output formats except the detailed html.

Maybe we should also extend the check of the gcovr output to also try the next working directory if there is no coverage data in the gcov file instead of only rely on the "Source not found" message.

Closes #524

PR #526, opened 2021-11-10
Fix C comments and dependency on spaces.

Spacetown:
As discussed in #525 C comments aren't ignored in `_is_non_code` and there is a dependency on spaces in `} else {` (see extended tests). A empty line or a line with a C++ coment is treated as non code but a C comment is treated as code and decreases the coverage. 

[no changelog]


Spacetown:
Failed test is OK. The lines `} // GCOV_EXCL_STOP` and `} // LCOV_EXCL_STOP´ are now `non_code`. In the previous version only a emty `}` was treated as `non_code`.

gcovr/gcov_parser.py
```diff
@@ -1029,20 +1029,30 @@ def _is_non_code(code: str) -> bool:
...
     False
     """
 
-    # This variant is more flexible, but would change JSON reports
-    # code = code.replace("{", "").replace("}", "").strip()
-    code = code.strip().replace("{", "").replace("}", "")
-    return len(code) == 0 or code.startswith("//") or code == "else"
+    code = _CPP_STYLE_COMMENT_PATTERN.sub("", code)
+    code = _C_STYLE_COMMENT_PATTERN.sub("", code)
+    code = code.replace(" ", "").strip()
```
latk, line 1054: The replace() is useless in this context.

```suggestion
    code = code.strip()
```

latk (commented):
Extremely good change, especially since the excl-branch test was previously wrong (claiming that a line was excluded although it wasn't).

However, the changes to the doc/examples seem out of place in this context. If the files should be removed, that should be discussed as part of #543.

Spacetown:
> However, the changes to the doc/examples seem out of place in this context. If the files should be removed, that should be discussed as part of #543.

That was an commit by accident, I will remove the changes.

latk (approved):

PR #527, opened 2021-11-13
Add tests for gcc-9 and clang-13

Spacetown:
Add test for `gcc-9`, the reference is copied from `gcc-8`. The new version of gcc is only available for ubuntu 20.04, therefore the creation of the docker image is adapted. I also switched to the new ubuntu for `gcc-8` and `clang-10`.

This is needed to start work on #282.


Spacetown:
@latk I've checked the available versions, shall we extend the tests also with `gcc-10`, `gcc-11`, `clang-11`, `clang-12`and `clang-13` or some of them?

PR #531, opened 2021-12-04
Only output 0.0% if really 0.0%.

Spacetown:
This is the counterpart for #389.


Spacetown:
@latk It seems that there are output files which haven't been compared in the past, e.g. coverage-excl-throw.tx. Only for HTML files a txte after the coverage was possible. I extended the test driver to find also coverage.999.txt and coverage.01.txt and now the files of the olde tests are compared also. Should I create a seperate PR for the fix of this issue?

Spacetown:
@latk In the text output the percent value is truncated, should I use 1% if there is at least one covered branch?

latk:
Oops, good that we found the reference files that were being ignored. Doesn't have to be a separate PR but could the order of commits be changes so that it is clear that updating the reference files has nothing to do with changing how percentages are represented?

If we want to round away from 0%, it should match the precision in that specific report type.

In the new test, that is a *lot* of code being added. I think it's possible to simplify by only considering branch coverage, and using macros to create the highly branchy code. Something like:

```cpp
#define BRANCH1 if(++x == number) { return; }
#define BRANCH4 BRANCH1 else BRANCH1 else BRANCH1 else BRANCH1
#define BRANCH16 BRANCH4 else BRANCH4 else BRANCH4 else BRANCH4
#define BRANCH64 BRANCH16 else BRANCH16 else BRANCH16 else BRANCH16
#define BRANCH256 BRANCH64 else BRANCH64 else BRANCH64 else BRANCH64
#define BRANCH1024 BRANCH256 else BRANCH256 else BRANCH256 else BRANCH256
#define BRANCH4096 BRANCH1024 else BRANCH1024 else BRANCH1024 else BRANCH1024

void use_number_of_branches( int number ) {
  int x = 0;
  BRANCH4096 else { return; }
}
```

Idea credit to the [billion laughs / XML bomb](https://en.wikipedia.org/wiki/Billion_laughs_attack) exploit.

I wouldn't have implemented this feature myself. While it's useful to avoid showing “everything covered, 100%” if that isn't the case, I think there's limited value in distinguishing “literally entirely uncovered” from “almost entirely uncovered”. But if you want it, I trust you on this. (In general, I don't want to decide everything about gcovr.)

Spacetown:
Ok. I'll reorder the commits and use the macros as suggested.

Spacetown:
I slept one night and will close this PR. For the fix I'll create a new one.

PR #533, opened 2021-12-08
Fix globbing of reference data

Spacetown:
Fix bug found in #531.


latk (approved):
LGTM! Thank you for finding and fixing this problem :)

PR #537, opened 2021-12-15
Add test to install wheel after build.

Spacetown:
As mentioned in #465 for GitHub Actions this is a general implementation of testing the installation of the wheel.


latk (approved):
This kind of check seems extremely sensible! Thanks for implementing it.

PR #543, opened 2021-12-26
Split examples because the details are changed after every test run.

Spacetown:
Since #525 where the test of the examples was extended with csv, json and html the files with the details where changed on every run. Now the HTML test is splitted and the call for the details is only used when generating the docs.

[no changelog]


latk:
Splitting the examples is quite reasonable. But skipping an example for the tests might not be a good idea. I just implemented #546 which makes it possible to have reproducible reports that do not change the timestamp in the report. Could you take a look to see if that kind of feature would make sense in this context?

PR #546, opened 2021-12-27
Add --timestamp option for reproducible reports.

latk:
Certain report formats include a timestamp (HTML, Coveralls, Cobertura). It may however be desirable to attribute the report to a different time:

* e.g. date of a commit instead of time when gcovr was run
* CI server might be running with wrong timezone
* the gcovr test suite

Currently, the gcovr test suite scrubs timestamps from the generated files in order to enable comparison. In the future, this option could be used instead. The HTML-example test was already modified to use fixed timestamps in order to avoid unnecessary Git diffs (the html-details example outputs are committed).

The supported date formats for the --timestamp option are the Unix epoch and formats that are like RFC-3339 ("ISO dates" without the more insane features). Time zones are optional.

This should be able to directly understand the outputs from the following GNU
date commands:

    date --rfc-3339=seconds
    date --iso-8601=seconds
    date +%s

It will also understand the following date formats used by Git:

    %aI, %cI, --date=iso-strict, --date=iso8601-strict
    %at, %ct, --date=unix
    %ad, %cd (when appropriate --date format is selected)

See: [`git show` pretty formats](https://git-scm.com/docs/git-show#_pretty_formats), [`git log` --date format](https://git-scm.com/docs/git-log#Documentation/git-log.txt---dateltformatgt)

Not supported is the RFC 2822 / RFC 5322 format known from emails (e.g. `Mon, 27 Dec 2021 19:09:44 +0100`). Also unsupported are timestamps with sub-second accuracy.

While the timestamp parser will guess the format (Epoch or RFC-3339) correctly, it is also possible to specify a particular format using a prefix:

    gcovr --html --timestamp=epoch:1640606727
    gcovr --html --timestamp=@1640606727
    gcovr --html --timestamp="rfc3339:2021-12-27 13:05:27"

That is, the supported prefixes are:

| prefix     |  description
|------------|--------------------------------
| `@`        |  Unix timestamps, matches GNU date
| `epoch:`   | Unix timestamps
| `rfc3339:` | RFC-3339 like timestamps

This will enable backwards-compatible future evolution of the supported formats. The prefixes are not yet mentioned in the `--help` message.

Context: #543 proposed disabling a test because it overwrote the timestamp in a committed file. Instead of disabling the test, this option prevents the timestamp from being changed.

Why this shouldn't be merged, perhaps: that's a lot of extra code for a little feature. Admittedly, a lot of the code is doctests. But my desire to support a reasonably strict timestamp format that can be maintained in a backwards-compatible manner seems to be a bad match for the Python standard library. The `datetime.fromisoformat()` function is almost suitable except that it's overly lax with respect to the time-separator (`2021-12-27318:53:12`) would parse successfully – note the `3` in the separator position) but also overly strict with respect to timezones (doesn't understand the extremely common `Z` for UTC-timezone).

### DIFF
--- a/doc/source/guide-timestamps.rst
+++ b/doc/source/guide-timestamps.rst
@@ -1,0 +1,114 @@
+.. _timestamps:
+
+Reproducible Timestamps
+=======================
+
+In some cases, it may be desirable to list a specific timestamp in the report.
+Timestamps are shown in
+the :ref:`html_output`,
+:ref:`coveralls_output`,
+and the :ref:`cobertura_output`.
+This can be achieved via the :option:`--timestamp <gcovr --timestamp>` option.
+This option does not affect the modification times or other filesystem metadata.
+
+.. versionadded:: NEXT
+
+   The :option:`gcovr --timestamp` option.
+
+
+Timestamp Syntax
+----------------
+
+The timestamp option understands different formats:
+Unix timestamps and RFC-3339 timestamps.
+
+Unix timestamps (also known as Posix time or Epoch)
+are the number of seconds since 1 Jan 1970.
+These timestamps are always resolved in the UTC timezone.
+Example usage:
+
+.. include:: ../examples/example_timestamps.sh
+   :code: bash
+   :start-after: #BEGIN simple epoch
+   :end-before: #END simple epoch
+
+`RFC 3339 <https://datatracker.ietf.org/doc/html/rfc3339>`_
+specifies a reasonable subset of ISO-8601 timestamps.
+This is the ``YYYY-MM-DDThh:mm:ss`` format,
+optionally followed by a timezone offset (``+hh:mm``, or ``Z`` for UTC).
+Example usage without a timezone:
+
+.. include:: ../examples/example_timestamps.sh
+   :code: bash
+   :start-after: #BEGIN simple RFC 3339
+   :end-before: #END simple RFC 3339
+
+Example usages that show equivalent specifications for UTC timestamps:
+
+.. include:: ../examples/example_timestamps.sh
+   :code: bash
+   :start-after: #BEGIN RFC 3339 with UTC timezone
+   :end-before: #END RFC 3339 with UTC timezone
+
+Differences and clarifications with respect to RFC-3339:
+
+* the time zone may be omitted
+* the date and time parts may be separated by a space character
+  instead of the ``T``
+* the date is parsed in a case insensitive manner
+* sub-second accuracy is not currently supported
+
+Additional formats may be added in the future.
+To ensure that timestamps are handled in the expected manner,
+it is possible to select a particular timestamp syntax with a prefix.
+
+* Epoch timestamps can be selected with a ``@`` or ``epoch:`` prefix.
+* RFC-3339 timestamps can be selected with a ``rfc3339:`` prefix.
+
+Examples of prefixes:
+
+.. include:: ../examples/example_timestamps.sh
+   :code: bash
+   :start-after: #BEGIN prefixes
+   :end-before: #END prefixes
+
+
+Using timestamps from Git commits
+---------------------------------
+
+As an example of using the timestamp feature,
+we might want to attribute a coverage report
+to the time when a Git commit was created.
+Git lets us extract the commit date from a commit
+with the `git show <https://git-scm.com/docs/git-show>`_ command.
+For the current HEAD commit::
+
+  git show --no-patch --format=%cI HEAD
+
+This can be combined into a Bash one-liner like this:
+
+.. include:: ../examples/example_timestamps.sh
+   :code: bash
+   :start-after: #BEGIN git commit
+   :end-before: #END git commit
+
+Each Git commit has two dates, the author date and the committer date.
+This information can be extracted with various format codes,
+e.g. ``%aI`` for the author date and ``%cI`` for the committer date.
+These format codes are also available in different formats.
+The supported Git formats are:
+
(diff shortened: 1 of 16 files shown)


gcovr/configuration.py
```diff
@@ -64,6 +65,15 @@ def check_input_file(value, basedir=None):
     return os.path.abspath(value)
 
 
+def timestamp(value: str) -> datetime.datetime:
+    from .timestamps import parse_timestamp  # lazy import
+    try:
+        return parse_timestamp(value)
+    except ValueError as ex:
```
Spacetown, line 72:
```suggestion
    except ValueError as ex:  # pragma: no cover
```

gcovr/timestamps.py
```diff
@@ -0,0 +1,206 @@
...
+    >>> parse_timestamp("@1640606727").isoformat()
+    '2021-12-27T12:05:27+00:00'
+    >>> parse_timestamp("epoch:1640606727").isoformat()
+    '2021-12-27T12:05:27+00:00'
+
+    Examples of parsing RFC 3339 like timestamps:
+
+    >>> parse_timestamp("2021-12-27 13:05:27").isoformat()  # no prefix
+    '2021-12-27T13:05:27'
+    >>> parse_timestamp("rfc3339:2021-12-27 13:05:27").isoformat()
```
Spacetown, line 52:
```suggestion
    >>> parse_timestamp("rfc3339:2021-12-27T3:05:27").isoformat()
```
To also test the default separator.

gcovr/timestamps.py
```diff
@@ -0,0 +1,206 @@
...
+      ...
+    ValueError: timestamp separator must be 'T' or space
+
+    Examples for timezone offsets:
+
+    >>> _parse_rfc3339("2021-12-27 13:05:27Z").isoformat()
+    '2021-12-27T13:05:27+00:00'
+    >>> _parse_rfc3339("2021-12-27 13:05:27+12:30").isoformat()
+    '2021-12-27T13:05:27+12:30'
+    >>> _parse_rfc3339("2021-12-27T13:05:27-07:23").isoformat()
```
Spacetown, line 141:
```suggestion
    >>> _parse_rfc3339("2021-12-27t13:05:27-07:23").isoformat()
```
To test also the lowercase separator.

Spacetown (commented):
LGFM. Only change doctest to also use lowercase separator.
Can you extend the documentation to a example which uses the commit dat from GIT for the option?
```shell
gcovr --timestamp=$(git show -s --format=%cI HEAD) ...

**Review by Spacetown — APPROVED** [review]

File: gcovr/configuration.py (hunk 1 above)
Comment (latk) [reply gcovr/configuration.py:72]: I added a test for the exception handler (and promptly found a bug with the exception formatting :grimacing:)

File: gcovr/timestamps.py (hunk 3 above)
Comment (latk) [reply gcovr/timestamps.py:145]: Additional test cases were added.

File: doc/source/guide-timestamps.rst [hunk 4]
```diff
@@ -0,0 +1,114 @@
...
+   :end-before: #END prefixes
+
+
+Using timestamps from Git commits
+---------------------------------
+
+As an example of using the timestamp feature,
+we might want to attribute a coverage report
+to the time when a Git commit was created.
+Git lets us extract the commit date from a commit date
```
Spacetown, line 82:
```suggestion
Git lets us extract the commit date from a commit
```

latk:
I updated the PR to add more tests and a documentation page.

The docs were added to a separate page instead of making the Guide file even larger. Unfortunately, the toctree to link the pages had to be added to a separate heading due to Sphinx limitations. **Future directions:** could make sense to split up the guide into smaller pages.
    
The new documentation includes a number of tests. These just ensure that various timestamps are parsed successfully and provide no further checks. If the tests are executed from within a Git repository, the HEAD timestamp is used for testing. This could lead to some fragility, but I'm reasonably confident that the timestamp parsing will always succeed. Note: the tests are skipped in docker-qa since the container doesn't include Git.
    
Unfortunately these tests are pretty slow. In the future, it might be worth considering strategies to improve gcovr startup time.

The tests seem to fail on Windows when running `bash example_timestamps.sh`. The error message is:

    Windows Subsystem for Linux has no installed distributions.

Not sure what this means – that no `bash` can be invoked even though it is clearly available for some other tests? Should I just skip this test on Windows?

Spacetown:
GNU make on windows use sh if available and a sh.exe is installed with MSYS2 (https://github.com/gcovr/gcovr/runs/4651563240?check_suite_focus=true#step:4:80).
On Windows 10 you can call for example `python` and the Microsoft store is opened. There are placeholders for several executables which redirect to the stor. I don't know who is needing this feature.

latk:
Will fix the typo and try with sh – Bash is not needed for the test script.

But it seems the release check in the deploy workflow is failing. I'll have to fix that before I can merge this PR. This probably means that the deploy workflow can't be run on every PR, but I think the workflow can be simplified to de-risk it a bit.

Spacetown:
> But it seems the release check in the deploy workflow is failing. I'll have to fix that before I can merge this PR. This probably means that the deploy workflow can't be run on every PR, but I think the workflow can be simplified to de-risk it a bit.

With the option `--no-verify-docs-next-version` this check can be turned of. Can you use the option depending on the trigger? E.g.:
https://github.com/gcovr/gcovr/blob/2b161648ceffbbffc8008d32c38a18318974eea7/.github/workflows/deploy.yml#L76

PR #548, opened 2021-12-29
Change attribute root in json summary

Spacetown:
If a output file is given, the value of the attribute is calculated from the filename instead from the directory of the file.


gcovr/writer/json.py
```diff
@@ -21,8 +21,14 @@
 import functools
```
latk, line 21: Once a file is reformatted with black, it should be added to the known-good list of files in the Noxfile.

CHANGELOG.rst
```diff
@@ -16,6 +16,7 @@ Improvements and new features:
 - Fix junctions on windows for Python 3.8 (:issue:`535`)
 - Can create reproducible reports with the :option:`--timestamp` option (:issue:`546`)
 - Add "Goto function" in html details. (:issue:`515`)
+- Change the value of the attribute "root" in JSON summary report. (:issue:`548`)
```
latk, line 19: This does not communicate clearly what the "Change" means. But it might be sufficient to indicate that the previous behaviour was incorrect:
```suggestion
- Fix "root" path in JSON summary report. (:issue:`548`)
```

gcovr/writer/json.py
```diff
@@ -28,7 +28,7 @@
 
 
 JSON_FORMAT_VERSION = "0.2"
-JSON_SUMMARY_FORMAT_VERSION = "0.3"
+JSON_SUMMARY_FORMAT_VERSION = "0.4"
```
latk, line 37: Gcovr 5.0 was released with format version "0.2", so we can leave it at 0.3.

latk (approved):

gcovr/writer/json.py
Spacetown: Of course.

PR #549, opened 2022-01-06
Add support for drive substitution on windows

Spacetown:
As discussed in #365 and first PR #535 a test with drive substitutions of windows is added. Several fixes where needed to get the tests working:
- Only add a directory separator if there isn't one at the end.
- Fix RelativeFilter to use `realpath` of root directory.
- Fix HTML generation (`realpath`missing for directory).
- Match filter against filename  and the `realpath` of filename.


gcovr/tests/simple1-drive-subst/Makefile
```diff
@@ -0,0 +1,56 @@
...
+	cd $(GCOVR_TEST_DRIVE)/simple1-drive-subst && ./testcase
+	cd $(GCOVR_TEST_DRIVE)/simple1-drive-subst && $(GCOVR) -d -x -o coverage.xml
+
+html: subst
+	cd $(GCOVR_TEST_DRIVE)/simple1-drive-subst && ./testcase
+	# these test cases also cover some CSS-linking permutations
+	cd $(GCOVR_TEST_DRIVE)/simple1-drive-subst && $(GCOVR) --html coverage-summary-includecss.html
+	cd $(GCOVR_TEST_DRIVE)/simple1-drive-subst && $(GCOVR) --html coverage-summary-linkcss.html --no-html-self-contained
+	cd $(GCOVR_TEST_DRIVE)/simple1-drive-subst && $(GCOVR) --html-details coverage-details-includecss.html --html-self-contained
+	cd $(GCOVR_TEST_DRIVE)/simple1-drive-subst && $(GCOVR) --html-details coverage-details-linkcss.html
```
latk, line 31: Do we expect significant differences depending on whether or not CSS is included? It seems this was copied from another test that *also* happened to cover CSS things as a side effect, but testing the same things here would just slow down the test suite. Similarly, some of the other report formats might not be needed for this test.

gcovr/tests/test_gcovr.py
```diff
@@ -233,6 +233,11 @@ def pytest_generate_tests(metafunc):
                 continue
 
             marks = [
+                pytest.mark.xfail(
+                    name == "simple1-drive-subst"
+                    and not is_windows,
+                    reason="drive substitution only available on windows",
+                ),
```
latk, line 240: I think this should be marked as `skip` rather than `xfail`. Xfail tests are still attempted, but a failure does not fail the entire test suite.

gcovr/utils.py
```diff
@@ -280,13 +280,15 @@ def match(self, path):
...
+        if not abspath.endswith(os.sep):
+            abspath += os.sep
         os_independent_dir = abspath.replace(os.path.sep, '/')
-        pattern = re.escape(os_independent_dir + '/')
+        pattern = re.escape(os_independent_dir)
         super(DirectoryPrefixFilter, self).__init__(pattern)
 
     def match(self, path):
-        normpath = os.path.normpath(path)
+        normpath = os.path.normpath(path).replace(os.sep, "/")
```
latk, line 291: If we're already working on this, shouldn't the `normpath()` call be replaced by `realpath()`?

gcovr/__main__.py
```diff
@@ -312,7 +312,6 @@ def collect_coverage_from_tracefiles(covdata, options, logger):
             for trace_file in trace_files:
                 datafiles.add(normpath(trace_file))
 
-    options.root_dir = os.path.abspath(options.root)
```
latk, line 312: Is this safe to remove? Do other parts of the code rely on the `root_dir` being absolute?

gcovr/gcov.py
```diff
@@ -575,12 +575,18 @@ def apply_filter_include_exclude(filename, include_filters, exclude_filters):
         excluded (bool): True when filename failed the exclude_filters
     """
 
-    filtered = not any(f.match(filename) for f in include_filters)
+    filtered = not any(
+        f.match(filename) or f.match(realpath(filename))
```
latk, line 579: If the filters need a realpath() to work, the filters should do this themselves. Most filters already do this correctly, only DirectoryPrefixFilter needs this.

latk (commented):
It's great to have these tests for Windows (and to see them pass)! Thank you! See various comments below.

gcovr/tests/test_gcovr.py
Spacetown: You're right.

gcovr/utils.py
Spacetown: I can check what happens. In my opinion normpath should be removed in all places.

gcovr/__main__.py
Spacetown: The same thing is done in the main function (line230).

gcovr/gcov.py
Spacetown: I'll check this.

gcovr/tests/test_gcovr.py
Spacetown: I also change the others which will never work to `skip`.

gcovr/gcov.py
Spacetown: Ok, it's working.

gcovr/utils.py
Spacetown: I've replaced this.

gcovr/tests/simple1-drive-subst/Makefile
```diff
@@ -0,0 +1,53 @@
+
+GCOVR_TEST_DRIVE ?= Z:
+
+override GCOVR += --verbose
+
+subst:
+ifeq ($(wildcard $(GCOVR_TEST_DRIVE)/*.*),)
+	cmd.exe /C "subst $(GCOVR_TEST_DRIVE) .." || exit 0
+endif
```
latk, line 9: What will happen if a Z:/ drive already exists but is used for a different purpose? Can the test destroy data? When we do similar stuff on Linux, we ensure that we're running in an isolated Docker container. That's not an option with Windows (AFAIK) but some level of safety might be desirable.

Perhaps the reasonable approach would be:

* do not hardcode a test drive in the Makefile
* if the Makefile is not given a test drive, show an error and fail the test
* in some part of the Python code (Noxfile? test_gcovr?) find a unused drive letter and pass it in a variable to the Makefile

Spacetown: The subst will fail if there is already a drive connected. I dropped the idea of searching a free drive letter because if the test fails or is aborted or the option `--skip_clean` is used the substitution isn't removed and a new drive is used in the next call until all drive letters are in use.
I'll change the makefile to subst the drive immediately before and remove the drive after each call.

gcovr/gcov.py
Spacetown: Test was green because of `skip` instead of `skipif`. Now the test is failing.
The `DirectoryPrefixFilter` fails for the realpath here. Thats the reason wy abspath and realpath was used.

gcovr/tests/simple1-drive-subst/Makefile
Spacetown: Now it's working with the dynamic selection of a free drive.

gcovr/gcov.py
Spacetown: Now we check in `DirectoryPrefixFilter` the abspath and the realpath. This is needed because realpath is used for the filter directory and if the filter directory is a link itself the the realpath is also needed for the file.

Spacetown:
@latk Can we merge this before checking the problem with `realpath`?

Spacetown:
Replaced by #568.

PR #550, opened 2022-01-07
Drop support for Python 3.6

latk:
Python 3.6 was EOL'ed on 23 Dec 2021. In line with gcovr's support policy for Python versions, this pull request removes support for Python 3.6.

References: [PEP 494 (Python 3.6 Release Schedule)](https://www.python.org/dev/peps/pep-0494/), [endoflife.date/python](https://endoflife.date/python)

This has no strong immediate benefits, but will allow us to use more modern Python features in the future. Highlight features of Python 3.7 include:

* [`dataclasses`](https://docs.python.org/3/library/dataclasses.html) to easily define struct-like classes. Could simplify a lot of places where we currently pass around tuples!
* `from __future__ import annotations` for easier type annotations
* guaranteed dict insertion order preservation
* significant improvements to asyncio. Could allow us to replace the multithreading approach for running concurrent gcov processes.

Following this PR, I plan to submit patches that introduce more type checking, with the eventual goal of enabling `mypy` as a linter. While using a type checker sometimes requires writing the code a bit differently, it is my experience that this can detect some bugs up front – a good counterpart to our test suite. It is also my hope that such type annotations will make other refactorings safer, make it easier to navigate the codebase, and help other people contribute more easily.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,29 +69,57 @@ jobs:
 
     needs: [milestone-check, changelog-check]
 
+    # Testing strategy
+    # ----------------
+    #
+    # We have different axes of the testing matrix:
+    #
+    # OS: Linux, Windows
+    # Compiler: GCC-5, GCC-6, GCC-8, Clang-10
+    # Python: 3.7 -- 3.10, pypy3
+    #
+    # Instead of testing all combinations, we try to achieve full coverage
+    # across each axis. The main test matrix just represents the Python axis on
+    # Linux. The OS=Windows and Compiler axis are added manually.
+    #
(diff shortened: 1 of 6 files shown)


Spacetown (approved):
LGFM.

Spacetown:
Should this be part of the 5.1?

latk:
We *could* hold off from this change until 5.1 is released so that 5.1 would be the final gcovr version with Python 3.6 support.

But I see no advantage from that since Python 3.6 is end-of-life. So I'd merge it now if that is fine with you.

Spacetown:
Please merge it back.

PR #553, opened 2022-01-08
Fix the if condition for exclusion markers.

Spacetown:
The condition was always true.

[no changelog]


latk:
good catch! Thanks for fixing this :)

Spacetown:
Catched this while refactoring to get the source lines for the decision coverage (#350) and markers only ones.

PR #554, opened 2022-01-09
Add function overview to html details

Spacetown:
If a detailed report is generated a overview page with alphabetical ordered table of all functions is created. The lines are linked to the details page where the function is defined.


gcovr/templates/base.html
```diff
@@ -16,21 +16,29 @@
...
-    <hr/>
+      <hr/>
+    </header>
 
-    {% block content %}
-    {% endblock %}
+    <navigation>
+      {% block navigation %}
+      {% endblock %}
+    </navigation>
```
latk, line 35: There is no `<navigation>` element, but there is a `<nav>` (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav)

gcovr/templates/base.html
```diff
@@ -16,21 +16,29 @@
...
 
-    <hr/>
+    <article>
+      {% block content %}
+      {% endblock %}
+      <hr/>
+    </article>
```
latk, line 41: I don't think using `<article>` makes sense here. If you want to surround this part with a HTML element, consider `<main>` or `<div>`.

latk (approved):
The function list works and is good :+1:

I have two comments about the HTML5 commit, though.

PR #556, opened 2022-01-16
Update the test driver to share the reference data between the different compiler versions

Spacetown:
This updates the test driver to search the reference data backwards for the different compiler versions and uses the first matching file. With this we can see the differences between the compiler versions and need less files to update if a format chenges.


gcovr/tests/test_gcovr.py
```diff
@@ -380,7 +401,22 @@ def test_build(
...
-                shutil.copyfile(coverage_file, reference_file)
+                if CC_REFERENCE in reference_file:
+                    shutil.copyfile(coverage_file, reference_file)
+                else:
+                    reference_dir = os.path.join("reference", CC_REFERENCE)
+                    try:
+                        os.makedirs(reference_dir)
+                    except FileExistsError:
+                        # directory already exists
+                        pass
```
latk, line 412: can be replaced with `os.makedirs(reference_dir, exist_ok=True)`

gcovr/tests/test_gcovr.py
```diff
@@ -143,12 +157,23 @@ def run(cmd, cwd=None):
...
+    seen_files = set([])
+    for reference_dir in reference_dirs:
+        for pattern in output_pattern:
+            for dir in REFERENCE_DIRS:
+                if (platform.system() == "Windows") and os.path.isdir(reference_dir + "-Windows"):  # pragma: no cover
+                    for file in glob.glob(os.path.join(f"{dir}-Windows", pattern)):
+                        if os.path.basename(file) not in seen_files:
+                            coverage = os.path.basename(file)
+                            seen_files.add(coverage)
+                            yield coverage, file
```
latk, line 170: Code duplication between the `if` and `elif` branches. This duplication can likely be avoided by appending `-Windows` if appropriate, similar to the previous code:

```python
if platform.system() == "Windows" and os.path.isdir(f"{reference_dir}-Windows"):
    reference_dir += "-Windows"
```

Also, there is no need to mark Windows-only branches as `pragma: no cover` as we do collect coverage on Windows.

gcovr/tests/test_gcovr.py
```diff
@@ -143,12 +157,23 @@ def run(cmd, cwd=None):
...
-    assert os.path.isdir(reference_dir), "Reference directory exists"
-    for pattern in output_pattern:
-        for reference in glob.glob(os.path.join(reference_dir, pattern)):
-            coverage = os.path.basename(reference)
-            yield coverage, reference
+def find_reference_files(reference_dirs, output_pattern):
+    seen_files = set([])
+    for reference_dir in reference_dirs:
+        for pattern in output_pattern:
+            for dir in REFERENCE_DIRS:
```
latk, line 164: Looping over both `reference_dirs` and `REFERENCE_DIRS` seems redundant? I think they contain the same values, so that the same files will be selected multiple times. Parts of the following code use `reference_dir`, other parts use `dir`. I think this is incorrect.

gcovr/tests/test_gcovr.py
```diff
@@ -380,7 +401,22 @@ def test_build(
         except Exception as e:  # pragma: no cover
             whole_diff_output += "  " + str(e) + "\n"
             if update_reference:
```
latk, line 403: I feel like the code for update-reference, generate-reference, and find-reference-files should each have their own function since they are closely related with each other, but distract from the main flow of the `test_build()` function.

latk (commented):
This is a brilliant change that will make subsequent code reviews much easier! Thank you!

Aside from the inline comments – do you think it would be possible to add a small test that ensures that there are no duplicate reference files? I'm worried that one run with `--generate_references` would restore the redundant files.

gcovr/tests/test_gcovr.py
Spacetown: Copied this part from several lines above. Will change both locations.

gcovr/tests/test_gcovr.py
Spacetown: The windows specific reference files are not present anymore therefore I will remove the special reference handling for windows.

gcovr/tests/test_gcovr.py
Spacetown: You're right. I'll remove the inner loop.

gcovr/tests/test_gcovr.py
Spacetown: I'll refactor this.

Spacetown:
> Aside from the inline comments – do you think it would be possible to add a small test that ensures that there are no duplicate reference files? I'm worried that one run with `--generate_references` would restore the redundant files.

I'll add a function for cleanup the duplicate data after the test if the data was generated or updated. I've used this code temporary for removing the duplicates.

gcovr/tests/test_gcovr.py
Spacetown: Forget my previous comment, the test `exclude-throw-branches` uses a seperate reference folder for Windows.

Spacetown: On Windows both folders need to be checked. I added a inner loop to remove the code duplication.

Spacetown:
@latk Can we merge this PR before #561? Than I can use this to create the reference data for the other compilers.

latk (commented):
I think this code is fine for Unix/Linux, but I'm not sure if the handling of the Windows directories is done correctly in all cases. Could you take a look?

doc/examples/test_examples.py
```diff
@@ -44,6 +44,8 @@ def find_test_cases():
         basename = os.path.basename(script)
         name, _ = os.path.splitext(basename)
         for ext in 'txt xml csv json html'.split():
+            if ext == "html" and os.path.split(os.getenv("CC"))[1] not in ["gcc-5", "gcc-6"]:
+                continue
```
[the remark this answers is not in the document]
Spacetown: The first element is the same as [os.path.basename](https://docs.python.org/3.10/library/os.path.html?highlight=os%20path%20split#os.path.basename) I will add a function and use basename instead.

gcovr/tests/test_gcovr.py
```diff
@@ -289,6 +312,56 @@ def parse_makefile_for_available_targets(path):
...
+                os.makedirs(REFERENCE_DIRS[0], exist_ok=True)
+                print("copying %s to %s" % (generated_file, reference_file))
+                shutil.copyfile(generated_file, reference_file)
+
+
+def update_reference_data(coverage_file, reference_file):  # pragma: no cover
+    if CC_REFERENCE in reference_file:
+        reference_dir = os.path.dirname(reference_file)
+    else:
+        reference_dir = os.path.join("reference", CC_REFERENCE)
```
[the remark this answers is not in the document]
Spacetown: No, the windows directories aren't considered in the else branch, therefore the construct is needed.

latk (approved):

PR #558, opened 2022-01-19
Fix black check and prepare copyright update

Spacetown:
There was a `not` missing. The already formatted files where not checked.
Prepared copyright update to also patch runtime print in `__main__.py`.

[no changelog]

### DIFF
--- a/admin/add_copyright.py
+++ b/admin/add_copyright.py
@@ -31,6 +31,13 @@ DATE = subprocess.check_output(
 )
 YEAR = DATE[:4]
 VERSION = gcovr.version.__version__
+COPYRIGHT = [
+    f"Copyright (c) 2013-{YEAR} the gcovr authors",
+    "Copyright (c) 2013 Sandia Corporation.",
+    "This software is distributed under the BSD License.",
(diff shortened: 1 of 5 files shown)


admin/add_copyright.py
```diff
@@ -118,7 +121,21 @@ def main():
...
+                    lines = re.sub(
+                        r"COPYRIGHT = \(\n(?:[^\n]+\n)+\)\n",
+                        "\n".join(copyright_string).replace("\\", "\\\\"),
+                        "\n".join(lines),
+                    ).split("\n")
+                    lines = [line.rstrip() for line in lines]
+                newLines = handler(
+                    fullname, copy.copy(lines)
+                )  # use a copy because of the compare in the next line
                 if newLines != lines:
```
latk, line 139: Modifying `lines` means that the `if newLines != lines` check will not work. Instead, the modification should be performed in a separate function that is then selected as a `handler` above.

Spacetown: You're right. Another solution is to move this block into the `if newLines != lines:` block because if the copiright in the code only needs to be updated if the comment is updated.

PR #559, opened 2022-01-20
Do not reuse virtual environment.

Spacetown:
The venv is reused if the session OR the global setting is True. Therefore if it is activated at global level it can't be turned off local. As a result we have to set it for each session separate.

This was found while testing the session `build_wheel`, the check always used the first installed wheel.

[no changelog]

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -21,7 +21,6 @@ BLACK_CONFORM_FILES = [
 ]
 
 nox.options.sessions = ["qa"]
-nox.options.reuse_existing_virtualenvs = True
 
 
 def set_environment(session: nox.Session, cc: str, check: bool = True) -> None:
@@ -50,7 +49,7 @@ def qa(session: nox.Session) -> None:
     session.notify(session_id)
 
 
-@nox.session
+@nox.session(reuse_venv=True)
 def lint(session: nox.Session) -> None:
     """Run the lint (flake8 and black)."""
     session.install("flake8")
@@ -75,7 +74,7 @@ def lint(session: nox.Session) -> None:
         )
 
 
-@nox.session
+@nox.session(reuse_venv=True)
 def black(session: nox.Session) -> None:
     """Run black, a code formatter and format checker."""
     session.install("black")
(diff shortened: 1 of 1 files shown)


latk:
Could you explain why we would want to reuse venvs? To me, it seems like this would make dependencies more difficult to track, at negligible performance benefit.

Spacetown:
It was a performance issue. Reusing also skips the download of the data. For more reliable tests it's better to always create a new venv.

latk:
> Reusing also skips the download of the data

The `pip` installer should cache the downloaded wheels, in which case creating a new virtualenv amounts to little more than unpacking a couple of ZIP archives. It's still noticeable lag, but the dependency caching seems to work on my development system.

The Nox documentation has more discussion about [re-using virtualenvs](https://nox.thea.codes/en/stable/usage.html#opt-reuse-existing-virtualenvs):

> This is usually fine for most projects and continuous integration environments as [pip’s caching](https://pip.pypa.io/en/stable/cli/pip_install/#caching) makes re-install rather quick.

The big exception is if we were using source distributions of C extensions, so that they would have to be compiled first. But AFAIK all of our dependencies are native Python or offer wheels (like lxml).

Spacetown:
I also checked the documentation again and read the hint about pip caching. Branch is updated.

latk (approved):
Ok, I think the remaining change (*removing* default reuse) is correct.

latk:
I noticed that reusing venvs does serve a practical purpose :)

When running the tests in Docker, the pip cache is within the container's non-persistent file system, and can't be reused across runs of the container.

Notes on solutions that don't quite work: If it were just for me I'd bind-mount the host's pip cache `~/.cache/pip` into the container, but from [looking at the pip source code](https://github.com/pypa/pip/blob/21.2/src/pip/_vendor/appdirs.py#L49-L101) doing that in a platform-independent manner would be excessively complicated. Unless we just ask pip like `pip cache dir`. So a docker invocation like `-v "$(pip cache dir)":/home/docker/.cache/pip` *might* work, except for the fact that this home directory doesn't currently exist in the container (`$HOME` is `/root`, and the pip inside the container has no permissions to write to it). I'm also worried that running pip like this on the host could lead to weird WSL interactions. Not sure what a robust approach would be.

Spacetown:
I saw this behavior also yesterday.
My idea was to set the the cache directory beside the .nox directories in the gcovr folder, than the cache is saved between the runs. At the moment we can use the same directory for all containers because they all use the same python.

PR #560, opened 2022-01-22
Fix add_copyright.

Spacetown:
As mentioned in #558 there is a problem in the detection of the line change. This PR fixes the issue by adding a second handler and working with a list of handlers.

[no changelog]

### DIFF
--- a/admin/add_copyright.py
+++ b/admin/add_copyright.py
@@ -43,7 +43,7 @@ HEADER_END = (
 )
 
 
-def getLicenseSection(filename, comment_char="#"):
+def getLicenseSection(comment_char="#"):
     yield comment_char + "  ************************** Copyrights and license ***************************"
     yield comment_char
     yield comment_char + f" This file is part of gcovr {VERSION}, a parsing and reporting tool for gcov."
@@ -58,12 +58,12 @@ def getLicenseSection(filename, comment_char="#"):
     yield comment_char + HEADER_END
 
 
-def addCopyrightToPythonFile(filename, lines):
+def addCopyrightHeaderToPythonFile(filename, lines):
     # Empty file should be kept empty
     if len(lines) == 0:
         return lines
 
-    newLines = []
+    newLines = list([])
     # Keep the Shebang
     if lines[0].startswith("#!"):
         newLines.append(lines.pop(0))
@@ -76,7 +76,7 @@ def addCopyrightToPythonFile(filename, lines):
     newLines.append("")
 
     # Add license information
-    for line in getLicenseSection(filename):
+    for line in getLicenseSection():
         newLines.append(line)
 
     iterLines = iter(lines)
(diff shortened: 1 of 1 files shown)


admin/add_copyright.py
```diff
@@ -58,12 +57,12 @@ def getLicenseSection(filename, comment_char="#"):
...
 
 
-def addCopyrightToPythonFile(filename, lines):
+def addCopyrightHeaderToPythonFile(filename, lines):
     # Empty file should be kept empty
     if len(lines) == 0:
         return lines
 
-    newLines = []
+    newLines = list([])
```
latk, line 65: The expressions `[]`, `list()`, and `list([])` all produce the equivalent result: an empty list. Thus, I'd usually use a list literal `[]`, which is also the shortest.

admin/add_copyright.py
```diff
@@ -76,7 +75,7 @@ def addCopyrightToPythonFile(filename, lines):
     newLines.append("")
 
     # Add license information
-    for line in getLicenseSection(filename):
+    for line in getLicenseSection():
         newLines.append(line)
```
latk, line 79: Can be simplified as `newLines.extend(getLicenseSection())`

admin/add_copyright.py
```diff
@@ -106,36 +105,62 @@ def addCopyrightToPythonFile(filename, lines):
...
+
+    iterLines = iter(lines)
+    copyrightReached = False
+    for line in iterLines:
+        newLines.append(line)
+        if line == "COPYRIGHT = (":
+            copyrightReached = True
+            break
+    if not copyrightReached:
+        raise RuntimeError(f"Start of copyright not found in {filename}.")
```
latk, line 119: Explicit state variable is not necessary when using for–else, but this is a matter of taste:

```suggestion
    for line in iterLines:
        newLines.append(line)
        if line == "COPYRIGHT = (":
            break
    else:
        raise RuntimeError(f"Start of copyright not found in {filename}.")
```

Similar for `copyrightEndReached` below.

admin/add_copyright.py
```diff
@@ -106,36 +105,62 @@ def addCopyrightToPythonFile(filename, lines):
...
+    for line in iterLines:
+        if line == ")":
+            newLines.append(line)
+            copyrightEndReached = True
+            break
+    if not copyrightEndReached:
+        raise RuntimeError(f"End of copyright not found in {filename}.")
+
+    for line in iterLines:
+        newLines.append(line)
```
latk, line 134: Can be simplified as `newLines.extend(iterLines)`

admin/add_copyright.py
```diff
@@ -106,36 +105,62 @@ def addCopyrightToPythonFile(filename, lines):
...
     for root, dirs, files in os.walk(".", topdown=True):
-        for skip_dir in [".git", "reference"]:
-            if skip_dir in dirs:
-                dirs.remove(skip_dir)
+        for skip_dir in [
+            dir
+            for dir in dirs
+            if dir in [".git", "reference"] or dir.startswith(".nox")
+        ]:
+            dirs.remove(skip_dir)
```
latk, line 146: I don't know if the new version is clearer: a for-loop over a list comprehension seems a bit confusing.

The clearest alternative might be:

```python
def skip_dir(dir: str) -> bool:
    return dir in [".git", "reference"] or dir.startswith(".nox")

dirs[:] = [dir for dir in dirs if not skip_dir(dir)]
```

admin/add_copyright.py
```diff
@@ -106,36 +105,62 @@ def addCopyrightToPythonFile(filename, lines):
...
+            handlers = list([])
             fullname = os.path.join(root, filename)
             if filename.endswith(".py"):
-                handler = addCopyrightToPythonFile
+                handlers.append(addCopyrightHeaderToPythonFile)
+            if filename == "__main__.py":
+                handlers.append(updateCopyrightString)
 
-            if handler is not None:
+            if len(handlers) != 0:
```
latk, line 156: Idiomatic would be `if handlers:`

latk (approved):
Looks good! I only have some comments about Python idioms, but that is subjective so you can ignore them if you want to.

admin/add_copyright.py
Spacetown: Ok. I thought for typing it's better to write the long form. I'll change it.

admin/add_copyright.py
Spacetown: That construct is new for me. I'll change it.

admin/add_copyright.py
Spacetown: Ok. The problem was that I don't want to add a second loop. I'll change it.

Spacetown: Correct me if I'm wrong. The difference between `dir[:] = [...]` and `dir = [...]` is that the first change the elements in the list and the second one creates a new list.

latk: That is correct. Using a slice on the left hand side updates the existing `dirs` list. We must modify the existing `dirs` list object due to the way `os.walk()` works. Compare also our use of `os.walk()` in the [`gcovr.utils.search_file()` function](https://github.com/gcovr/gcovr/blob/master/gcovr/utils.py#L74-L82).

PR #561, opened 2022-01-24
Add option to configure exclude marker

### LINKED ISSUE #557: Support overrides to the default EXCL patterns (LINE/START/STOP) to allow builds to exclude different sets of code as needed
I work on a project that has a need to exclude certain types of coverage that is not conditionally compiled for various reasons. We cannot functionally increase coverage across build types to match expectations because we will always have sections of code that never expect to execute for certain clients. A way to resolve this would be to allow us to change the patterns used when marking code as excluded (via LINE/START/STOP markers).

An easy solution is to just add a configuration option to override this default regex as necessary. I already have a fork that does this.

Obviously the alternative would be to conditionally compile to prevent this code being in the executable to begin with. This is not always possible (mainly due to constraints with agreed-upon interfaces / features for the builds with code we need to exclude).

**Additional context**
Primarily, this hits hardest, for our usages, in places where the interface is expected to remain the same but naturally upgrade / progress when clients move to newer hardware using our codebase.

The branch I am using for this: https://github.com/zphensley42/gcovr/tree/task/flexible_exclusions

I will make a PR into mainline if acceptable as an option.

zphensley42:
As discussed in: https://github.com/gcovr/gcovr/issues/557

This PR adds the ability to provide a prefix to existing EXCL_LINE/START/STOP markers to allow for projects to use differing markers as necessary among builds that do not conditionally exclude executable code. The existing pattern matching option does not allow for START/STOP markers to be modified (i.e. only provides for line marking) and thus this extends that with a new option.

Closes #557


gcovr/configuration.py
```diff
@@ -1027,6 +1027,11 @@ def merge_options_and_set_defaults(partial_namespaces, all_options=None):
         type=int,
         default=1,
     ),
+    GcovrConfigOption(
+        "exclude_pattern_prefix", ["--exclude-pattern-prefix"],
+        help="Define the regex prefix used in markers / line exclusions (i.e _EXCL_START, _EXCL_START, _EXCL_STOP)",
+        type=str
+    ),
```
Spacetown, line 1034: Set the default prefix in the the argument parser.
```suggestion
    GcovrConfigOption(
        "exclude_pattern_prefix", ["--exclude-pattern-prefix"],
        help="Define the regex prefix used in markers / line exclusions (i.e _EXCL_START, _EXCL_START, _EXCL_STOP)",
        type=str,
        default=r"[GL]COVR?",
    ),
```

gcovr/gcov_parser.py
```diff
@@ -332,6 +334,7 @@ def parse_coverage(
             lines=src_lines,
             warnings=_ExclusionRangeWarnings(logger, filename),
             exclude_lines_by_pattern=exclude_lines_by_pattern,
+            exclude_pattern_prefix=exclude_pattern_prefix if exclude_pattern_prefix is not None else r"[GL]COVR?"
```
Spacetown, line 337: Always use the argument.
```suggestion
            exclude_pattern_prefix=exclude_pattern_prefix,
```

Spacetown (changes requested):
The default for the argument shall be handled by the command line parser.
Please also add the issue to the changlog.rst.

gcovr/configuration.py
zphensley42: Pushed.

gcovr/gcov_parser.py
zphensley42: Pushed.

CHANGELOG.rst
```diff
@@ -23,6 +23,7 @@ Improvements and new features:
 - Add :option:`--no-markers` to ignore exclusion markers in code. (:issue:`361`)
 - Generate also a Report of Excluded Coverage. (:issue:`503`)
 - Added :option:`--decisions` to add decision coverage to HTML and JSON output. (:issue:`350`)
+- Added :option:`--exclude-pattern-prefix` to allow projects to define custom prefixes used in _EXCL_LINE/START/STOP markers
```
Spacetown, line 26:
```suggestion
- Added :option:`--exclude-pattern-prefix` to allow projects to define custom prefixes used in _EXCL_LINE/START/STOP markers. (:issue:`561`)
```

Spacetown (approved):
LGFM.
@latk Shall we merge this and you extend the documentation or shall we first merge the documentation and update it with this PR?

Spacetown:
Please can you fix the formatting errors?

zphensley42:
On it (--nox lint not catching it locally, odd).

Spacetown:
There was a error in the noxfile. Please Reba's to current master, there it's fixed.

gcovr/gcov_parser.py
```diff
@@ -53,7 +53,7 @@
 from .utils import Logger
 
 _EXCLUDE_LINE_FLAG = "_EXCL_"
-_EXCLUDE_LINE_PATTERN = re.compile(r"([GL]COVR?)_EXCL_(LINE|START|STOP)")
+_EXCLUDE_LINE_PATTERN = r"_EXCL_(LINE|START|STOP)"
```
Spacetown, line 56: Can we add a `_POSTFIX` to the variable name?

zphensley42:
Do I need to push something for the references?

gcovr/gcov_parser.py
zphensley42: Sure thing. Will update in the morning (for me).

kollix: Since I'm currently looking into gcovr as a replacement for the MS tool we are currently using, I would rather much like the possibility to configure not only the PREFIX but the WHOLE regexp.
E.g. our huge codebase (hundreds of thousands of lines)
has already code marked with exclusion markers and they look like this:
start of an excluded block: COVINFO BLOCK
end of a block: COVINFO BLOCKEND
exclude a line: COVINFO LINE
There is also a way to exclude a whole file, where the comment in the first line is: COVINFO FILE

Would be great if I would not need to change that in all files!

zphensley42: If @Spacetown and others agree to that, I can update my PR to support. Otherwise, I will continue with the prefix path.

Spacetown: @kollix At the moment I don't want to support this because we need three options to support this. The current used postfixes are also used by gcov and lcov. From which tool are your markers?

@zphensley42 Please continue with the prefix path.

zphensley42: Will do.

Spacetown (changes requested):
The reference data is only present for `gcc-5`. Do you have the the other compiler versions or docker installed? If not, I can create the other reference data for you.

zphensley42:
Ah, I see. I don't have them on this machine, so I kindly request them to be created elsewhere :)

kollix:
On Dienstag, 25. Jänner 2022 20:09:04 CET Michael Förderer wrote:

> @kollix At the moment I don't want to support this because we need three options to support this.

ok, if not this moment, is there a better moment then ?

> The current used postfixes are also used by gcov and lcov.

which does not exclude my feature wish and sadly does not help my use case.

> From which tool are your markers?

as said, MS coverage (our code is cross platform and up till now we do coverage only
on windows, but the tool is slow, so I'm looking for a better alternative but want
to avoid the need to change thousands of existing markers)

-- 
Best regards/Schöne Grüße

Martin
A: Because it breaks the logical sequence of discussion
Q: Why is top posting bad?

()  ascii ribbon campaign - against html e-mail 
/\                        - against proprietary attachments

Frühstück, Geschenkideen, Accessoires, Kulinarisches: www.lillehus.at

Spacetown:
> as said, MS coverage

Do you have a link to that? I find no documentation with this markers.

The exclusion of a single line is possible with `--exclude-line-pattern`. I've a problem with adding several options to support migration from another tool. One solution can be to use the JSON coverage files and filter them by scanning the source files. The filtered files can than be used to generate a report.

kollix:
On Mittwoch, 26. Jänner 2022 08:27:52 CET Michael Förderer wrote:
> > as said, MS coverage
> 
> Do you have a link to that? I find no documentation with this markers.
> 
> The exclusion of a single line is possible with `--exclude-line-pattern`. I've a problem with adding several options to support migration from another tool. One solution can be to use the JSON coverage files and filter them by scanning the source files. The filtered files can than be used to generate a report.

Let's forget about my wish. I now learned that the mentioned COVINFO markers are post-processed
by an internal tool, not by MS coverage itself, so we can adjust that tool on our own if needed.
Thanks.

-- 
Best regards/Schöne Grüße

Martin
A: Because it breaks the logical sequence of discussion
Q: Why is top posting bad?

()  ascii ribbon campaign - against html e-mail 
/\                        - against proprietary attachments

Frühstück, Geschenkideen, Accessoires, Kulinarisches: www.lillehus.at

Spacetown (approved):
LGFM, I'll update the reference later.
@latk Do you want to finish the documentation change first or will you update your branch with this change?

Spacetown (approved):
LGFM

PR #564, opened 2022-01-30
Pin version of black.

Spacetown:
Adapt code to black 22.1.0 and pin the version.

[no changelog]


latk (approved):
Pinning specific versions of these tools is good. That was an easy to review PR :)

PR #565, opened 2022-01-31
Don't use realpath in data model, only use them for the filter

Spacetown:
With this PR the paths in the data model aren't changed to `realpath` anymore. Only in the filter or the HTML output the paths are changed to realpaths. This should workaround #365.

The test results aren't affected by this change.


gcovr/gcov.py
```diff
@@ -183,14 +183,14 @@ def guess_source_file_name(
 
 def guess_source_file_name_via_aliases(gcovname, currdir, data_fname):
     common_dir = commonpath([data_fname, currdir])
-    fname = realpath(os.path.join(common_dir, gcovname))
+    fname = os.path.abspath(os.path.join(common_dir, gcovname))
```
latk, line 186: (Remark) This change looks like a significant deviation from the previous semantics, which might cause some previous workflows to break. But from my perspective it's 50/50 whether this breaks stuff or fixes long-standing problems, so I'm OK with this change.

gcovr/tests/test_gcovr.py
```diff
@@ -57,6 +57,17 @@
 CC = os.path.split(env["CC"])[1]
 IS_CLANG = True if CC.startswith("clang") else False
 
+IS_WINDOWS = platform.system() == "Windows"
+if IS_WINDOWS:
+    import win32api
+    import string
+
+    used_drives = win32api.GetLogicalDriveStrings().split("\0")
+    free_drives = list(set([c for c in string.ascii_uppercase]) - set(used_drives))
```
latk, line 66: (Style) The `set()` constructor should accept any iterable, not just a list. This should be equivalent:

```suggestion
    free_drives = list(set(string.ascii_uppercase) - set(used_drives))
```

gcovr/tests/test_gcovr.py
```diff
@@ -57,6 +57,17 @@
...
+    if len(free_drives) == 0:
+        raise RuntimeError("No free drive letter found for test.")
```
latk, line 68: (Style) Looks like a candidate for a `# pragma: no cover` annotation, or using an assert instead:

```suggestion
    assert free_drives, "Must have at least one free drive letter"
```

gcovr/tests/test_gcovr.py
```diff
@@ -249,21 +258,25 @@ def pytest_generate_tests(metafunc):
...
-                pytest.mark.xfail(
+                pytest.mark.skipif(
                     name == "html-source-encoding-cp1252" and IS_CLANG,
                     reason="clang doesnt understand -finput-charset=...",
                 ),
-                pytest.mark.xfail(
+                pytest.mark.skipif(
                     name == "gcc-abspath"
                     and (
                         not env["CC"].startswith("gcc-")
```
latk, line 282: These source-encoding test were xfail so that they would still run under Clang, in case future Clang versions add support for this command line option. But since the test suite pins certain Clang versions (and since this relates to an extremely niche feature), skipping that test entirely is fine.

On the other hand, the “gcc-abspath” test relates to the extremely useful `-fprofile-abs-path` option. It would be good to keep attempting this test, so that we are alerted if the option becomes available with Clang as well.

gcovr/utils.py
```diff
@@ -47,7 +47,7 @@ def already_visited(self, path):
     DOS_DEVICE_PATH_PREFIX = "\\\\?\\"
     DOS_DEVICE_PATH_PREFIX_UNC = DOS_DEVICE_PATH_PREFIX + "UNC\\"
 
-    def realpath(path):
+    def _realpath(path):
```
latk, line 50: Why was this function renamed? A leading underscore typically indicates that the variable would be private, but we do want to import this function e.g. in the HTML writer.

gcovr/utils.py
```diff
@@ -234,8 +234,12 @@ def __init__(self, pattern):
         flags = re.IGNORECASE if is_fs_case_insensitive else 0
         self.pattern = re.compile(pattern, flags)
 
+    @classmethod
+    def _os_independent_path(cls, path):
+        return path.replace(os.path.sep, '/')
```
latk, line 239: (Style) This can probably be a free function outside of a class.

latk (approved):
I think this can be merged and would likely improve the state of things :)

Nevertheless, some comments. The Remark is just me thinking out loud. The ones marked Style are entirely optional.

gcovr/utils.py
Spacetown: Nowhere in the code realpath shall be used anymore except in the filter therefor it's now private to the module. Maybe it's even better to add it as private method to the filter.

Spacetown: Ok, you're right with the HTML writer. I'll revert the renaming.

Spacetown:
The test for Ubuntu-20.04 should be ignored at the moment, see actions/virtual-environments#5237.

PR #570, opened 2022-03-16
Change file level percent values in JSON summary report

### LINKED ISSUE #517: Percentage values in json-summary report
**Describe the bug**
In the json-summary report, the file-summary uses ratios in the attributes 'line_percent' and 'branch_percent'.
The value for full coverage is 1.0

**To Reproduce**
1. Generate a coverage report using the --json-summary-pretty reporter with any coverages.

**Expected behavior**
The value for full coverage should be 100.0 (i.e. a percentage value)

**Example output**
`
{
    "branch_covered": 2,
    "branch_percent": 100.0,
    "branch_total": 2,
    "files": [
        {
            "branch_covered": 2,
            "branch_percent": 1.0,
            "branch_total": 2,
            "filename": "impl/my_library/src/my_base_class.cpp",
            "line_covered": 6,
            "line_percent": 1.0,
            "line_total": 6
        }
    ],
    "gcovr/summary_format_version": "0.2",
    "line_covered": 6,
    "line_percent": 100.0,
    "line_total": 6,
    "root": "../../../../test"
}
`

Spacetown:
On file level the range was from 0.0 to 1.0 and on summary level from 0.0 to 100.0. Now it's always from 0.0 to 100.0.

Closes #517


gcovr/tests/csv/coverage.csv
```diff
@@ -1,2 +1 @@
 filename,line_total,line_covered,line_percent,branch_total,branch_covered,branch_percent,function_total,function_covered,function_percent
-main.cpp,2,2,1.0,0,0,,1,1,1.0
```
latk, line 2: unexplained change

latk (commented):
This looks good, except that the CSV test now seems to have changed? Do you have any idea what is going on there??

gcovr/tests/csv/coverage.csv
Spacetown: That file was by accident under SCM. It's the output file of the run but the rm command in the makefile was commented. The reference didn't change.
I've removed the file and the comment in the makefile.

Spacetown: Should we add a check if no file from SCM is changed by a test (`git status`)?

latk: I think no further checks are needed since the .gitignore file already excludes coverage.csv files by default.

PR #582, opened 2022-03-26
Rename add copyright

Spacetown:
Rename the script to bump_version.
Update also the version in deploy.yml which is used for release_checklist and add an additional line break in front of a class and a function since this is required by black. 

[no changelog]


admin/bump_version.py
```diff
@@ -124,13 +129,30 @@ def updateCopyrightString(filename, lines):
...
     return newLines
 
 
+def updateCallOfReleaseChecklist(filename: str, lines: List[str]):
+    newLines = []
+
+    callReleaseChecklist = "admin/release_checklist"
+    callFound = False
+    iterLines = iter(lines)
+    for line in iterLines:
```
latk, line 145: Explicit iter() call seems unnecessary here since we don't want to suspend/resume or manually advance it here.

```suggestion
    for line in lines:
```

latk (approved):
good change!

Spacetown:
This was a catch from updating the version yesterday but I thought a separate PR is better.

PR #584, opened 2022-03-27
Update references to NEXT version with current version in documentation.

Spacetown:
Bump version also to documentation files:
- The references used in the added, removed and deprecated parts.
- The headline in the CHANGELOG.rst.

[no changelog]

### DIFF
--- a/admin/bump_version.py
+++ b/admin/bump_version.py
@@ -23,6 +23,7 @@ import os
 import logging
 import re
 import subprocess
+import time
 from typing import List
 
 import gcovr.version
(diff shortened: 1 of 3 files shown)


admin/bump_version.py
```diff
@@ -152,6 +153,37 @@ def updateCallOfReleaseChecklist(filename: str, lines: List[str]):
...
+
+    nextLine = None
+    for line in lines:
+        if line == "Next Release":
+            line = f"{VERSION} ({time.strftime('%d %B %Y')})"
+            nextLine = "-" * len(line)
+        elif nextLine:
+            line = nextLine
+            nextLine = None
+        newLines.append(line)
```
latk, line 167: This part feels somehow complicated. If we can trust that the `----` line always exist (and the current code does trust this because it always overwrites the next line), then it could be simplified to:

```suggestion
    for line in lines:
        if line == "Next Release":
            newLines.append(f"{VERSION} ({time.strftime('%d %B %Y')})")
        else:
            newLines.append(line)
```

latk (approved):
I love this kind of automation!

admin/bump_version.py
Spacetown: The problem ist that the length of the string must be the same. If not, you'll get a RST error.

latk: Ok, that makes total sense (your code, not the RST syntax…).

latk:
Test failure is due to https://github.com/psf/black/issues/2964

If you want, can be fixed by pinning `click==8.0.4` in the Noxfile

```diff
-BLACK_PINNED_VERSION = "black==22.1.0"
+BLACK_PINNED_VERSION = ["black==22.1.0", "click==8.04"]
```

and then updating all occurrences of that constant by `*BLACK_PINNED_VERSION`

Issue #587, opened 2022-03-28
`FileNotFoundError` after update to 5.1


sergiud:
**Describe the bug**
After updating from gcovr 5.0 to 5.1, collecting coverage using same setup causes the following error

```console
$ cd build/
$ gcovr -r .. -s --html -o ${PROJECT_DIR}/coverage/report.html --xml-pretty --xml ${PROJECT_DIR}/cobertura.xml
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/gcovr/workers.py", line 77, in worker
    work(*args, **kwargs)
  File "/usr/lib/python3.10/site-packages/gcovr/gcov.py", line 317, in process_datafile
    done = run_gcov_and_process_files(
  File "/usr/lib/python3.10/site-packages/gcovr/gcov.py", line 490, in run_gcov_and_process_files
    process_gcov_data(fname, covdata, abs_filename, options)
  File "/usr/lib/python3.10/site-packages/gcovr/gcov.py", line 93, in process_gcov_data
    with io.open(
FileNotFoundError: [Errno 2] No such file or directory: '/builds/sergiu/tracker/stl_pair.h##47d0af436808dced1f869194db0c00bc.gcov'
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/gcovr/workers.py", line 77, in worker
    work(*args, **kwargs)
  File "/usr/lib/python3.10/site-packages/gcovr/gcov.py", line 317, in process_datafile
    done = run_gcov_and_process_files(
  File "/usr/lib/python3.10/site-packages/gcovr/gcov.py", line 490, in run_gcov_and_process_files
    process_gcov_data(fname, covdata, abs_filename, options)
  File "/usr/lib/python3.10/site-packages/gcovr/gcov.py", line 93, in process_gcov_data
    with io.open(
FileNotFoundError: [Errno 2] No such file or directory: '/builds/sergiu/tracker/stl_pair.h##47d0af436808dced1f869194db0c00bc.gcov'
(ERROR) Uncaught EXCEPTION
Traceback (most recent call last):
  File "/usr/sbin/gcovr", line 33, in <module>
    sys.exit(load_entry_point('gcovr==5.1', 'console_scripts', 'gcovr')())
  File "/usr/lib/python3.10/site-packages/gcovr/__main__.py", line 308, in main
    collect_coverage_from_gcov(covdata, options)
[...]

PR #589, opened 2022-03-28
Add exception handling for processing of files.

Spacetown:
Add exception handler to get more infos on gcov failures as mentioined in #587.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,8 @@ Breaking changes:
 
 New features and notable changes:
 
+- Log additional infop on gcov parsing errors (:issue:`589`)
+
 Bug fixes and small improvements:
 
 Documentation:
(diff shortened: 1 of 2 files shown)


latk (approved):
Once the tests pass I'm cool with this, and merging it would allow people having issues to install the gcovr development version to benefit from this additional context.

If time/motivation permits, it might be worth thinking about whether this information can be presented via `logger.error()` or something, in order to avoid those excessively long tracebacks Python throws at end users. On the other hand, preserving those tracebacks is valuable for debugging.

Spacetown:
> If time/motivation permits, it might be worth thinking about whether this information can be presented via `logger.error()` or something, in order to avoid those excessively long tracebacks Python throws at end users. On the other hand, preserving those tracebacks is valuable for debugging.

This was my first approach. 😄 Changed back to use `logger.error` and re-raise the exception.

PR #594, opened 2022-03-29
Fix black check: Option --check needed.

Spacetown:
The option was removed by accident when reformatting all files and apply check to all files. Session isn't failing on error.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,12 +10,13 @@ Breaking changes:
 
 New features and notable changes:
 
-- Log additional infop on gcov parsing errors (:issue:`589`)
+- Log additional infop on gcov parsing errors. (:issue:`589`)
 
 Bug fixes and small improvements:
 
-- Remove function coverage from sonarcube report (:issue:`591`)
-- Fix parallel processing of gcov data (:issue:`592`)
+- Remove function coverage from sonarcube report. (:issue:`591`)
+- Fix parallel processing of gcov data. (:issue:`592`)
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -97,7 +97,9 @@ def black(session: nox.Session) -> None:
     if session.posargs:
         session.run("python", "-m", "black", *session.posargs)
     else:
-        session.run("python", "-m", "black", "--diff", *DEFAULT_LINT_ARGUMENTS)
+        session.run(
```
latk, line 100: Is the `black` session supposed to *reformat* files, or to *check* for correct formatting? Depending on whether session arguments are provided this will do different things. Instead of adding `--check`, removing `--diff` might be a better solution.

CHANGELOG.rst
```diff
@@ -10,12 +10,13 @@ Breaking changes:
...
-- Log additional infop on gcov parsing errors (:issue:`589`)
+- Log additional infop on gcov parsing errors. (:issue:`589`)
 
 Bug fixes and small improvements:
 
-- Remove function coverage from sonarcube report (:issue:`591`)
-- Fix parallel processing of gcov data (:issue:`592`)
+- Remove function coverage from sonarcube report. (:issue:`591`)
+- Fix parallel processing of gcov data. (:issue:`592`)
+- Fix black session to fail on format errors. (:issue:`594`)
```
latk, line 19: This is an internal change, so the changelog entry should be moved to the other section.

latk (approved):

Issue #596, opened 2022-03-30
Parsing gcov file fails on "Source is newer than graph" line


ovianao:
**Describe the bug**
I can't provide much more detail, since it's an internal / private project, but I've nailed it down to the stack trace below and the following content in the gcov file along with gcovr 5.1, as we hit this line:

https://github.com/gcovr/gcovr/blob/e71e883521b78122c49016eb4e510e6da06c6916/gcovr/gcov_parser.py#L724

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/gcovr/workers.py", line 77, in worker
    work(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/gcovr/gcov.py", line 545, in process_existing_gcov_file
    process_gcov_data(filename, covdata, None, options)
  File "/usr/local/lib/python3.8/dist-packages/gcovr/gcov.py", line 100, in process_gcov_data
    metadata = parse_metadata(lines)
  File "/usr/local/lib/python3.8/dist-packages/gcovr/gcov_parser.py", line 226, in parse_metadata
    parsed_line = _parse_line(line)
  File "/usr/local/lib/python3.8/dist-packages/gcovr/gcov_parser.py", line 724, in _parse_line
    key, value = source_code.split(":", 1)
ValueError: not enough values to unpack (expected 2, got 1)

**To Reproduce**
1. ... Have a gcov file starting with these lines:
-:    0:Source:some_path_to_a_header_file.h  (in my case, gtest.h)
-:    0:Programs:3
-:    0:Source is newer than graph

**Expected behavior**
gcovr should parse the 3rd line.  Instead, it does not since it's expecting "Source is newer than graph" to be split-able by a ":".
Moving back to gcovr 5.0 allows us to go beyond the issue, since this change appears to have introduced it:
https://github.com/gcovr/gcovr/commit/304518453d352b21852e94f79ff78a4f4e4d2cb9

**Screenshots**

**Desktop (please complete the following information):**
    - Ubuntu 20.04
 - GCC version [e.g. MinGW 8.1.0]
   - gcov-5
 - GCOVR version [e.g. 4.2]
   - 5.1
 - Project directory layout:
   - private

**Additional context**

latk:
Confirmed. Thanks for reporting this! Looking at the `gcov` source code (gcc 11), this message is the *only* case where the `-: 0:header:content` format is violated. It seems I had missed this case when writing the new parser.

[...]

PR #597, opened 2022-03-30
Add tests for gcc10 and gcc11

Spacetown:
For testing the fix of #596 we first need to introduce the gcc-10 and gcc-11.

- On my computer the tests `test_multiple_output_formats_to_stdout` and `test_multiple_output_formats_to_stdout_1` aborted with a timeout because all directories where searched for data files (venvs of nox). Therefore the option `--root gcovr` was added to reduce the number of directories.
- For gcc-11 our tests detected an error. The gcda files are created in the sub directory test and the source file wasn't found anymore.


CONTRIBUTING.rst
```diff
@@ -155,11 +155,18 @@ on **Windows** a Pro license or the WSL (Windows subsystem for Linux)
...
+    the PATH environment. If on of the following command groups are
+    everything is OK.
+    -  gcc-5/g++-5/gcov-5
+    -  gcc-6/g++-6/gcov-6
+    -  gcc-8/g++-8/gcov-8
+    -  gcc-9/g++-9/gcov-9
+    -  gcc-10/g++-10/gcov-10
+    -  gcc-11/g++-11/gcov-11
+    -  clang-10/clang++-10/llvm-cov
+    -  clang-13/clang++-13/llvm-cov
```
latk, line 167: I think RST requires a blank line before/after the nested list

gcovr/gcov.py
```diff
@@ -231,15 +231,20 @@ def guess_source_file_name_heuristics(
         if os.path.exists(fname):
             return fname
 
+    # 4. Try using the path to the gcov file
+    fname = os.path.join(os.path.dirname(data_fname), gcovname)
+    if os.path.exists(fname):
+        return fname
+
```
latk, line 238: I'm quite afraid of touching any code in this part of gcovr – whether I like the heuristics or not, any change could break backwards compatibility. Unless there's a good reason, new heuristics should therefore be added at the end.

Could you explain a bit more about why this heuristic is appropriate and correct? As far as I understand, the `data_fname` is the `.gcov` file, which will be created in whatever directory that `gcov` was invoked in as per the `potential_wd` heuristics in `process_datafile()`.

For which specific test cases does the current behavior become a problem?

Spacetown: I can move it after the g da part. The cause is written in the commit comment.
```
gcc-11 creates in the gcda files in a sub directory which results in an error later. The file is used as testApp/test/tmp.cpp but correct is testApp/tmp.cpp.

(DEBUG) Running gcov: '/usr/bin/gcov-11 /gcovr/gcovr/tests/shared_lib/testApp/test/a-tmp.gcda --branch-counts --branch-probabilities --demangled-names --hash-filenames --object-directory /gcovr/gcovr/tests/shared_lib/testApp/test' in '/gcovr/gcovr/tests/shared_lib/testApp'
(DEBUG) Finding source file corresponding to a gcov data file
  currdir      /gcovr/gcovr/tests/shared_lib
  gcov_fname   /gcovr/gcovr/tests/shared_lib/testApp/tmp.cpp##95fbba3ae32e8157841b28787cf829e3.gcov
  source_fname /gcovr/gcovr/tests/shared_lib/testApp/test/a-tmp.gcda
  root         /gcovr/gcovr/tests/shared_lib
  fname        /gcovr/gcovr/tests/shared_lib/testApp/test/tmp.cpp
(DEBUG) Parsing coverage data for file /gcovr/gcovr/tests/shared_lib/testApp/test/tmp.cpp
```

Previous versions created the gcda file directly in the test app folder.

latk:
Regarding the new file name heuristic:

Ok, the location of output files is indeed a change in GCC 11, see https://github.com/gcc-mirror/gcc/commit/1dedc12d186a110854537e1279b4e6c29f2df35a#diff-9e31bd2d2496420aa34b8fc99ff6d1bf30f606032fe68eabbca9eac3c3189e69R1410-R1427

Personally, I think the new behaviour is *far* more logical.

My next thought was that this would finally be the reason to implement support for

```none
        -:    0:Working directory:/home/gcc/testcase
```

lines, only to realize that this was never actually implemented in ANY gcc version!!! Especially not GCC-9 where it [appeared in the docs](https://gcc.gnu.org/onlinedocs/gcc-9.1.0/gcc/Invoking-Gcov.html#Invoking-Gcov). That info is only provided in JSON reports…

So, I think the new heuristic you suggest is not only needed, it's also going to be **almost always correct!** Let's look into the `output_lines()` function in gcov.c (here, version 11.2):

```c++
static void
output_lines (FILE *gcov_file, const source_info *src)
{
#define  DEFAULT_LINE_START "        -:    0:"
#define FN_SEPARATOR "------------------\n"

  FILE *source_file;

  ...

  fprintf (gcov_file, DEFAULT_LINE_START "Source:%s\n", src->coverage.name);
  ...

  source_file = fopen (src->name, "r");
  if (!source_file)
    fnotice (stderr, "Cannot open source file %s\n", src->name);
```

Here, the header contains `src->coverage.name` whereas the file being opened is `src->name`. These will typically be the same if they were assigned in `find_source()`:

```c++
      sources.push_back (source_info ());
      src = &sources.back ();
      src->name = canon;
      src->coverage.name = src->name;
      src->index = idx;
      if (source_length
#if HAVE_DOS_BASED_FILE_SYSTEM
	  /* You lose if separators don't match exactly in the
	     prefix.  */
	  && !strncasecmp (source_prefix, src->coverage.name, source_length)
#else
	  && !strncmp (source_prefix, src->coverage.name, source_length)
#endif
	  && IS_DIR_SEPARATOR (src->coverage.name[source_length]))
	src->coverage.name += source_length + 1;
...
```

where the removed source_prefix can be set by `gcov -s PREFIX` which we don't use, so the two fields will contain the same value.

I'm not entirely sure whether any of these fields could be set in another place, but I don't see anything.

And how is `canon` created? It is just the canonicalized path taken directly from the .gcno file, which will be a relative path from the gcc working directory. (← I haven't verified how GCC mangles these files).

As far as I understand, the .gcov file is always generated in the gcov working directory.

This would mean:

* the .gcno contains a relative path from compiler working directory to source file
* gcov tries to read this file
* this will only succeed if compiler working directory = gcov working directory
* the .gcov is written to the gcov working directory
* thus, unless `gcov -s` is used, the reported source file is always relative to the .gcov file's directory
* which means your new heuristic is probably always correct

Open questions for me: does the same hold for all other GCC versions and for Clang?

I have run some experiments with replacing the entire `guess_source_file_name_heuristics()` function with the new approach and it *almost* works across our entire test suite (some problem with `subfolder-includes-html` that I don't understand, is probably unrelated).

Given this background, I would be tempted to check the new heuristic first, and fall back to the old heuristics otherwise, possibly with a warning message.

Spacetown:
> I have run some experiments with replacing the entire `guess_source_file_name_heuristics()` function with the new approach and it _almost_ works across our entire test suite (some problem with `subfolder-includes-html` that I don't understand, is probably unrelated).

I've updated this test. The expected exitcode 7 (Error from writer) wasn't set. The expectation was that the source file isn't found but no it is.

gcovr/gcov.py
```diff
@@ -203,29 +203,36 @@ def guess_source_file_name_via_aliases(gcovname, currdir, data_fname):
...
     # separator is different, so we replace it with the correct separator
     gcovname = gcovname.replace("/", os.sep)
 
-    # 0. Try using the current working directory as the source directory
+    # 0. Try using the path to the gcov file
+    fname = os.path.join(os.path.dirname(data_fname), gcovname)
+    if os.path.exists(fname):
+        return fname
+
+    logger.debug("Fallback to heuristic of gcovr 5.1")
```
Spacetown, line 218: It seems, that the heuristic isn't used anymore in our tests.

latk:
I'll have to look at the changed test very carefully to understand what's going on and whether this change is safe – it strikes me as odd that we might create that kind of test to ensure that a file *isn't* found.

Even though the new source file finding logic seems to make the previous heuristics obsolete, I'd prefer to keep them around in case something goes wrong. If there are no issues in 5.2, I think it would be a good idea to make them trigger a deprecation warning in case they are used. Then, if there are no reports about the warning, it can be safely removed at a later date.

For cases like these, I really miss having (anonymous) usage analytics to gain insight into which features are actually being used (I'm also surprised by the large sample of people who seem to activate the `-j` option…). But to clarify: I have no intention of adding any analytics or usage reports.

Spacetown:
> it strikes me as odd that we might create that kind of test to ensure that a file _isn't_ found.

The HTML details can't be generated in the past. This was extended with #358 which introduces the exit code 7 if there was a problem from the output writer. With this PR the test was introduced.

latk:
Ok, I think it's good that this test had to be changed.

But this means now there's no test for the HTML resilience feature. I think it's no longer possible to test this feature when using gcov, but it would be possible to create a test with JSON input, for example:

```json
{
  "gcovr/format_version": "...",
  "files": [{
    "file": "does-not-exist.cpp",
     "lines": []
  }]
}
```

Spacetown:
Test is added. For this test and possible others the clang-10 data is the same as the data of gcc-5. Should I add the gcc-5 data also to the list where the data of clang is searched?

latk (approved):

latk:
> Should I add the gcc-5 data also to the list where the data of clang is searched?

All compilers should see the same data for that test because the test doesn't depend on the compiler in any way. I don't think combining the gcc and clang reference inheritance hierarchies is a good idea, but maybe it would make sense to think about adding a `generic` reference directory for these cases. Are there other tests with a similar situation?

Spacetown:
I'll check this.

PR #601, opened 2022-04-03
Fix parsing of gcov metadata

### LINKED ISSUE #596: Parsing gcov file fails on "Source is newer than graph" line

Spacetown:
Gcc-11 has metadata lines without values in the gcov file (see #596). This PR add support for this lines and uses as value `None`.

Closes #596.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,7 @@ Bug fixes and small improvements:
 - Remove function coverage from sonarcube report. (:issue:`591`)
 - Fix parallel processing of gcov data. (:issue:`592`)
 - Better diagnostics when dealing with corrupted input files. (:issue:`593`)
+- Accept metadata lines without values (introduced in gcc-11). (:issue:`601`)
 
 Documentation:
 
--- a/gcovr/gcov.py
+++ b/gcovr/gcov.py
@@ -99,8 +99,11 @@ def process_gcov_data(data_fname, covdata, gcda_fname, options, currdir=None):
     # Find the source file
     # TODO: instead of heuristics, use "working directory" if available
     metadata = parse_metadata(lines)
+    source = metadata.get("Source")
+    if source is None:
+        raise RuntimeError("Unexpected value 'None' for metadata 'Source'.")
     fname = guess_source_file_name(
-        metadata["Source"].strip(),
+        source,
         data_fname,
         gcda_fname,
         root_dir=options.root_dir,
--- a/gcovr/gcov_parser.py
+++ b/gcovr/gcov_parser.py
@@ -220,11 +220,7 @@ def parse_metadata(lines: List[str]) -> Dict[str, str]:
     RuntimeError: Missing key 'Source' in metadata. GCOV data was >>
       -: 0:Foo:bar
       -: 0:Key:123<< End of GCOV data
-    >>> parse_metadata('''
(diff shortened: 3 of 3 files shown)


latk (approved):
simple and elegant :)

Using a None value does kind of violate the type annotations around that code, but for now this won't break anything and that can be fixed in the next refactoring push.

(*If* you want to fix this, note that now the metadata values are not `str` but `Optional[str]`. This also means the `metadata["Source"]` access in process_gcov_data() could be None, as far as the type system is concerned. The most elegant approach to solve this would be to move the “Source must exist” check from the metadata-parser into process-gcov-data, to something like `source = metadata.get("Source"); if source is None: raise RuntimeError(...)`)

Spacetown:
@latk I've moved the `strip` to the parsing of the metadata. How can I force a leading whitespace in the doctest if the editor removes them.

latk:
> How can I force a leading whitespace in the doctest if the editor removes them.

I don't entirely understand. Could you show an example of what you want to do?

Spacetown:
I wan't to add leading whitespaces to the following doctest:
```
    >>> parse_metadata('''
    ...   -: 0:Foo:bar
    ...   -: 0:Key:123
    ... '''.splitlines())
```
The whitepaces shouldn't be removed by the editor on save.

One solution is to write a single line string with ` \n`.

PR #603, opened 2022-04-04
Minor cosmetic improvements to HTML legend

Pesa:
* Use `≥` sign instead of `>=`
* ~~Define low coverage as "below medium" rather than ">= 0%"~~


gcovr/templates/functions_page.html
```diff
@@ -15,12 +15,12 @@
     <tr>
       <th scope="row">Legend:</th>
       <td class="legend">
-        <span class="coverage-low">low: &gt;= 0%</span>
+        <span class="coverage-low">low: &lt; {{COVERAGE_MED}}%</span>
```
Spacetown, line 18: Please don't change the direction. In the past it was this way but it's clearer to always use >=.

Spacetown (changes requested):

gcovr/templates/functions_page.html
Pesa: Well, ok, I don't think saying ">= 0%" is very informative, but I don't really care that much. I'll revert this part.

Spacetown: But it's straight forward in the legend.

Spacetown (approved):

PR #604, opened 2022-04-07
Use ubuntu-20.04 for gcc-8 in docker.

Spacetown:
In #597 for gcc-9 Ubuntu 20.04 was used but for gcc-8 the newer 22.04.
In 22.04 the gcc-8 isn't available anymore.

[no changelog]


latk (approved):

latk:
Thanks for fixing this!

PR #605, opened 2022-04-07
Fix package level function coverage in cobertura.

Spacetown:
As mentioned in #600 the package level function coverage was wrong. This commit fixes this.


latk (approved):
Looks good, the reference files now look more correct.

Spacetown:
> Looks good, the reference files now look more correct.

I've checked some random files. It would be easier if we add line breaks to the reference files.

latk:
>  It would be easier if we add line breaks to the reference files.

If you want to enable `--cobertura-pretty` in the tests I have no objections, provided that at least one test without this option exists.

PR #611, opened 2022-04-16
Respect excluded/noncode lines for aggregated branch coverage.

Spacetown:
Do not count excluded ore noncode lines for aggregated branch coverage as mentioned in #600.


latk (approved):
looks good! I haven't checked the changes to the reference files in detail, but the new output seems to be more correct.

Spacetown:
I've dumped the data to check which branches aren't counted anymore.

PR #612, opened 2022-04-16
Fix list options in configuration files.

### LINKED ISSUE #608: Using "search-path" in config file, __main__.py iterates over the individual characters of the path as a list of paths to search
**Describe the bug**
In my config file, I have a single entry for the key "search-path" (e.g. `search-path = build/`). When I try to run gcovr using that configuration, I'll get an OSError for `Unknown directory 'b'`. It looks like it is breaking down the string for the search-path value. I've confirmed this when I tried to set the path to `./build/`. With the verbose output on, you can see gcovr search the path `.`, then move onto the path `/`.

[I've tracked this down to \_\_main\_\_.py where the list of search paths will be built.](https://github.com/gcovr/gcovr/blob/master/gcovr/__main__.py#L352) By default, this is the root path and the objdir. \_\_main\_\_.py expects this option to be a list type. Therefore, it tries to iterate through the list. When a user is setting the options value using the config file, it is set to the options dictionary as a `str`. \_\_main\_\_.py should check to see if `options.search_paths` is a `str` or `list` before trying to use it as a list.

Current code:
```
for search_path in options.search_paths:
    datafiles.update(find_files(search_path, options.exclude_dirs))
```

Suggested change:
```
if isinstance(options.search_paths, str):
  datafiles.update(find_files(options.search_paths, options.exclude_dirs))
else:
    for search_path in options.search_paths:
        datafiles.update(find_files(search_path, options.exclude_dirs))
```

**To Reproduce**
1. Create a config file where the only value for the key `search-path` is a single path (e.g. `build/`).
2. Run gcovr.
3. Observe that an exception is raised (happens pretty quickly).

**Expected behavior**
A user should be able to put in a single search path from the config file. gcovr should detect that only a single path is being used and search only that one path entry.

Spacetown:
Fix options with more than one argument in configuration files.
The config key `search-path` was stored as a single value. The iteration was done on the characters of the last config entry instead of the path entries.

Only options with `nargs="*"` are fixed. If a option with e.g. 2 arguments is added (I don't think we will add this in the near future) the code has to be adapted.

Closes #608.


latk (approved):
Thank you for fixing this!

PR #617, opened 2022-04-26
Fix decision coverage for excluded lines

Spacetown:
Do not add lines to coverage data by decision analysis.


gcovr/merging.py
```diff
@@ -190,6 +190,13 @@ def insert_line_coverage(
     return _insert_coverage_item(target.lines, line.lineno, line, merge_line, options)
 
 
+def get_line_coverage(target: FileCoverage, lineno: int) -> LineCoverage:
+    """Get the LineCoverage object for that line if available."""
+    if lineno in target.lines:
+        return target.lines[lineno]
+    return None
```
latk, line 197: Instead of creating this function, I'd consider `target.lines.get(lineno)`.

latk (approved):
LGTM. The coverage reports look a lot more realistic now :)
Thanks for fixing this!

Issue #621, opened 2022-05-17
No longer compatible with Keil uVision coverage export


tacooper:
According to the latest Keil MDK/uVision docs:
https://developer.arm.com/documentation/101407/0537/Debug-Commands/COVERAGE/COVERAGE-GCOV-Export?lang=en
`*.gcov` files are still generated with 6 octothorpes (`######`) instead of 5 (`#####`). 

This used to work before adding the new `gcovr/gcov_parser.py` in https://github.com/gcovr/gcovr/commit/304518453d352b21852e94f79ff78a4f4e4d2cb9.
Old parsing for any number of `#` symbols:
https://github.com/gcovr/gcovr/blob/aa7c0a1179fee66f7abc5b83586d4ec933844ed0/gcovr/gcov.py#L355
New parsing for only 5 `#` symbols:
https://github.com/gcovr/gcovr/blob/7279c52e1c33b00b593415cb967dbf1c5b2fc805/gcovr/gcov_parser.py#L768

Is it possible to update this parsing to account for lines with any number of `#` symbols? Or should other coverage tools (such as Keil uVision) be expected to update their coverage generation to use exactly 5 `#` symbols? Is this change or incompatibility documented anywhere?

**Desktop Environment:**
 - OS: Windows 10
 - Keil uVision: 5.37
 - Gcovr: 5.1
 - Python: 3.9.12

Spacetown:
Thanks for reporting this.
In our changelig there is following point (in internal changes):
```
New parser for .gcov file format, should be more robust. (#512)
```
We expected less problems but now this is the second issue. The first issue which was detected by the new parser seems to be a memory corruption in the coverage counters which produces negative values for GCC.

In the documentation of GCC for invoking gcov (https://gcc.gnu.org/onlinedocs/gcc-11.1.0/gcc/Invoking-Gcov.html) the format is described as follow:
```
Unexecuted lines are marked ‘#####’ or ‘=====’, depending on whether they are reachable by non-exceptional paths or only exceptional paths such as C++ exception handlers, respectively. Given the ‘-a’ option, unexecuted blocks are marked ‘$$$$$’ or ‘%%%%%’, depending on whether a basic block is reachable via non-exceptional or exceptional paths.
```

According to this, the previous behavior was wrong. Please raise a bug report for the Keil MDK/uVision.

tacooper:
Thanks for your quick and helpful attention. I have opened a Keil MDK support case.

Spacetown:
[...]

PR #623, opened 2022-05-19
More flexible gcov file parsing

latk:
In #621, it was shown that some third party tools generate `.gcov` files with a slight modification: in lines without coverage, a 6-character `######` marker was used instead of GCC's 5-character marker:

```
 #####:  123: some_code();
```

While gcovr only really supports GCC's gcov and Clang's llvm-cov as coverage data sources, this is a simple fix.

**Question:** should we merge this, or is this unnecessary tech debt?


tacooper (approved):
This flexibility seems mostly harmless™ to me, while also allowing compatibility with older versions of 3rd party tools (for example, even if ARM updates their coverage tool, users bound to older versions of the ARM tool can still benefit from the latest gcovr version).

Spacetown:
If the marker is the same and only the length is different it's ok for me. But if a tool introduce different markers I won't add support for them.

For me this would be more clear:
```python
    def _is_marker(count: str, expected_marker: str) -> bool:
        """Recognize markers for uncovered items like ``#####``."""
        if len(count) < 5:
            return false

        unique_chars = set(count)
        if len(unique_chars) > 1:
            return false

        # Also allow if the marker char is repeated more often,
        # see: <https://github.com/gcovr/gcovr/issues/621>
        return expected_marker == unique_chars[0]
```
And using it with the single character of the marker:
```python
_is_marker(count_str, "#")
```

latk:
@Spacetown That's a very good point. I updated the function to the far more clearer:

```python
def _is_marker(count_str: str, marker_char: str) -> bool:
    """Recognize markers for uncovered items like ``#####``."""
    # The GCC gcov marker is exactly 5 chars long,
    # but some tools generate longer markers
    # (see <https://github.com/gcovr/gcovr/issues/621>).
    return len(count_str) >= 5 and all(c == marker_char for c in count_str)
```

Spacetown (approved):
LGFM.

Spacetown:
@latk Shall we merge this? Pro is that the user of Keil uVision can still use the latest version of `gcov´ without any patch.

PR #638, opened 2022-07-03
Fixed usage of realpath vs. normpath

### LINKED ISSUE #635: gcovr not scanning files if source files located below virtual path
**Describe the bug**

**To Reproduce**

 - OS: Windows
 - GCC version 6.3.0
 - GCOVR version 5.1

simonCor:
These functions were equivalent. This changed with I think python 3.8.

This produced problems with virtual drives (subst) on windows: Resulted in filtering all files.

(This currently is a theoretical fix. Will test this on monday. But already posting it for comments.)

This may also fix #635


simonCor:
Ok tested it was not the only problem - will update soon

Issue #121, opened 2016-01-11
Supporting "new" LCOV / GCOV markers


vito-wagl:
Adding the three "new" LCOV / GCOV markers, which will be excluded code from branch coverage:
- LCOV_EXCL_BR_LINE
- LCOV_EXCL_BR_START
- LCOV_EXCL_BR_STOP

[geninfo(1) man page](http://ltp.sourceforge.net/coverage/lcov/geninfo.1.php)

Blabla51:
You might be able to close this issue @Spacetown :)

PR #644, opened 2022-07-31
Add EXCL_BR_* markers

Blabla51:
Adding the branch exclude markers, as requested in #121 

Renamed `gcovr/tests/excl-branch` to `gcovr/tests/excl-line-branch`, and added a new `gcovr/tests/excl-branch` test for the introduced markers.

Some test references might be missing for gcc-5 / clang-10 / clang-13


Spacetown:
Thank you for working on this.
Please check the changed doc files, this should be undone. The reference can be generated with docker for all compilers.

Blabla51:
I fixed the permissions, as well as the changelog.

I did not manage to run update the tests files with docker. I'm running it in WSL, and it does not update any file using this command: `python3 -m nox --session 'docker_qa_compiler(all)' -- --update_reference -k 'excl-branch'`

gcovr/gcov_parser.py
```diff
@@ -1024,16 +1051,67 @@ def _find_excluded_ranges(
...
+    for header, lineno in exclusion_stack_branch:
         warnings.start_without_stop(
             lineno, f"{header}_EXCL_START", f"{header}_EXCL_STOP"
         )
 
-    return _make_is_in_any_range(exclude_line_ranges)
+    return [
+        _make_is_in_any_range(exclude_line_ranges),
+        _make_is_in_any_range(exclude_branch_ranges),
+    ]
```
Spacetown, line 1114:
```suggestion
    return (
        _make_is_in_any_range(exclude_line_ranges),
        _make_is_in_any_range(exclude_branch_ranges),
    )
```

gcovr/gcov_parser.py
```diff
@@ -959,58 +972,72 @@ def _find_excluded_ranges(
     *,
     warnings: _ExclusionRangeWarnings,
     exclude_lines_by_pattern: Optional[str] = None,
+    exclude_branches_by_pattern: Optional[str] = None,
     exclude_pattern_prefix: str,
 ) -> Callable[[int], bool]:
```
Spacetown, line 977: Not sure about the syntax here:
```suggestion
) -> Tuple(Callable[[int], bool], Callable[[int], bool]):
```

gcovr/gcov_parser.py
```diff
@@ -1024,16 +1051,67 @@ def _find_excluded_ranges(
...
+                        )
+
+                    exclude_branch_ranges.append((start_lineno, lineno))
+
+                else:  # pragma: no cover
+                    pass
+
         if exclude_lines_by_pattern_regex:
             if exclude_lines_by_pattern_regex.match(code):
                 exclude_line_ranges.append((lineno, lineno + 1))
```
Spacetown, line 1095: This should be moved to the end of the first loop.

gcovr/gcov_parser.py
```diff
@@ -1024,16 +1051,67 @@ def _find_excluded_ranges(
                 else:  # pragma: no cover
                     pass
 
+            # branche exclusion
+
+            excl_branch_pattern = re.compile(
```
Spacetown, line 1056: Could the two loops be replaced by a function with the regex and lists as argument and returning the excluded ranges? They are doing the same with different variables.

gcovr/gcov_parser.py
```diff
@@ -1024,16 +1051,67 @@ def _find_excluded_ranges(
...
+                "(" + exclude_pattern_prefix + ")" + _EXCLUDE_BRANCH_PATTERN_POSTFIX
+            )
+            for header, flag in excl_branch_pattern.findall(code):
+
+                if flag == "LINE":
+                    if exclusion_stack_branch:
+                        warnings.branch_after_start(
```
Spacetown, line 1063: The warning is the same. Replaced by accident?
```suggestion
                        warnings.line_after_start(
```

Spacetown (changes requested):
Some minor changes as suggested.

Spacetown:
> I fixed the permissions, as well as the changelog.
> 
> I did not manage to run update the tests files with docker. I'm running it in WSL, and it does not update any file using this command: `python3 -m nox --session 'docker_qa_compiler(all)' -- --update_reference -k 'excl-branch'`

This command is for updating existing reference files with new content but doesn't create new ones. Please copy the reference folder from the renamed test and run the command again.

gcovr/gcov_parser.py
```diff
@@ -959,58 +972,72 @@ def _find_excluded_ranges(
...
-    Scan through all lines to find line ranges covered by exclusion markers.
+    Scan through all lines to find line ranges and branch ranges covered by exclusion markers.
 
     Example:
-    >>> lines = [(11, '//PREFIX_EXCL_LINE'), (13, '//IGNORE'), (15, '//PREFIX_EXCL_START'), (18, '//PREFIX_EXCL_STOP')]
-    >>> exclude = _find_excluded_ranges(
-    ...     lines, warnings=..., exclude_lines_by_pattern = '.*IGNORE', exclude_pattern_prefix='PREFIX')
-    >>> [lineno for lineno in range(20) if exclude(lineno)]
+    >>> lines = [(11, '//PREFIX_EXCL_LINE'), (13, '//IGNORE_LINE'), (15, '//PREFIX_EXCL_START'), (18, '//PREFIX_EXCL_STOP'),
+    ...     (21, '//PREFIX_EXCL_BR_LINE'), (23, '//IGNORE_BR'), (25, '//PREFIX_EXCL_BR_START'), (28, '//PREFIX_EXCL_BR_STOP')]
```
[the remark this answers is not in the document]
Blabla51: Do we want to have something different than the default LCOV marker `LCOV_EXCL_BR_LINE` ?

Since the line exclusion markers are compatible, I would have expect the same for the branch ones. 

http://ltp.sourceforge.net/coverage/lcov/geninfo.1.php

gcovr/gcov_parser.py
Blabla51: Fixed

gcovr/gcov_parser.py
Blabla51: Fixed

gcovr/gcov_parser.py
Blabla51: Fixed

gcovr/gcov_parser.py
```diff
@@ -1024,16 +1051,67 @@ def _find_excluded_ranges(
                 else:  # pragma: no cover
                     pass
 
+            # branche exclusion
+
```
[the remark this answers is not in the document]
Blabla51: Fixed

gcovr/gcov_parser.py
```diff
@@ -959,58 +972,72 @@ def _find_excluded_ranges(
...
+        if _EXCLUDE_FLAG in code:
             # process the exclusion marker
             #
             # header is a marker name like LCOV or GCOVR
             #
             # START flags are added to the exlusion stack
             # STOP flags remove a marker from the exclusion stack
+
+            # line exclusion
+
```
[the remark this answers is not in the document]
Blabla51: Fixed

gcovr/gcov_parser.py
```diff
@@ -472,15 +480,18 @@ def _gather_coverage_from_line(
...
-            branchno,
-            BranchCoverage(
-                count=count,
-                fallthrough=(annotation == "fallthrough"),
-                throw=(annotation == "throw"),
-            ),
-        )
+        is_excluded = branch_is_excluded(state.lineno)
+
+        if not is_excluded:
```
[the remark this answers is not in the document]
Blabla51: Fixed

gcovr/gcov_parser.py
```diff
@@ -360,14 +364,16 @@ def parse_coverage(
         ]
 
     if flags & ParserFlags.RESPECT_EXCLUSION_MARKERS:
-        line_is_excluded = _find_excluded_ranges(
+        [line_is_excluded, branch_is_excluded] = _find_excluded_ranges(
```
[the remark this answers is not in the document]
Blabla51: Fixed

gcovr/gcov_parser.py
Blabla51: There is only one loop, that's why it's currently here.

We can split it into two loops, but that means going through the file twice. Is it worth it ?

gcovr/gcov_parser.py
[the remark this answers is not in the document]
Spacetown: LCVO wasn't in my mind. if there `BR` is used please ignore the suggestions regarding the renaming.

gcovr/gcov_parser.py
Spacetown: By mail I got a comment from you for this here but it isn't visible in the web interface. Did you remove it?

gcovr/gcov_parser.py
Spacetown: ...or only the if/elif/else block.

Blabla51: I just replaced it all by a function, it's easily changeable now

gcovr/gcov_parser.py
Blabla51: Yes sorry, I should have only edit my comment. I was not looking at the correct line.

It's now part of the whole function `_process_exclusion_marker`

gcovr/gcov_parser.py
```diff
@@ -954,86 +967,155 @@ def line_after_start(self, lineno: int, start: str, start_lineno: str) -> None:
...
+    STOP flags remove a marker from the exclusion stack
+    """
+
+    if _EXCLUDE_FLAG in code:
+
+        # marker exclusion
+        excl_pattern = re.compile(
+            "(" + exlcude_prefix + ")" + _EXCLUDE_FLAG + exclude_word + exclude_postfix
+        )
+        for header, flag in excl_pattern.findall(code):
```
Spacetown, line 996: I think it's better to keep this outside of the function to reduce the arguments.

gcovr/gcov_parser.py
```diff
@@ -954,86 +967,155 @@ def line_after_start(self, lineno: int, start: str, start_lineno: str) -> None:
...
+                pass
+
+    if exclude_by_pattern_regex:
+        if exclude_by_pattern_regex.match(code):
+            exclude_ranges.append((lineno, lineno + 1))
+
+    return (
+        exclude_ranges,
+        exclusion_stack,
+    )
```
Spacetown, line 1041: Don't need to return this. This lists are only modified by append and therefore the modification is visible outside of the function.

gcovr/gcov_parser.py
```diff
@@ -954,86 +967,155 @@ def line_after_start(self, lineno: int, start: str, start_lineno: str) -> None:
         )
 
 
+def _process_exclusion_marker(
+    code: str,
+    lineno: int,
+    exlcude_prefix: str,
```
Spacetown, line 973: Typo
```suggestion
        exclude_prefix: str,
```

gcovr/gcov_parser.py
Spacetown: No problem, I was only searching the comment and thought I'm old.

gcovr/gcov_parser.py
Blabla51: Thanks for the feedback, fixed

gcovr/gcov_parser.py
Blabla51: Fixed

gcovr/gcov_parser.py
Blabla51: Fixed, I kept only what's in the for loop inside the function.

Blabla51:
I added the possible gcovr argument `--exclude-lines-by-pattern`, as well as its test. I hope it's all good

Thanks for the review!

gcovr/gcov_parser.py
```diff
@@ -954,86 +967,166 @@ def line_after_start(self, lineno: int, start: str, start_lineno: str) -> None:
...
+        + _EXCLUDE_PATTERN_POSTFIX
+    )
+    excl_branch_pattern = re.compile(
+        "("
+        + exclude_pattern_prefix
+        + ")"
+        + _EXCLUDE_FLAG
+        + _EXCLUDE_BRANCH_WORD
+        + _EXCLUDE_PATTERN_POSTFIX
+    )
```
Spacetown, line 1080: This operation is only needed if `_EXCLUDE_FLAG` is found in code (performance issue). You can set the values to `None` here and init the regex on first entry of if block in line 1083.

Blabla51: Indeed, I thought it was already improving the performance if it was defined only once (it was for each exclude line before), but it's even better if it's declared only when it's needed!

Fixed

Blabla51:
Do I need to do anything else before merging @Spacetown ?

Spacetown (approved):

PR #650, opened 2022-08-07
Move themes to CSS, restore --html-tab-size

latk:
Previously, colors for the themes were specified in Python code.

Now, the colors for all themes are listed in the CSS styles. The current theme is selected via a `theme-green` or `theme-blue` class on the `<html>` element.

In the future, this would make it possible to switch themes at runtime with a bit of JavaScript that changes the classes in the DOM.

While making this change, it was discovered that the "tab_size" variable wasn't used properly. This fixes the regression in b186a6f04f8b62a8f88c6fb0f60af36af778fe34 / #381.


gcovr/templates/style.css
```diff
@@ -1,19 +1,26 @@
...
+    --high_color: #85E485;
+    --covered_color: #85E485;
+    --uncovered_color: #FF8C8C;
+    --excluded_color: #53BFFD;
+    --warning_color: orangered;
+    --takenBranch_color: green;
+    --notTakenBranch_color: red;
+    --takenDecision_color: green;
+    --uncheckedDecision_color: darkorange;
+    --notTakenDecision_color: red;
```
Spacetown, line 16: Shall we move this to a `:root.theme-green` and the common parts to a `:root.theme-green :root.theme-blue` to get a report with no colors if the theme isn't green or blue? Or better rename the theme `green` to `default`?

Spacetown (approved):
LGFM. The windows reference needs to be updated.

gcovr/templates/style.css
latk: I think its sensible to provide some colors by default. This makes it possible for the blue theme to just override a couple of variables. 

We could rename the `green` theme, but should it then only be renamed in CSS or also on the command line? Because changing the CLI argument would break backwards compatibility, whereas using different names in different parts of the program would be confusing. So I'd tend to leave all names as they currently are.

It would of course be possible to only add `theme-X` classes for non-default themes, but I think it's clearer to mention the selected theme, even if it doesn't do anything right now.

Spacetown: You're right, since `green` is a CLI argument it shouldn't be renamed. Following CSS is working in Firefox and Safari and seems more clear for me:
```css
:root {
    font-family: sans-serif;
    --tab_size: 4;
}

:root.theme-green, :root.theme-blue {
    --unknown_color: lightgray;
    --low_color: #FF6666;
    --medium_color: #F9FD63;
    --high_color: #85E485;
    --covered_color: #85E485;
    --uncovered_color: #FF8C8C;
    --excluded_color: #53BFFD;
    --warning_color: orangered;
    --takenBranch_color: green;
    --notTakenBranch_color: red;
    --takenDecision_color: green;
    --uncheckedDecision_color: darkorange;
    --notTakenDecision_color: red;
}

:root.theme-blue {
    --high_color: #66B4FF;
    --covered_color: #66B4Ff;
    --takenBranch_color: blue;
}
```

Spacetown: Or move the theme specific colors in a separate block for each theme.

latk: yes that looks a lot clearer

PR #652, opened 2022-08-07
Print a warning if root directory contains symlinks.

### LINKED ISSUE #635: gcovr not scanning files if source files located below virtual path

Spacetown:
As suggested in #635 print a warning if the root directory contains symlinks.

Closes #635

### DIFF
--- a/gcovr/__main__.py
+++ b/gcovr/__main__.py
@@ -255,6 +255,11 @@ def main(args=None):
 
     options.starting_dir = os.path.abspath(os.getcwd())
     options.root_dir = os.path.abspath(options.root)
+    if options.root_dir != os.path.realpath(options.root_dir):
+        logger.warning(
+            "It seems, that your root contains a symlink. This may result in problems. "
+            "See issue https://github.com/gcovr/gcovr/issues/635."
+        )
 
     #
     # Setup filters
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,23 @@
 .. program is needed to resolve option links
 .. program::  gcovr
 
+Next Release
+------------
+
+Known bugs:
+
+Breaking changes:
+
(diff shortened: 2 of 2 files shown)


gcovr/__main__.py
```diff
@@ -255,6 +255,11 @@ def main(args=None):
 
     options.starting_dir = os.path.abspath(os.getcwd())
     options.root_dir = os.path.abspath(options.root)
+    if options.root_dir != os.path.realpath(options.root_dir):
+        logger.warning(
+            "It seems, that your root contains a symlink. This may result in problems. "
+            "See issue https://github.com/gcovr/gcovr/issues/635."
+        )
```
latk, line 262: I'd consider telling the user more explicitly how they can resolve the problem:

```suggestion
    root_dir_realpath = os.path.realpath(options.root_dir)
    if options.root_dir != root_dir_realpath and not options.filters:
        logger.warning(
            "Your project --root directory seems to contain a symlink. "
            "This will EXCLUDE source files in your root directory! "
            "To fix this, you may have to add a --filter option:\n"
            "    --filter=%s\n"
            "For details, see https://github.com/gcovr/gcovr/issues/635",
            re.escape(root_dir_realpath)
        )
```

Untested, of course. Speaking of which: is there an easy way to test this warning?

I also think that the period `.` after the URL should be removed to make it easier to click/copy.

Spacetown: I'll try to add a test. A funny think on Linux, if I open python inside a symlink and running the commands `os.path.abspath('.')` and `os.path.realpath('.')`, the output is for both the target of the symlink. If I use the path to the directory instead of `'.'` the `os.abspath` doesn't resolve the links.

PR #653, opened 2022-08-07
Do not ignore returncode of `gcov`.

### LINKED ISSUE #254: gcov segfaults but gcovr ignores this fact
the gcov call produced a segfault and the exit code of that particular call wasn't taken into account.

Spacetown:
If `gcov` exits with a value != 0 raise an exception.

Closes #254

### DIFF
--- a/gcovr/gcov.py
+++ b/gcovr/gcov.py
@@ -486,19 +486,35 @@ class GcovProgram:
     def run_with_args(self, args, **kwargs):
         """Run the gcov program
 
-        >>> GcovProgram("python3").run_with_args(["-c", "import sys; sys.exit(1);"])
+        >>> import platform
+        >>> if platform.system() == "Windows":
+        ...     print("kill not working on Windows")  # doctest: +SKIP
+        ... else:
+        ...     GcovProgram("bash").run_with_args(["-c", "exit 1"])
         Traceback (most recent call last):
         ...
         RuntimeError: GCOV returncode was 1.
-        >>> GcovProgram("python3").run_with_args(
-        ...     [
-        ...         "-c",
-        ...         "import os, signal; os.kill(os.getpid(), signal.SIGTERM)"
-        ...     ]
-        ... )
+        >>> if platform.system() == "Windows":
+        ...     GcovProgram("bash").run_with_args(["-c", "exit 1"])
+        ... else:
+        ...     print("kill not working on Windows")  # doctest: +SKIP
+        Traceback (most recent call last):
+        ...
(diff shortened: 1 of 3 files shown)


gcovr/gcov.py
```diff
@@ -483,15 +483,22 @@ def run_gcov_and_process_files(abs_filename, covdata, options, error, toerase, c
...
                 cwd=chdir,
-            ).communicate()
+            )
+            out, err = gcov_process.communicate()
+            if gcov_process.returncode < 0:
+                raise RuntimeError(
+                    f"GCOV returncode was {gcov_process.returncode} (exited by signal)."
+                )
+            elif gcov_process.returncode != 0:
+                raise RuntimeError(f"GCOV returncode was {gcov_process.returncode}.")
```
latk, line 501: Can this code be tested with some fault injection technique? I think something like `--gcov-executable 'bash -c "exit 1"'` could work for normal return codes, and `--gcov-executable 'bash -c "kill -9 \$\$"'` for the signal case (Unixoid systems only). But this doesn't have to be a full integration test, if this part of the logic is extracted into a separate function.

Spacetown: I was also thinking about how to test this. A doctest is a good idea. Since we are testing with a installed Mays on Windows this should also work there.

gcovr/gcov.py
```diff
@@ -471,6 +471,29 @@ def run_with_args(self, args, **kwargs):
...
+        """Run the gcov program
+
+        >>> gcov("bash").run_with_args(["-c", "exit 1"])
+        Traceback (most recent call last):
+        ...
+        RuntimeError: GCOV returncode was 1.
+        >>> gcov("bash").run_with_args(["-c", "kill $$"])
+        Traceback (most recent call last):
+        ...
+        RuntimeError: GCOV returncode was -15 (exited by signal).
```
latk, line 484: Hmm, seems like this is more complicated after all. The `gcov()` constructor invokes those programs and tries to determine valid command line options, which seems to cause test failures on Windows. I'm not sure how this can be sidestepped elegantly, in particular because the gcov class maintains significant global state.

My guess is that it would make sense to rename the gcov class (maybe `GcovProgram`?), with a constructor that doesn't do any sniffing. Then, a function `gcov()` could manage option checking + caching. This would avoid changes to other parts of the code.

Spacetown: I'll try but instead of `gcov` I prefer something like `cache_gcov_capabilities()`.

PR #655, opened 2022-08-09
Add methods to cobertura report

### LINKED ISSUE #654: Method level coverage in the xml generated
Issue:

The xml file generated from gcovr is not having method/function level coverage.

```
<class branch-rate="0.285714285714" complexity="0.0" filename="src/absc/mno/xyz/ahgs.cpp" line-rate="0.231481481481" name="os_clib_hxx">
<methods/>
<lines>
<line branch="false" hits="0" number="200"/>
<line branch="false" hits="0" number="202"/>
<line branch="false" hits="3" number="208"/>
<line branch="false" hits="3" number="210"/>
<line branch="false" hits="63" number="213"/>
<line branch="true" condition-coverage="50% (1/2)" hits="63" number="215">
<conditions>
<condition coverage="50%" number="0" type="jump"/>
</conditions>
</line>
<line branch="false" hits="0" number="218"/>
.........
..........
```

Expected:

Function level coverage to be visible in xml

Spacetown:
As suggested in #654 this PR adds methods to the cobertura report.

The evaluation of the line coverage for functions a dummy `FileCoverage` is created to which the lines of the function are added.

Closes #654.


gcovr/writer/cobertura.py
```diff
@@ -74,14 +74,52 @@ def print_cobertura_report(covdata: CovData, output_file, options):
...
+            for lineno in function_linenos:
+                line_cov = data.lines[lineno]
+                dummy_file_coverage.lines[lineno] = line_cov
+                if not (line_cov.is_covered or line_cov.is_uncovered):
+                    continue
+
+                b = line_cov.branch_coverage()
+                if b.total:
+                    function_branch += b
+                lines.append(_line_element(line_cov))
```
latk, line 115: This looks like duplicate code to the line element generation below – could probably extracted into a function.

gcovr/writer/cobertura.py
```diff
@@ -74,14 +74,52 @@ def print_cobertura_report(covdata: CovData, output_file, options):
...
+
+            lines = etree.Element("lines")
+            elem.append(lines)
+            function_branch = CoverageStat(0, 0)
+            function_linenos = []
+            for lineno in iter_reversed_sorted_lines:
+                if lineno >= function_cov.lineno:
+                    function_linenos.insert(0, lineno)
+                else:
+                    break
```
latk, line 103: This seems to assume that every line belongs to a function, and that there's no code between functions. Perhaps it would be better to avoid making such guesses, and just generate an empty `<lines/>` element in each method section.

latk (commented):
I have two general concerns: method-level line coverage, and the concept of dummy FileCoverage objects.

## Attributing lines to functions is impossible for us (right now)

I am worried that the line-to-function attribution might not work in GNU-C or C++ since functions can be nested.

In GNU-C, we can have simple nested functions:

```c
int outer() {
  int inner() {
    return 40;
  }

  return inner() + 2;
}
```

In C++11, we'll frequently see lambdas, or other inner classes:

```c++
int outer() {
  auto inner = []() { return 40; };
  return inner() + 2;
}
```

If I read your code correctly, the outer return would be misattributed to the inner function. There is no way for us to fix this without parsing the source code, and we just can't do that. I'd rather provide no data than wrong data (at least by default).

The only way to do this correctly would be to use compiler-level data. Gcov does actually provide the necessary data in its JSON output, which we still aren't parsing :sweat_smile:

## Dummy FileCoverage

This seems like a bit of a code smell. This suggests the need for some kind of RegionCoverage that contains branch/line coverage but does not have a filename. The FileCoverage might inherit from such a RegionCoverage.

Spacetown:
@latk You're complete right. Since the methods only make sense with the lines (needed to get the coverage), I think we should close this without merge.

gcovr/writer/cobertura.py
Spacetown: But then the method element has no information about coverage.

latk:
> Since the methods only make sense with the lines (needed to get the coverage)

Oh right. Hmm. This is more difficult than expected. So I think this is blocked on adapting our internal data model to include additional necessary information and implementing a gcov-json reader (#282).

That reader should also enable significant performance improvements, and might be able to circumvent the need for the heuristics for finding the working directory. But handling exclude markers + noncode lines would be much more difficult. I think the necessary logic would have to be extracted first from the parser, probably as a function `apply_exclude_markers(cov: FileCoverage, lines: list[str], options) -> FileCoverage`.

PR #656, opened 2022-08-10
Add base path for json.

### LINKED ISSUE #397: Merging coverage run options
Hello there,
I hope I'm not addressing the wrong community. We're using gcovr to produce coverage stats for our library tests. Over the time test applications at different layers were developed, for which we seek to produce a merged coverage report of the one library they all test. Up to now, the applications would fetch the source-code of the library from the repo and build it in place in order to have the gcnos in place naturally. Now the natural course of action which should work in my eyes would be, to upgrade the build-system of the applications so one could configure them to use a central (yet still local) build of the library with gcov instrumentation. Then the runs should produce aggregated .gcdas and we're all fine. In my eyes this would be the right thing to do (tm) anyways, as this assures naturally that the same binary was tested.

Yet still I feel responsible to discover, whether there maybe is a way to merge coverage information from different build-trees of the same library (to pursue a quick an low-effort way in kinds), provided of course that the revisions and thus the sources match exactly. In my understanding of things, this would only be possible on the gcovr level of the processing tools, but I will gladly take be disproved.

Do you maybe have any advice in this regards?

Spacetown:
As suggested in #397 an option `--json-base` is added which is prepended to the file names in the JSON reports. The value is normalized first and `\`are replaced by `/`.

Closes #397

### DIFF
--- a/doc/source/guide/merging.rst
+++ b/doc/source/guide/merging.rst
@@ -1,3 +1,7 @@
+
+.. program is needed to resolve option links
+.. program::  gcovr
+
 .. _merging_coverage:
 
 Merging Coverage Data
@@ -27,8 +31,7 @@ you have to place your pathnames with wildcards in double quotation marks::
     gcovr --add-tracefile "run-*.json" --html-details coverage.html
 
 If you want to merge coverage reports generated in different `--root` directories you
-can use the option `--json-base=dir/to/prepend` to get the same root directory for all
-reports.
+can use the :option:`--json-base` to get the same root directory for all reports.
 
 .. versionadded:: NEXT
 
--- a/doc/source/guide/merging.rst
+++ b/doc/source/guide/merging.rst
@@ -25,3 +25,11 @@ duplicating :option:`-a/--add-tracefile<gcovr --add-tracefile>`. With this optio
 you have to place your pathnames with wildcards in double quotation marks::
 
     gcovr --add-tracefile "run-*.json" --html-details coverage.html
+
+If you want to merge coverage reports generated in different `--root` directories you
+can use the option `--json-base=dir/to/prepend` to get the same root directory for all
+reports.
+
+.. versionadded:: NEXT
+
+   The :option:`gcovr --json-base` option.
--- a/gcovr/configuration.py
+++ b/gcovr/configuration.py
@@ -28,7 +28,7 @@ import datetime
 import os
 import re
 
-from .utils import FilterOption
+from .utils import FilterOption, force_unix_separator
 from .writer.html import CssRenderer
 
 
@@ -1087,7 +1087,7 @@ GCOVR_CONFIG_OPTIONS = [
         group="output_options",
         metavar="PATH",
         help="Prepend the given path to all file paths in JSON report.",
-        type=str,
+        type=lambda p: force_unix_separator(os.path.normpath(p)),
         default=None,
     ),
     GcovrConfigOption(
--- a/gcovr/configuration.py
+++ b/gcovr/configuration.py
@@ -1082,6 +1082,15 @@ GCOVR_CONFIG_OPTIONS = [
         action="store_true",
     ),
     GcovrConfigOption(
+        "json_base",
+        ["--json-base"],
+        group="output_options",
+        metavar="PATH",
+        help="Prepend the given path to all file paths in JSON report.",
+        type=str,
+        default=None,
+    ),
+    GcovrConfigOption(
         "csv",
         ["--csv"],
         group="output_options",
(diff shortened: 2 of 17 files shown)


gcovr/writer/json.py
```diff
@@ -58,6 +57,26 @@
 PRETTY_JSON_INDENT = 4
 
 
+get_filename = None
+
+
+def set_filename_function(options):
+    if options.json_base:
+        options.json_base = os.path.normpath(options.json_base).replace("\\", "/")
```
latk, line 65: Should this replace `\\`, or only `os.path`?

gcovr/writer/json.py
```diff
@@ -58,6 +57,26 @@
 PRETTY_JSON_INDENT = 4
 
 
+get_filename = None
```
latk, line 60: Not a fan of global variables. The code can instead call a function like `prefixed_filename(filename, options)`.

gcovr/configuration.py
```diff
@@ -1081,6 +1081,15 @@ def merge_options_and_set_defaults(
         ),
         action="store_true",
     ),
+    GcovrConfigOption(
+        "json_base",
+        ["--json-base"],
+        group="output_options",
+        metavar="PATH",
+        help="Prepend the given path to all file paths in JSON report.",
```
latk, line 1089: Could also use a mention/example/explanation in the "Merging Coverage Data" user guide page.

gcovr/writer/json.py
Spacetown: The backslash shall be replaced with forward slash. This is done in several places, maybe better create a function `gcovr_normalize`.

gcovr/writer/json.py
Spacetown: I wanted to save the time for normalizing the value of json_base but this can also be done in the commandlin parser.

gcovr/configuration.py
Spacetown: Will add this.

gcovr/writer/json.py
latk: A quick grep for this code pattern yields the following occurrences:

```none
gcovr/configuration.py:            value = str(self.value).replace("\\", os.sep).replace("/", os.sep)
gcovr/gcov.py:    gcovname = gcovname.replace("/", os.sep)
gcovr/tests/test_gcovr.py:    contents = contents.replace("\\", "/")
gcovr/tests/test_gcovr.py:    contents = contents.replace("\\", "/")
gcovr/utils.py:    return path.replace(os.path.sep, "/")
gcovr/utils.py:    return normalized.replace("\\", "/")
gcovr/writer/cobertura.py:        package.set("name", packageName.replace("/", "."))
gcovr/writer/cobertura.py:    etree.SubElement(sources, "source").text = options.root.strip().replace("\\", "/")
gcovr/writer/html.py:        return "." if self.directory == "" else self.directory.replace("\\", "/")
gcovr/writer/html.py:        ).replace("\\", "/")
gcovr/writer/html.py:    filename = filename.replace(os.sep, "/")
gcovr/writer/txt.py:        fh.write("Directory: " + options.root.replace("\\", "/") + "\n")
noxfile.py:    session.env["CC"] = shutil.which(session.env["CC"]).replace(os.path.sep, "/")
noxfile.py:    session.env["CXX"] = shutil.which(session.env["CXX"]).replace(os.path.sep, "/")
noxfile.py:    ).replace(os.path.sep, "/")
```

Replacing `\` to `/` is fine when done for display purposes, i.e. if it is not necessary to resolve the path afterwards. But to be ultra-pedantic, `\` is a legal character for filenames on Linux, so replacing it could technically break some use cases.

But it seems that `presentable_filename()` does exactly this transformation as well, and is used in the JSON format, so that your suggested code is in line with how things currently work.

Spacetown: I can change this with a separate PR to only replace this on Windows. Than `os.path.sep` is correct.

doc/source/guide/merging.rst
```diff
@@ -25,3 +25,11 @@ duplicating :option:`-a/--add-tracefile<gcovr --add-tracefile>`. With this optio
 you have to place your pathnames with wildcards in double quotation marks::
 
     gcovr --add-tracefile "run-*.json" --html-details coverage.html
+
+If you want to merge coverage reports generated in different `--root` directories you
+can use the option `--json-base=dir/to/prepend` to get the same root directory for all
```
latk, line 30:
```suggestion
can use the option ``--json-base=dir/to/prepend`` to get the same root directory for all
```

RST is a very “special” markup format. Single delimiters like `` `text` `` or `'text'` are just *italics*. For code markup, there must be two backticks.

Alternatively, we could place a link to the ``:option:`--json-base-dir` `` here.

latk (approved):

PR #657, opened 2022-08-16
Add option to write changelog entry to separate file for each PR.

Spacetown:
Move the content of the `CHANGELOG.rst` to `doc/source/changelog.rst`. New fragments can be added in folder `doc/source/changelog.new` which is included into the changelog by glob.
Now we don''t have a merge bottleneck with conflicts in most PRs. From time to time or at the end of the release the fragments can be moved to the changelog by a separate PR or pushed direct.


latk:
I think it would be good to a have a release-checklist check to ensure that the fragments have been (manually) added to the changelog before a release. It may also be useful to require that the news fragments start with some keyword, e.g. `Internal:`, `Bugfix:`, `New Feature:` etc.

I've looked at the [Towncrier documentation](https://towncrier.readthedocs.io/en/latest/index.html) which has a broadly similar goal, but it's somewhat opinionated and might not quite fit our use case. Thus, rolling our own solution is OK. But we can look at Towncrier for inspiration regarding useful features and workflows.

Spacetown:
I'll add the check. Shall also the file start with a keyword?

latk:
> Shall also the file start with a keyword?

Including a keyword in the filename sounds easier to manage than a prefix in the title, if I think about it. But do whatever you think is best to keep the changelog easy to manage and releases easy to publish.

Spacetown:
I've changed the `release_checklist` to use GitHub error messages and notes. In console it looks like this:
```
>./admin/release_checklist --no-verify-change-fragments 5.2
::notice file=doc/source/changes/feature.Remove_bottleneck.rst :: Move fragment to doc/source/changelog.rst
SUCCESS: release may proceed
```
which results in a annotation like [this](https://github.com/gcovr/gcovr/pull/657/files#annotation_4356265972).

latk:
The more I think about this the less enthusiastic I get. Yes, this approach avoids merge conflicts. But that is the only positive aspect I can see. Downsides:

* Writing changelog entries is now a lot more effort: have to pick a filename, have to add a title just to make RST/Sphinx happy.
* The rendered changelog for unreleased snapshots is now a lot less readable.
* Cutting releases is now more effort: changelog entries must be manually moved into the main changelog file. While I think it's good to check, reorganize, and maybe rewrite the changelog before releasing, this feels like an unnecessary chore.
* The release-checklist warnings might be confusing for new contributors.

A recurring problem is that Sphinx-Doc has annoying restrictions that make it uncomfortable to use. An `.. include:: changes/*` directive would greatly simplify things, addressing criticism 1 and 2. But that doesn't exist, and I'm not quite frustrated enough to start writing custom Sphinx plugins.

Approaches I would favour:

* keep the status quo ← probably best
* adopt some tool that does changelog management outside of Sphinx, for example Towncrier
* add tooling to automatically update the changelog for an active PR, just before it is merged. Such a tool could also insert PR numbers automatically. I've looked at tools like [Bors](https://bors.tech/) but it can't modify the code. Could hack something together with a GitHub Action, though.

What do you think?

Spacetown:
I'm with you. For solving merge conflicts we get other problems. Introducing a new tool seems to be a overkill for me.
A good solution will be something like the `[no changelog]` in the commit message which is used when the PR is merged to update the changelog.
I'll close this PR. If I've a better solution I can reopen it.

PR #661, opened 2022-08-19
Add option to bundle app with pyinstaller

### LINKED ISSUE #486: Build portable single executable
Is it possible to build a single executable for simpler distribution of the tool?
For example with use of pyinstaller build a single `gcovr.exe` (for windows) file and use it in without need of 
installing python and pip?

Note: I think this is useful only for Windows platform.

Spacetown:
This PR adds an additional option to bundle the current version installed with pip into a standalone executable with pyinstaller:
```
Experimental Options:
  Experimental options are for testing only and can be changed/removed in
  future releases.

  --bundle-app          Build a standalone application from the current gcovr
                        installation. To add additional options to pyinstaller
                        use e.g.: --bundle-app -- --help
```

The main script used to generate the application is part of gcovr installation. In the bundled application the option and option group is removed.
The pipeline test run also a test for the bundled application which checks if the help is printed.

Closes #486


latk:
I have no idea how PyInstaller works. Does this option have to be part of the program being packaged, or could it be a separate script, or could we just tell users interested in this to clone the repository and run some Nox command?

I think it's great to include (low-maintenance) entry-points that can be used for non-PyPI distribution, but the proposed design strikes me more like a hack, not like something I'd feel comfortable shipping.

Spacetown:
Pyinstaller needs a script which is parsed to get the imports. For the modules the pyc files are created and packed together with the python interpreter. A bootstrap code unpacks all, setsup the environment and executes it.
Of course it can also be done in nox. Since this is something for more experienced users I thought a option would be better. But you're right, it's more like a little hack. I'll move it completely to the nox level and add it to the documentation. Should this go to cookbook?

latk:
> I'll move it completely to the nox level and add it to the documentation. Should this go to cookbook?

Ok, from that description it really sounds like this should be a separate script, not part of the gcovr program. Doing this via Nox sounds fine.

Yes, adding it to the docs would be good for discoverability. Maybe a cookbook entry, and a link to that entry from the installation guide?

.github/workflows/test.yml
```diff
@@ -137,11 +137,6 @@ jobs:
         sudo apt-get clean
     - name: Install ninja
       uses: seanmiddleditch/gha-setup-ninja@master
-    - name: Install libxml2 and libxslt dev (pypy3)
-      if: ${{ startsWith(matrix.os,'ubuntu-') && ( matrix.python-version == 'pypy3' ) }}
-      run: |
-        sudo apt-get install -y libxml2-dev libxslt-dev
-        sudo apt-get clean
```
latk, line 144: Why was this deleted? If it isn't necessary, why was it added previously?

CHANGELOG.rst
```diff
@@ -11,15 +11,20 @@ Known bugs:
 Breaking changes:
 
 New features and notable changes:
+
 - Accept `NAN %` which is used in GCOV 7.5.0 instead of an invalid value. (:issue:`651`)
 - New :option:`--json-base` to define a base bath used in JSON reports. (:issue:`656`)
+- New nox session `bundle_app` to generate a portable application with pyinstaller. (:issue:`661`)
```
latk, line 17: Could link to the cookbook entry here.

Also, remember that RST needs double backticks for code markup :)

noxfile.py
```diff
@@ -221,6 +222,63 @@ def upload_wheel(session: nox.Session) -> None:
...
+    """Bundle a standalone executable."""
+    session.install(
+        "jinja2",
+        "lxml",
+        "pygments",
+        "pyinstaller",
+    )
+    session.install("-e", ".")
+    if not os.path.exists("build"):
+        os.mkdir("build")
```
latk, line 236: Can be simplified as:

```suggestion
    os.makedirs("build", exists_ok=True)
```

noxfile.py
```diff
@@ -221,6 +222,63 @@ def upload_wheel(session: nox.Session) -> None:
...
+                """\
+                import sys
+                import runpy
+
+                if __name__ == "__main__":
+                    runpy.run_module("gcovr", run_name="__main__", alter_sys=True)
+                    sys.exit(0)
+                """
+            )
+        )
```
latk, line 251: If we're creating a temporary file like this, we might as well add it directly as `scripts/pyinstaller_entrypoint.py` if that works.

noxfile.py
latk, line 236: Should add `/build/` to the .gitignore

noxfile.py
```diff
@@ -221,6 +222,63 @@ def upload_wheel(session: nox.Session) -> None:
...
+@nox.session
+def bundle_app(session: nox.Session) -> None:
+    """Bundle a standalone executable."""
+    session.install(
+        "jinja2",
+        "lxml",
+        "pygments",
+        "pyinstaller",
+    )
+    session.install("-e", ".")
```
latk, line 234: Is it necessary to install all these modules? I would guess that `install(".", "pyinstaller")` would be sufficient.

latk (commented):
Ok, I like this. Lots of smaller questions though.

.github/workflows/test.yml
Spacetown: While trying to get pyinstaller working on pypy I found out that this wasn't active since #550. With this PR pypy3 was replaced by pypy-3.7 which deactivates this step.

CHANGELOG.rst
Spacetown: 🙈

noxfile.py
Spacetown: Is already ignored. That's why I reused this folder.

noxfile.py
Spacetown: If I don't use the wheel I need to install all needed modules. On my computer, e.g. jinja isn't installed.

Spacetown: The same as for the tests.

noxfile.py
Spacetown: This I need to do after my holiday. That's impossible on my mobile phone 😄

Spacetown: If we use the wheel and there is an easy way to get the path to gcovr (independent of installation type global, local or venv) this will also work.

Spacetown: @latk I've managed to use the script generated by setuptool. For Windows I need to find a way to extract the script from the executable.

PR #664, opened 2022-08-29
To make heirarchical HTML reports (#160)

macetw:


Spacetown:
Can you add a description and resolve the conflicts?

macetw:
Working on the conflicts. To put this pr on the original repo (and not my fork) was unintentional. I also need to make the build green. I will convert to draft

macetw:
I could use some help from @latk , who rewrote the coverage data model. This is difficult to integrate with my changes, with this commit:
https://github.com/gcovr/gcovr/commit/375cb376081bd5d0e17d410f858a6ac62b4104d6?diff=split

Spacetown:
What is the Problem? Do you need to get the statistics on directory level?

latk:
@macetw I'll be glad to help if you can point out more concrete problems.

The old approach – on which basis you have made your changes – is to aggregate coverage statistics incrementally in the RootInfo object. I found this difficult to work with, for the following reasons:

* There was no clear distinction between intermediate state and final output of the RootInfo object. Mutable state is (unnecessarily) complicated.
* Due to the use of ad-hoc datastructures like dictionaries, type checking (e.g. with IDEs or the mypy tool) was made more difficult.
* A lot of this coverage statistics calculation was duplicated across writers, though sometimes with subtle differences/bugs.

My commit changed this to the following more “functional” workflow:

* Dedicated functions like `SummarizedStats.from_covdata()` query the coverage data model to extract relevant statistics.
* The stats are then formatted for the HTML writer datastructures in one go.

What I would do with your changes:

* Create a DirectoryCoverage class that represents nested DirectoryCoverage + FileCoverage objects. At first approximation, such a class might look like:

  ```python
  @dataclass
  class DirectoryCoverage:
    name: str
    children: Dict[str, Union[DirectoryCoverage, FileCoverage]]
  ```

* Create a (recursive?) function to convert CovData to the DirectoryCoverage tree data structure.
* Implement query functions to gather SummarizedStats from DirectoryCoverage
* Create a (recursive?) function to render pages for the DirectoryCoverage and its children.

By the way, I think that the existing root page can be repurposed as a directory page, without having to create a completely new template.

macetw:
Thanks, @latk . This is very helpful feedback

macetwatargo:
With the given merge conflicts with the latest "master," this Pull Request will not be workable as is. See: https://github.com/gcovr/gcovr/pull/687

Spacetown:
Close because a new PR is opened.

PR #676, opened 2022-10-03
Add test with gtest

Spacetown:
This PR modifies the existing CMake tests to use the `-DCMAKE_BUILD_TYPE=PROFILE` as described in the docs.

In addition a new test is added which use the gtest library from the docker container.


admin/Dockerfile.qa
```diff
@@ -57,6 +58,18 @@ ENV \
 ENV \
   XDG_CACHE_HOME=$NOX_ENV_DIR/.cache
 
+RUN git clone \
+    -n \
+    -q https://github.com/google/googletest.git /tmp/googletest \
+  && cd /tmp/googletest \
+  && git checkout 58d77fa8070e8cec2dc1ed015d66b454c8d78850 \
```
latk, line 65: Can use a shallow clone if we refer to the target commit by its tag name:

```suggestion
RUN git clone -q \
    --branch=release-1.12.1 \
    --depth=1 \
    https://github.com/google/googletest.git /tmp/googletest \
  && cd /tmp/googletest \
```

gcovr/tests/cmake_oos_ninja/CMakeLists.txt
```diff
@@ -1,7 +1,3 @@
-
 cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
 
-project(gcovrtest)
-
-    add_executable(gcovrtest main.cpp)
-
+include(../cmake_oos/CMakeLists.txt)
```
latk, line 3: Is this kind of coupling between tests sensible?

* pro: less duplication, easier maintenance
* con: might accidentally break this test in the future

latk (approved):
Looks like a reasonable test!

Once the tests start passing, of course. We'd have to either install gtest in the Github CI runner and on developer machines, or skip the new tests outside of Docker (see the `GCOVR_ISOLATED_TEST` token that's set in the Dockerfile and used in `test_args.py`)

gcovr/tests/cmake_oos_ninja/CMakeLists.txt
Spacetown: The intention was to reduce the copying of test data and to use a little bit complexer cmake files for testing.

admin/Dockerfile.qa
Spacetown: The suggestion from CMake tutorial was to rely on the commit ID because of security reasons.
I think here we can switch to the shallow clone.

Spacetown:
The test fails because the DTD for cobertura wasn't available. I'll update the test to use a copy of the DTD.

Spacetown:
> Looks like a reasonable test!
> 
> Once the tests start passing, of course. We'd have to either install gtest in the Github CI runner and on developer machines, or skip the new tests outside of Docker (see the `GCOVR_ISOLATED_TEST` token that's set in the Dockerfile and used in `test_args.py`)

I searched for a solution but missed `GCOVR_ISOLATED_TEST`. Is there a easy way to check in `test_gcovr.py` if the lib is a available?

latk:
> Is there a easy way to check in `test_gcovr.py` if the lib is a available?

At least on Linux, can query for the existence of a library with a suitable version like

```bash
pkg-config --atleast-version=1.12.0 gtest
```

Spacetown:
> > Is there a easy way to check in `test_gcovr.py` if the lib is a available?
> 
> At least on Linux, can query for the existence of a library with a suitable version like
> 
> ```shell
> pkg-config --atleast-version=1.12.0 gtest
> ```

Since there is no command for Windows I'll rely on `GCOVR_ISOLATED_TEST`.

admin/Dockerfile.qa
```diff
@@ -18,17 +18,20 @@ ARG CXX
...
-                       make \
-                       $CC \
-                       $CXX \
-                       python3-setuptools \
-                       $(if [ "$DOCKER_OS" != "ubuntu:22.04" ]; then echo python3.7; fi) \
-                       ninja-build \
-                       curl \
-                       libxml2-utils && \
-    apt-get clean
+    export DEBIAN_FRONTEND=noninteractive&& apt-get update \
```
cole701, line 21: Does not follow the same pattern as the follow on commands. It would be cleaner if: 
export DEBIAN_FRONTEND=noninteractive \
&& apt-get update \

PR #685, opened 2022-10-24
Issue 684: Fix sorting of html reports when there are symlinks.

### LINKED ISSUE #684: html: symlinked file sorted based on symlink name, not on relative name which is shown in report.
**Describe the bug**
As seen in the test case `linked-html`, we show the file "A/C/D/file6.cpp" in the html report, but we sort it based on it's other symlink name. We sort with the text "m/n/D/file6.cpp" not sorted by what is shown.

This also shows to be not sorted right in the text report. In that case, the symlink text is showing. (a separate bug?)

**To Reproduce**
1. Observe the test case, linked-html, the reference is shown in a way that is not alphabetical.

**Expected behavior**
files with symlinks with relative paths should be sorted alphabetically by filename with the text shown in the report.

**Screenshots**
https://htmlpreview.github.io/?https://github.com/gcovr/gcovr/blob/28e5c88e11fee50856f1347110a8da6153a46503/gcovr/tests/linked/reference/gcc-8/coverage.html

 - GCC version [e.g. MinGW 8.1.0]
 - GCOVR version [e.g. 4.2]
 - Project directory layout:
   - Roots
   - Object directories

**Additional context**
This was discovered while working on my fix for cascaded-html reporting. I changed how the files got sorted to be in jinja (in the very last phase of report rendering) and so I was inadvertently fixing this sort.

macetw:
This change fixes the sorting of html reports, for when there are symlinks within the nested directory structure. The sorting of filename will consider the absolute paths, which is already what it renders in the report with.

Note, since we only do the relative-path trick for the html report, we only make this sort-fix for the html report. The other reports are still showing the alternate path, as seen in the linked-txt test report.

Closes #684

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -68,6 +68,7 @@ The following developers contributed to gcovr (ordered alphabetically):
     TFA-N,
     Thibault Gasc,
     Tilo Wiedera,
+    Tyler W. Mace,
     trapzero,
     Will Thompson,
     William Hart,
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,8 +24,8 @@ New features and notable changes:
 Bug fixes and small improvements:
 
 - Fix :option:`--html-tab-size` feature. (:issue:`650`)
-
 - Do not ignore returncode of `gcov`. (:issue:`653`)
+- Fix alphabetical sort of html report, for when there are symlinks. (:issue:`685`)
 
 Documentation:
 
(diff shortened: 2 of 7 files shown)


latk (approved):
LGTM once all tests pass. Thanks for finding and fixing this :)

Spacetown:
@macetw Are you using docker for your tests?

Spacetown (approved):
LGFM

macetw:
> @macetw Are you using docker for your tests?

I am now, but I wasn't before. I was able to reproduce the failing test cases and fix them with my latest force-push.

Spacetown:
Thanks for fixing this bug.

macetwatargo:
> Thanks for fixing this bug.
You betcha. More to come.

PR #686, opened 2022-10-24
Detect compiler version from gcc output if not specified.

Spacetown:
This PR selects the latest GCC version on the system for the tests. If only `gcc` is available the version is detected from the output of `gcc --version`.

If using `qa_compiler` session the default is switched from `gcc-5` to `gcc-11` (the current latest version for the reference data).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,7 +24,6 @@ New features and notable changes:
 Bug fixes and small improvements:
 
 - Fix :option:`--html-tab-size` feature. (:issue:`650`)
-
 - Do not ignore returncode of `gcov`. (:issue:`653`)
 
 Documentation:
@@ -33,6 +32,7 @@ Internal changes:
 
 - Select the :option:`--html-theme` using CSS classes. (:issue:`650`)
 - Change and extend ``cmake`` tests. (:issue:`676`)
+- Detect the verion of the ``gcc``. (:issue:`686`)
 
 5.2 (06 August 2022)
 --------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -32,7 +32,7 @@ Internal changes:
 
 - Select the :option:`--html-theme` using CSS classes. (:issue:`650`)
 - Change and extend ``cmake`` tests. (:issue:`676`)
-- Detect the verion of the ``gcc``. (:issue:`686`)
+- Detect ``gcc`` version for running tests. (:issue:`686`)
 
 5.2 (06 August 2022)
 --------------------
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -144,8 +144,8 @@ How to set up a development environment
 ---------------------------------------
 
 For working on gcovr, you will need a supported version of Python 3,
-GCC version 5, 6 or 8 (other GCC versions are supported by gcovr,
-but will cause spurious test failures) or clang version 10, ``make``,
+GCC version 5, 6, 8, 9, 10 and 11 (other GCC versions are supported by gcovr,
+but will cause spurious test failures) or clang version 10 and 13, ``make``,
 ``cmake`` and ``ninja``.
 Please make sure that the tools are in the system ``PATH``.
 On **Windows**, you will need to install a GCC toolchain as the
@@ -168,11 +168,12 @@ is needed.
     -  clang-13/clang++-13/llvm-cov
 
     are available everything is OK.
-    For gcc-6, gcc-8, gcc-9, gcc-10, gcc-11, clang-10 and clang-13 you should use the
-    option ``CC=...`` see :ref:`run and filter tests <run tests>`.
-    If this isn't OK you
-    can set the reference data to use by setting the environment ``CC_REFERENCE=gcc-8``
-    or you have to create symlinks for the gcc executables with the following steps.
+    The test suite uses the newest GCC found in the PATH. To use another one you
+    need to set the environment ``CC=...`` see
+    :ref:`run and filter tests <run tests>`.
+    If you only have ``gcc`` in your path the version is detected to select the
+    correct reference.
+    You can also create symlinks for the gcc executables with the following steps.
     You can check the GCC version with gcc --version. If the output says
     version 8, you should also be able to run gcc-8 --version. Your Linux
     distribution should have set all of this up already.
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -144,8 +144,8 @@ How to set up a development environment
 ---------------------------------------
 
 For working on gcovr, you will need a supported version of Python 3,
-GCC version 5, 6, 8, 9, 10 and 11 (other GCC versions are supported by gcovr,
-but will cause spurious test failures) or clang version 10 and 13, ``make``,
+GCC version 5, 6, 8, 9, 10 or 11 (other GCC versions are supported by gcovr,
+but will cause spurious test failures) or clang version 10 or 13, ``make``,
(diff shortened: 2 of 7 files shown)


CHANGELOG.rst
```diff
@@ -33,6 +32,7 @@ Internal changes:
 
 - Select the :option:`--html-theme` using CSS classes. (:issue:`650`)
 - Change and extend ``cmake`` tests. (:issue:`676`)
+- Detect the verion of the ``gcc``. (:issue:`686`)
```
latk, line 35: I'd provide context that this version detection only applies to tests, not to user-visible behaviour.

```suggestion
- Detect ``gcc`` version for running tests. (:issue:`686`)
```

CONTRIBUTING.rst
```diff
@@ -144,8 +144,8 @@ How to set up a development environment
 ---------------------------------------
 
 For working on gcovr, you will need a supported version of Python 3,
-GCC version 5, 6 or 8 (other GCC versions are supported by gcovr,
-but will cause spurious test failures) or clang version 10, ``make``,
+GCC version 5, 6, 8, 9, 10 and 11 (other GCC versions are supported by gcovr,
+but will cause spurious test failures) or clang version 10 and 13, ``make``,
```
latk, line 148: While ideally someone would run all the tests, it is sufficient to have any one of these compilers installed.

```suggestion
GCC version 5, 6, 8, 9, 10 or 11 (other GCC versions are supported by gcovr,
but will cause spurious test failures) or clang version 10 or 13, ``make``,
```

noxfile.py
```diff
@@ -49,6 +50,43 @@
...
+def get_gcc_version_to_use():
+    CC = os.environ.get("CC")
+    if CC is None:
+        for gcc in reversed([gcc for gcc in GCC_VERSIONS]):
+            gcc = shutil.which(gcc)
+            if gcc is not None:
+                CC = os.path.split(gcc)[1]
+                break
+        else:
+            CC = shutil.which("gcc")
```
latk, line 62: I think I've mentioned this previously, that I have a personal preference for early returns. Of course you're not required to rewrite the function, but it could make the data flow clearer:

```python
def get_gcc_verson_to_use() -> str:
    # If the user explicitly set CC variable, use that directly without checks.
    cc = os.environ.get("CC")
    if cc:
        return cc

    # find the first installed compiler version that we support
    for cc in GCC_VERSIONS:
        if shutil.which(cc):
            return cc

    # check if there's a "gcc" or "clang" binary with a suitable version
    for cc in ["gcc", "clang"]:
        ...  # sniff version number from --version message
```

Remarks in that context:

* why reverse the `GCC_VERSIONS` list?
* I think that `gcc` will always equal `os.path.split(shutil.which(gcc))[1]`, unless symlinks are involved. But even in the case of symlinks, we'd rather have the exact name from `GCC_VERSIONS`.

noxfile.py
```diff
@@ -49,6 +50,43 @@
...
+            # warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+            output = subprocess.check_output([CC, "--version"]).decode().split("\n")
+            if output[0].startswith("gcc"):
+                version = output[0].split(" ")[-1]
+                major_version = version.split(".")[0]
+                CC = f"gcc-{major_version}"
+            elif " clang version " in output[0]:
+                version = output[0].split(" clang version ")[1]
+                major_version = version.split(".")[0]
+                CC = f"clang-{major_version}"
```
latk, line 80: I think this could be more readable with a regex (famous last words).

```suggestion
            # look for a line "gcc WHATEVER VERSION.WHATEVER"
            search_gcc_version = re.search(r"^gcc\b.* ([0-9]+)\.\S+$", output, re.M)
            
            # look for a line "WHATEVER clang version VERSION.WHATEVER"
            search_clang_version = re.search(r"\bclang version ([0-9]+)\.")
            
            if search_gcc_version:
                major_version = search_gcc_version.group(1)
                CC = f"gcc-{major_version}"
            elif search_clang_version:
                major_version = search_clang_version.group(1)
                CC = f"clang-{major_version}"
```

latk (approved):
Thank you for working on this! A couple of minor comments though.

Do the changes to the compiler selection interact with the set_environment() function?

FYI, the [Github Actions logs](https://github.com/gcovr/gcovr/actions/runs/3315750815) are full of deprecation warnings, we should eventually address those. This is actually partially related to the GCC version problem since the Actions rely on some really ancient Ubuntu versions in order to install some compilers.

noxfile.py
Spacetown: Ok, but than we have to split in each return because the user can set the compiler with a full path.
The reverse and the `for cc in ["gcc", "clang"]:` was because the old list was sorted ascending. This can be removed now as I changed the order.

noxfile.py
```diff
@@ -20,22 +20,24 @@
...
-    "gcc-8",
-    "gcc-9",
-    "gcc-10",
     "gcc-11",
-    "clang-10",
+    "gcc-10",
+    "gcc-9",
+    "gcc-8",
+    "gcc-6",
+    "gcc-5",
```
macetwatargo, line 37: No more numerical order?

noxfile.py
```diff
@@ -282,7 +325,7 @@ def docker_container_id(session: nox.Session, version: str) -> str:
 @nox.session(python=False)
 def docker_qa_build(session: nox.Session) -> None:
     """Build the docker container for the default GCC version."""
-    session_id = f"docker_qa_build({GCC_VERSION2USE})"
+    session_id = f"docker_qa_build({GCC_VERSIONS[0]})"
```
macetwatargo, line 328: Nice.

noxfile.py
```diff
@@ -64,7 +107,7 @@ def set_environment(session: nox.Session, cc: str, check: bool = True) -> None:
 @nox.session(python=False)
 def qa(session: nox.Session) -> None:
     """Run the quality tests for the default GCC version."""
-    session_id = f"qa_compiler({GCC_VERSION2USE})"
+    session_id = f"qa_compiler({get_gcc_version_to_use()})"
```
macetwatargo, line 110: Nice.

macetwatargo (approved):

noxfile.py
Spacetown: Changed from Ascending to descending to get the latest supported gcc by accessing item 0.

Spacetown:
The following warnigng is not resolvable yet since PR https://github.com/seanmiddleditch/gha-setup-ninja/pull/15 is still open.
```
Node.js 12 actions are deprecated. For more information see: 
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. 
Please update the following actions to use Node.js 16: seanmiddleditch/gha-setup-ninja
```

Spacetown:
@latk Most warnings are fixed but the following warning I don't understand:

```
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files.
For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
``` 

We don't use `set-output` we only use the [environment files](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files) linked in the information linked above. In this documentation following syntax is used:

```bash
echo "VARIABLE=MyValue" >> $GITHUB_ENV
```

But we use a different quoting since a long time:

```bash
echo VARIABLE="MyValue" >> $GITHUB_ENV
```

With the quoting from the documentation it doesn't work. The environment isn't set.

PR #687, opened 2022-10-25
    Issue 160: Cascaded-directory html reports

### LINKED ISSUE #544: HTML index by folder
# HTML report folder navigation

`lcov` lets me navigate the coverage report like a file system, showing coverage for both files and folders.

I'd like to see the detailed html report organized by file _and_ by folder

---

For instance, with a file system like this

```
source/
├── some_folder
│   ├── file_one.cpp
│   └── file_two.cpp
└── some_other_folder
    └──   file_three.cpp
```

Currently, with the command:

```
gcovr --root source/ --html-details coverage.html
```

You get a report which looks like this

current

---

I'm proposing this be broken down into its subfolders, `some_folder` and `some_other_folder`. This way, the index would show:

proposed-folder

And clicking on `some_folder` would show the index for that folder:

proposed-file

Thoughts?

macetw:
This change implements a hierarchical (aka cascaded) directory-based html report. It adds a new argument: --html-cascaded-directories

This change re-implements how we sort the lists of source files in html, that it uses Jinja to sort, using data stored with what we pass to the template. This way, the sorting can also take into account the subdirectories that have aggregated statistical data that goes with it.

This change needed to implement a new class that manages different directories and the source files they go in. This class also knows about the parent directory that it is inside and the statistics for all of the given directory's contents.

#160

Closes #544


Spacetown:
For the Windows specific data you have to download the artefacts and override the files locally.

macetwatargo:
> For the Windows specific data you have to download the artefacts and override the files locally.

I wish I had a windows box to practice on. :-/ I'll have to come up with something.

gcovr/tests/cascaded-nonsort/Makefile
```diff
@@ -0,0 +1,34 @@
...
+	rm -rf ./subdir
+	rm -f *.gc* */*.gc* */*/*.gc* */*/*/*.gc* */*/*/*/*.gc*
+	rm -f *.o */*.o */*/*.o */*/*/*.o */*/*/*/*.o
+	rm -f coverage*.*
+
+copy:
+ifeq ($(filter $(BASE_OS),MSYS_NT MINGW64_NT),)
+	cp -fr ../nested/subdir .
+else
+	COPY /Y ../nested/subdir subdirs
```
macetw, line 33: These lines are not running windows-style in the unit tests.

gcovr/coverage.py
```diff
@@ -314,6 +318,110 @@ def call_coverage(self) -> CoverageStat:
...
+
+    @staticmethod
+    def directory_root(subdirs: CovData_subdirectories, root_filter: re.Pattern) -> str:
+        key = next(iter(subdirs))
+        while True:
+            next_key = DirectoryCoverage.directory_key(key, root_filter)
+            if not next_key:
+                return key
+            else:
+                key = next_key
```
Spacetown, line 419: The error on Windows is here but I need to investigate deeper what you're doing here.

Spacetown: The iterator is at the end and the exception isn't caught.

gcovr/tests/cascaded-nonsort/Makefile
Spacetown: Sorry, I don't understand this.

Spacetown:
Please don't force push. After this we don't see the changes since last review.

gcovr/templates/directory_page.html
```diff
@@ -139,7 +139,7 @@
     {% endif %}
   </tr>
 
-  {% for row in info.files %}
+  {% for row in info.files|sort(attribute=row_sort_key, reverse=row_sort_reverse) %}
```
Spacetown, line 142: Add spaces to the filter for better readability.

Spacetown (changes requested):

gcovr/writer/html.py
```diff
@@ -367,34 +384,47 @@ def print_html_report(covdata: CovData, output_file, options):
...
+            covdata, sorted_keys, options.root_filter
+        )
+        DirectoryCoverage.collapse_subdirectories(
+            root_info.subdirs, options.root_filter
+        )
+
     cdata_fname = {}
     cdata_sourcefile = {}
-    for f in keys:
+    for f in list(covdata.keys()) + list(root_info.subdirs.keys()):
```
Spacetown, line 413: This list isn't sorted anymore.

gcovr/coverage.py
```diff
@@ -314,6 +320,112 @@ def call_coverage(self) -> CoverageStat:
...
+                subdirs, filecov.filename, filecov, root_filter
+            )
+        return subdirs
+
+    @staticmethod
+    def directory_key(filename: str, root_filter: re.Pattern):
+        key = os.path.dirname(filename.replace("\\", "/"))
+        if root_filter.search(key + "/") and key != filename:
+            return key
+        return None
```
Spacetown, line 411: USe `SEPARATOR` instead of `"/"`. Add a trailing `SEPARATOR` to value of `key` to show clearly that it's a directory.

gcovr/writer/html.py
```diff
@@ -367,34 +384,47 @@ def print_html_report(covdata: CovData, output_file, options):
...
-        dirs.append(os.path.dirname(filtered_fname) + os.sep)
         cdata_fname[f] = filtered_fname
         if options.html_details:
-            cdata_sourcefile[f] = _make_short_sourcename(output_file, filtered_fname)
+            if os.path.normpath(f) == os.path.normpath(options.root_dir):
+                cdata_sourcefile[f] = output_file
+            else:
+                cdata_sourcefile[f] = _make_short_sourcename(
+                    output_file, filtered_fname
+                )
```
Spacetown, line 423: Why was this changed?

Spacetown (changes requested):
Looks good, only some little comments.

Some general questions:
- Shall directories always be at the top or sorted between the files? Reflect this in the test.
- What shall be done if only one file is in a directory? Is the directory needed in this case or shall the file be added directly, removing one hierarchy level?

setup.cfg
```diff
@@ -51,6 +51,7 @@ exclude_lines =
 max-line-length = 88
 ignore = E203, E501, W503
 exclude = .git, __pycache__, gcovr/tests/cmake_gtest/build/_deps
+builtins = Options
```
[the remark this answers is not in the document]
macetw: This is needed for proper type-defing of "Options" within gcovr/coverage.py. It seems I cannot import this from __main__.py, so I must tell flake that it's a valid class. Interested in other ways to do this properly, that would work with flake8 (or mypy)

gcovr/coverage.py
```diff
@@ -314,6 +320,112 @@ def call_coverage(self) -> CoverageStat:
...
+    parent_key: str
+
+    @classmethod
+    def new_empty(cls) -> DirectoryCoverage:
+        return cls("", SummarizedStats.new_empty(), list(), "")
+
+    @staticmethod
+    def add_directory_coverage(
+        subdirs: CovData_subdirectories,
+        filename: str,
```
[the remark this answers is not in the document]
macetw: No, this is not duplicate info, but i can see why one might interpret that. It is a location to process. Because this is a recursive function, we have to manage the directory location we are putting this statistic within. If the statistic is within A/B/C/D/fileE.cpp, we need to run this function as A/B/C/D/fileE.cpp first, then A/B/C/D, then A/B/C, then A/B, then A, then again at the top level directory. (which is a blank key, or a single Separator-i-forget)

Given the ambiguity, I'll change the name to location and hope that's the best. If not, let me know.

macetw:
> Looks good, only some little comments.
> 
> Some general questions:
> 
> * Shall directories always be at the top or sorted between the files? Reflect this in the test.
> * What shall be done if only one file is in a directory? Is the directory needed in this case or shall the file be added directly, removing one hierarchy level?

1) The directories will always be sorted based on the sorting method the user supplied. See function row_sort_keys(), which respects this. This translates into the sorting style applied at the Jinja level.
https://github.com/macetw/gcovr/blob/cdce979872c61b89363e997e8b6bf1f5dcc132aa/gcovr/writer/html.py#L612
https://github.com/macetw/gcovr/blob/cdce979872c61b89363e997e8b6bf1f5dcc132aa/gcovr/templates/directory_page.html#L142

2) If there is only one file in a directory ... or only one directory in a directory, these conditions cause the hierarchy to be collapsed. It is tested on the tests I put in. The "nested" test case that was already there exercises this, and I tested different sort methods on the cascaded case.
https://github.com/macetw/gcovr/blob/cdce979872c61b89363e997e8b6bf1f5dcc132aa/gcovr/coverage.py#L364
https://github.com/macetw/gcovr/tree/hierarchical_html_new_data_model/gcovr/tests/nested/subdir/A/C/D
https://github.com/macetw/gcovr/tree/hierarchical_html_new_data_model/gcovr/tests/nested/subdir/B

gcovr/writer/html.py
macetw: That's right, it's not sorted any more. We cannot rely on sorting at this level when managing files deeper in the directory hierarchy.

For example, not all data for all subdirectories are aggregated at this point. For example, you can't sort data for files like "A/B/fileC.cpp," when "A/B/C/fileD.cpp" hasn't been set up yet for the entry for "A/B/C"

Sorting is now handled at the Jinja level, since we are prepared with this information for Jinja rendering. Jinja has all the data needed to sort correctly. I would suggest using Jinja based sorting for the other formats also, for the future.

gcovr/writer/html.py
```diff
@@ -233,33 +236,46 @@ def set_coverage(self, covdata: CovData) -> None:
...
+        stats = (
+            cdata.stats
+            if isinstance(cdata, DirectoryCoverage)
+            else SummarizedStats.from_file(cdata)
+        )
 
         lines = {
             "total": stats.line.total,
             "exec": stats.line.covered,
+            "uncovered": stats.line.uncovered,
```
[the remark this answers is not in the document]
macetw: This is needed now that Jinja is sorting.

gcovr/writer/html.py
Spacetown: > Sorting is now handled at the Jinja level, since we are prepared with this information for Jinja rendering. Jinja has all the data needed to sort correctly. I would suggest using Jinja based sorting for the other formats also, for the future.

As I wrote in https://github.com/gcovr/gcovr/pull/687#discussion_r1008750922, I think the sorting shall be changed in a separate PR. Also the sorting is changed here since you implemented a multilevel sort and changed the order if `--sort-percentage` is given.
https://github.com/gcovr/gcovr/blob/cdce979872c61b89363e997e8b6bf1f5dcc132aa/gcovr/writer/html.py#L612-L623

In general it's better to sort the keys even if it's not relevant for the result. If there is a problem on several files it's always better to abort on the same key than on a random key.

gcovr/writer/html.py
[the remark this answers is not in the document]
Spacetown: Why moving the sorting to Jinja level? Prefer to do this in python as for all writers.

gcovr/writer/html.py
```diff
@@ -367,34 +384,47 @@ def print_html_report(covdata: CovData, output_file, options):
         data["css_link"] = css_link
 
     data["theme"] = options.html_theme
+    data["row_sort_key"], data["row_sort_reverse"] = row_sort_keys(options)
```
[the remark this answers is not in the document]
macetw: There's a lot in this PR that would be part of the other PR. I'll do my best to differentiate those changes to a new PR.

gcovr/coverage.py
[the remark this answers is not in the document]
Spacetown: Is the `filename` really needed? When the recursion is entered the `newchield` is None. What about using:
```python
    @staticmethod
    def add_directory_coverage(
        subdirs: CovData_subdirectories,
        root_filter: re.Pattern,
        filecov: FileCoverage,
        dircov: DirectoryCoverage = None,
    ) -> None:
        if dircov is None:
            key = DirectoryCoverage.directory_key(filecov.filename, root_filter)
            filecov.parent_key = key
        else:
            key = DirectoryCoverage.directory_key(dircov.dirname, root_filter)
            dircov.parent_key = key
        if key:
            if key not in subdirs:
                subdir = DirectoryCoverage.new_empty()
                subdir.dirname = key
                subdirs[key] = subdir

            if dircov is None:
                subdirs[key].children.append(filecov)
            else:
                if dircov not in subdirs[key].children:
                    subdirs[key].children.append(dircov)

            subdirs[key].stats += SummarizedStats.from_file(filecov)
            DirectoryCoverage.add_directory_coverage(
                subdirs, root_filter, filecov, subdirs[key]
            )
```

The normalizing of the seperator (`force_unix_separator(arg)`) should be done in the `__init__` functions of the `FileCoverage` and the `DirectoryCoverage`.

setup.cfg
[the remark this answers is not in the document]
Spacetown: I think it's better to move the class to utils because it doesn't have a dependency.
This will also remove the error which is shown by Pylance in VSCode.

gcovr/coverage.py
```diff
@@ -35,9 +35,14 @@
...
 
 from __future__ import annotations
-from typing import Dict, Optional, TypeVar, Union
+from collections import OrderedDict
+import os
+import re
+from typing import Dict, Final, List, Iterable, Optional, TypeVar, Union
 from dataclasses import dataclass
 
+SEPARATOR: Final[str] = "/"
```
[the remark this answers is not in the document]
Spacetown: As I saw now, in `utils.py` we have a `force_unix_separator` which should be used instead of a `replace`.

gcovr/writer/html.py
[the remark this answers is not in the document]
macetw: Because we cannot sort at the time we need to, for comparing statistics of files with directories, when directories store an aggregate of files underneath it.

Why prefer this in python? Why not do this in Jinja for all writers. In this PR, we pass in to the template the field we sort on and the reverse-or-nonreverse sort approach and we can do it all. Jinja knows all the data and can sort, so why do you prefer python, more than just "that's the way we always do it." ??

Note there's still an issue with the relative-path used with the txt writer. See "Note" I left in the linked PR, that we should just make a list of datums, not worry about sorting in Python, just worry about the tabular data, and have jinja do the sort.
https://github.com/gcovr/gcovr/pull/685

gcovr/writer/html.py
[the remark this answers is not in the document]
Spacetown: In my opinion it's not needed to move the sorting to the Jinja level and change the kind of sorting. Removing this would remove a little bit of complexity.

Changeing the sorting to a multilevel sorting in general can be done in a separate PR.

gcovr/writer/html.py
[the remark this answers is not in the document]
macetw: Another reason: We are adding either DirectoryCoverage or we are adding FileCoverage. Those datas come from different datasets that we cannot sort them in one place, lest we create a more complex keying system. We now must call add_row with the different types in a place that isn't sorted:
https://github.com/macetw/gcovr/blob/45cb8e5c6fc94aa789ff909972d8fd73a3f3465e/gcovr/writer/html.py#L658

We could add a lambda sort here, going over this list. This feels more complicated to me.

gcovr/writer/html.py
macetw: (... I don't think you meant to use the word "abort" here, did you? )

gcovr/writer/html.py
macetw: The "output_file" parameter needs to be preserved for the top-most directory report we need to give. Otherwise, we need to manage a regular _make_short_sourcename.

gcovr/writer/html.py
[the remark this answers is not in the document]
macetw: This would be needed for sorting at Jinja level.

gcovr/writer/html.py
```diff
@@ -408,34 +438,79 @@ def print_html_report(covdata: CovData, output_file, options):
...
     root_info.set_directory(root_directory)
 
-    for f in keys:
-        root_info.add_file(covdata[f], cdata_sourcefile[f], cdata_fname[f])
+    (output_prefix, output_suffix) = os.path.splitext(os.path.abspath(output_file))
+    if output_suffix == "":
+        output_suffix = ".html"
+    functions_output_file = f"{output_prefix}.functions{output_suffix}"
+    data["FUNCTIONS_FNAME"] = os.path.basename(functions_output_file)
+
```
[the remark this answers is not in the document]
macetw: We should explore: I might be introducing a bug about including that file when details are not requested. I'm not testing that iteration: I'm always requesting details in the tests.

gcovr/writer/html.py
Spacetown: Abort as a result of an exception or coding error.

gcovr/writer/html.py
Spacetown: In the old version `f` was always a source file. Can it now also be the root directory?

macetwatargo: Yes. Or any subdirectory. They are each entries with their own statistics.
(exception, if there are directories that have only 1 entry they are collapsed for easier navigation)

Spacetown: Ok. In this case also `filtered_fname` should be an empty string.
If this isn't true because of the trailing slash in the `options.root_filter` this should be handled in line 414.
Here we can then rely on the value of `filtered_fname`.

gcovr/writer/html.py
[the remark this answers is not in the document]
macetw: FYI, another likely feature request is to sort dynamically by the user in Javascript (or similar), where a user could sort on the UI by any field.

macetw:
> > Looks good, only some little comments.
> > Some general questions:
> > 
> > * Shall directories always be at the top or sorted between the files? Reflect this in the test.
> > * What shall be done if only one file is in a directory? Is the directory needed in this case or shall the file be added directly, removing one hierarchy level?
> 
> 1. The directories will always be sorted based on the sorting method the user supplied. See function row_sort_keys(), which respects this. This translates into the sorting style applied at the Jinja level.
>    https://github.com/macetw/gcovr/blob/cdce979872c61b89363e997e8b6bf1f5dcc132aa/gcovr/writer/html.py#L612
>    https://github.com/macetw/gcovr/blob/cdce979872c61b89363e997e8b6bf1f5dcc132aa/gcovr/templates/directory_page.html#L142
> 2. If there is only one file in a directory ... or only one directory in a directory, these conditions cause the hierarchy to be collapsed. It is tested on the tests I put in. The "nested" test case that was already there exercises this, and I tested different sort methods on the cascaded case.
>    https://github.com/macetw/gcovr/blob/cdce979872c61b89363e997e8b6bf1f5dcc132aa/gcovr/coverage.py#L364
>    https://github.com/macetw/gcovr/tree/hierarchical_html_new_data_model/gcovr/tests/nested/subdir/A/C/D
>    https://github.com/macetw/gcovr/tree/hierarchical_html_new_data_model/gcovr/tests/nested/subdir/B

Update:
With the most recent commits, the row_sort_keys() function no longer exists, but the sort_coverage() function is made to sort using the same strategies for files as for directories. Note that the line_coverage() and branch_coverage() functions with DirectoryCoverage had to be implemented but are simple wrapper functions.

gcovr/writer/html.py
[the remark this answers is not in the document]
Spacetown: In this case the sort must be implemented in the javascript and not in the jinja2 template. This has to be checked when this feature is going to be implemented.

gcovr/coverage.py
```diff
@@ -398,22 +437,21 @@ def from_covdata(
...
+        key = os.path.dirname(filename.replace("\\", GCOVR_PATH_SEPARATOR))
+        if root_filter.search(key + GCOVR_PATH_SEPARATOR) and key != filename:
             return key
         return None
 
     @staticmethod
     def directory_root(subdirs: CovData_subdirectories, root_filter: re.Pattern) -> str:
         if not subdirs:
-            return SEPARATOR
+            return GCOVR_PATH_SEPARATOR
```
Spacetown, line 454:
```suggestion
            return "/"
```

gcovr/coverage.py
```diff
@@ -398,22 +437,21 @@ def from_covdata(
...
-            )
+            DirectoryCoverage.add_directory_coverage(subdirs, root_filter, filecov)
         return subdirs
 
     @staticmethod
     def directory_key(filename: str, root_filter: re.Pattern):
-        key = os.path.dirname(filename.replace("\\", "/"))
-        if root_filter.search(key + "/") and key != filename:
+        filename = filename.replace("\\", GCOVR_PATH_SEPARATOR)
+        key = os.path.dirname(filename.replace("\\", GCOVR_PATH_SEPARATOR))
```
Spacetown, line 446:
```suggestion
        filename = force_unix_separator(filename)
        key = os.path.dirname(filename)
```

gcovr/coverage.py
```diff
@@ -38,10 +38,10 @@
 from collections import OrderedDict
 import os
 import re
-from typing import Dict, Final, List, Iterable, Optional, TypeVar, Union
+from typing import Dict, Iterable, Optional, TypeVar, Union
 from dataclasses import dataclass
 
-SEPARATOR: Final[str] = "/"
+GCOVR_PATH_SEPARATOR: str = "/"
```
Spacetown, line 44:
```suggestion
```

gcovr/utils.py
```diff
@@ -26,7 +26,7 @@
 import sys
 from contextlib import contextmanager
 
-from .coverage import CovData, CoverageStat, SEPARATOR
+from .coverage import CovData, CoverageStat, GCOVR_PATH_SEPARATOR
```
Spacetown, line 29: The dependency from utils to coverage must be removed to be able to import the `force_unix_separator` in coverage.py. I think the function `sort_coverage` is better implemented in coverage.py than in utils.py.

```suggestion
GCOVR_PATH_SEPARATOR = "/"
```

gcovr/utils.py
```diff
@@ -70,7 +70,7 @@ def realpath(path):
 
 
 def get_os_independent_path(path):
-    return path.replace(os.path.sep, SEPARATOR)
+    return path.replace(os.path.sep, GCOVR_PATH_SEPARATOR)
```
Spacetown, line 73: This function does the same than `force_unix_separator` but is only used here. Can you remove this function and replace the usage with `def sort_coverage`?

Spacetown (changes requested):
Some suggestions regarding path normalization.

gcovr/tests/cascaded-nonsort/Makefile
macetw: This is tested now to be working correctly.

gcovr/coverage.py
macetw: This line is now gone.

gcovr/utils.py
macetw: This is no longer part of the diff.

gcovr/coverage.py
macetw: This should return os.sep, for proper support on Windows.

macetw:
> > Looks good, only some little comments.
> > Some general questions:
> > 
> > * Shall directories always be at the top or sorted between the files? Reflect this in the test.
> > * What shall be done if only one file is in a directory? Is the directory needed in this case or shall the file be added directly, removing one hierarchy level?
> 
> 1. The directories will always be sorted based on the sorting method the user supplied. See function row_sort_keys(), which respects this. This translates into the sorting style applied at the Jinja level.
>    https://github.com/macetw/gcovr/blob/cdce979872c61b89363e997e8b6bf1f5dcc132aa/gcovr/writer/html.py#L612
>    https://github.com/macetw/gcovr/blob/cdce979872c61b89363e997e8b6bf1f5dcc132aa/gcovr/templates/directory_page.html#L142
> 2. If there is only one file in a directory ... or only one directory in a directory, these conditions cause the hierarchy to be collapsed. It is tested on the tests I put in. The "nested" test case that was already there exercises this, and I tested different sort methods on the cascaded case.
>    https://github.com/macetw/gcovr/blob/cdce979872c61b89363e997e8b6bf1f5dcc132aa/gcovr/coverage.py#L364
>    https://github.com/macetw/gcovr/tree/hierarchical_html_new_data_model/gcovr/tests/nested/subdir/A/C/D
>    https://github.com/macetw/gcovr/tree/hierarchical_html_new_data_model/gcovr/tests/nested/subdir/B

Since I wrote this, I changed the sorting method to run python-based sort_coverage() even within the subdirectory level, based on feedback on this PR. I think there's still a future improvement that a client-user would like to sort based on any column, that it would be javascript/typescript-based sort, but that's just for future-thinking.

gcovr/utils.py
Spacetown: But it is tied to the coverage classes defined in coverage.py. It could be a member function of `CovData`.

gcovr/coverage.py
Spacetown: No, it is described in the documents that we use Unix seperators also on Windows.

macetwatargo: If this was true, a force_unix_separators() call would enclose over this line:
https://github.com/gcovr/gcovr/blob/7111438d5cfdced67dd46c99b2dab7caede0e62f/gcovr/__main__.py#L257
... and on to the corresponding root_filter set on line 275 of __main__.py. Should this also be force_unix_separators() ?

macetwatargo: Instead, we only force-unix-separators at the time we display on reports and when we sort. Not for managing the data.

macetwatargo: Since my changes manage filenames to sort through directory management, I need to standardize use of directory-separators how they are managed with the root_filter. If these filenames are coming from os.path.* functions, I also need to use os.sep.

gcovr/coverage.py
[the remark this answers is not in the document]
Spacetown: > The normalizing of the seperator (`force_unix_separator(arg)`) should be done in the `__init__` functions of the `FileCoverage` and the `DirectoryCoverage`.

@macetw What do you think about this? Than the writer can rely on unix separators and it would be much clearer.

gcovr/coverage.py
Spacetown: See my comment above about normalizing to unix separators in the data model. But after checking the point with the `root_filter` this should be done in a separate PR.

latk:
Naming command line options is the kind of thing where we only have one attempt to get it right – once some behaviour is implemented, people *will* depend on that for the next half decade or so.

I agree with @Spacetown that cascaded-directories will typically imply detail-pages.

I expect that this option will be used quite frequently, and will replace `--html-details` as the typical way to get HTML output. Thus, the command line options to get a cascaded report should be short and easy to type.

My preferred solution would be three levels of detail in the HTML report, with each level including the content of the previous levels:

* `--html` for a summary of the entire project in a single HTML file
* `--html-details` for that summary plus annotated per-source-file reports
* `--html-nested` for per-directory summaries plus annotated per-source-file reports, intended for large projects where summarizing all coverage data in a single HTML file would be overwhelming

If we want to enable nested summaries without per-source-file annotated reports, I'd rather do this via something like `--html-nested --no-html-details`, but I'm not sure if our configuration data model can express that (since these options aren't boolean flags, but take a destination filename).

gcovr/writer/html.py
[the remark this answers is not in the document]
Spacetown: Since we now use option `--html-nested` the details are always present.

gcovr/utils.py
Spacetown: I moved the function to `coverage.py`.

Spacetown (approved):

Spacetown:
@macetw I've pushded the open changes on your branch. If you don't have pending changes I can merge the PR.

macetw:
For what it's worth, these additional changes by @Spacetown look good. Thank you.

macetw:
> @macetw I've pushded the open changes on your branch. If you don't have pending changes I can merge the PR.

Please proceed as you see fit. Thank you

Spacetown:
@macetw Thanks for implementing this long requested feature.

PR #694, opened 2022-11-25
Headers containing definitions appear twice in results.

LukiBa:
Issue #329: Headers containing definitions appear once with all lower-case and once with correct case characters in the results under Windows. That leads to wrong code coverage results when using templates.
Implementation of @gmc-devel suggested solution: https://github.com/gcovr/gcovr/issues/329#issuecomment-922831892

Closes: #329


gcovr/gcov.py
```diff
@@ -184,6 +185,7 @@ def guess_source_file_name(
         fname = guess_source_file_name_heuristics(
             gcovname, data_fname, currdir, root_dir, starting_dir, obj_dir, gcda_fname
         )
+    fname = str(pathlib.Path(fname).resolve())  # resolves case insensitively of windows paths
```
Spacetown, line 188: This also resolves symlinks which is a big change here.

Spacetown: If `os.path.realpath` doesn't fix the case we should first replace every usage of `os.path` with `pathlib.Path`.

Spacetown: I tried to replace `os.path` with `pathlib.Path` but stopped this because there are differences under the hood on Windows which results in Exceptions if used in SCM Clearcase.
Win errors allowed in os.path https://github.com/python/cpython/blob/a87c46eab3c306b1c5b8a072b7b30ac2c50651c0/Lib/ntpath.py#L615-L629 and here the errors allowed in pathlib https://github.com/python/cpython/blob/4710a7589aee48bec9099af308bf5938285ba336/Lib/pathlib.py#L33-L36.

LukiBa: In that case simply using ``os.path.realpath`` instead of ``pathlib.Path().resolve()`` seam to be the better solution.

Spacetown: But this also resolves the symlinks. I think it would be better fix the casing on our own.

LukiBa: If we want to fix the casing by our self, I think to only reasonable soltuion would be to recursivlely resolve the path and check if the case of the child fits the resolved path.
An alternative easy soltuion would be to convert all paths to lower case if the OS is case insensitve.

Spacetown: I prefer the first solution by splitting the path and fix the case for each part.

gcovr/utils.py
```diff
@@ -69,6 +69,29 @@ def realpath(path):
...
+def resolvePathCaseStyle(path):
+    if (os.path.dirname(path) == path):
+        return path
+
+    if os.path.islink(path):
+        matchedFileName = [f for f in os.listdir(os.path.dirname(path)) if f.casefold() == os.path.basename(path).casefold()][0]
+    else:
+        matchedFileName = os.path.basename(os.path.realpath(path))
+
+    return os.path.join(resolvePathCaseStyle(os.path.dirname(path)), matchedFileName)
```
Spacetown, line 92: I think `casefold` is wrong here since it converts e.g. `ß` to `ss` which would give a wrong result if two files differ in this letter.
The different handling of symlinks can't be tested in an easy way, therefore I prefer to always use the same code to get the correct case.

Since the file system is queried for each part of each file path this function should use a cache.

Untested:
```suggestion
@functools.lru_cache(maxsize=None)
def resolvePathCaseStyle(path: str):
    if (os.path.dirname(path) == path):
        return path

    matching_files = glob.glob(os.path.basename(path), root_dir=os.path.dirname(path)
    assert len(matching_files) == 1  # Seems that we have a case sensitve filesystem
    name = matching_files[0]

    return os.path.join(resolvePathCaseStyle(os.path.dirname(path)), name)
```

Spacetown (commented):
I've some suggestions. Is it possible to write a test which triggers the change of the case?

gcovr/gcov.py
```diff
@@ -185,7 +187,9 @@ def guess_source_file_name(
         fname = guess_source_file_name_heuristics(
             gcovname, data_fname, currdir, root_dir, starting_dir, obj_dir, gcda_fname
         )
-    fname = str(pathlib.Path(fname).resolve())  # resolves case insensitively of windows paths
+
+    if fs_case_insensitive:
```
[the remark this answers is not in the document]
LukiBa: The intention was to not call is_fs_case_insensitive() for every file. However, your suggetion of using @functools.lru_cache(maxsize=1) instead is better.

gcovr/utils.py
LukiBa: > I've some suggestions. Is it possible to write a test which triggers the change of the case?

Unfurtunately it has shown that reproducing the error is not that easy. It requires following steps: 
1. The source code has to contain definitions in header files
2. The tests are compiled on Windows using a cross compiler
3. The tests are executed on the target and the .gcda files are uploaded to the windows machine
4. gcovr is used to evaluates the code coverage  on the windows machine. 

I would suggest to wirte a UnitTest for resolvePathCaseStyle()

Spacetown:
Maybe we can add two gcov files and use the existing files with option `--keep` instead of compiling the code in our own.

gcovr/gcov.py
```diff
@@ -313,6 +307,8 @@ def process_datafile(filename, covdata, options, toerase):
     logger.debug(f"Processing file: {filename}")
 
     abs_filename = os.path.abspath(filename)
+    if platform.system() == "Windows":
+        abs_filename = abs_filename.replace("\\", "/")  # gcov requires posix style path
```
Spacetown, line 311:
```suggestion
    abs_filename = os.path.abspath(filename).replace(os.path.sep, "/")  # gcov requires posix style path
```

gcovr/utils.py
```diff
@@ -69,6 +70,42 @@ def realpath(path):
...
+def resolvePathCaseStyle(path: str):
+    rest, cur = os.path.split(path)
+    # e.g path = ".." happens if original path is like "../dir/subdir/file.cpp"
+    if not rest:
+        return cur
+    if not cur:  # e.g path = "C:/"
+        return os.path.realpath(rest)  # resolves the case of c:/
+
+    curL = cur.lower()
+    matchedFileName = [f for f in os.listdir(rest) if f.lower() == curL]
```
Spacetown, line 98: Why not using the glob function?
```suggestion
    matchedFileName = glob.glob(cur, root_dir=rest)
```

CHANGELOG.rst
```diff
@@ -26,6 +26,7 @@ Bug fixes and small improvements:
 - Fix :option:`--html-tab-size` feature. (:issue:`650`)
 - Do not ignore returncode of `gcov`. (:issue:`653`)
 - Fix alphabetical sort of html report, for when there are symlinks. (:issue:`685`)
+- Fix: Headers containing definitions appear once with all lower-case and once correct case characters in the results under Windows. (:issue:`329`)
```
Spacetown, line 29:
```suggestion
- Fix: Headers containing definitions appear once with all lower-case and once correct case characters in the results under Windows. (:issue:`694`)
```

gcovr/utils.py
```diff
@@ -69,6 +70,42 @@ def realpath(path):
...
+    cwd = os.getcwd()
+    # Guessing if file system is case insensitive.
+    # The working directory is not the root and accessible in upper and lower case.
+    ret = (
+        (cwd != os.path.sep)
+        and os.path.exists(cwd.upper())
+        and os.path.exists(cwd.lower())
+    )
+    if ret:
+        logger.info("File system is case insensitive.")
```
Spacetown, line 84: If this should be logged, then both cases and as debug message.
```suggestion
        logger.debug(f"File system is case {'in' if ret else ''}sensitive.")
```

Spacetown, line 84: If this should be logged, then both cases and as debug message.
```suggestion
        logger.debug(f"File system is case {'insensitive' if ret else 'sensitive'}.")
```

gcovr/utils.py
```diff
@@ -69,6 +70,42 @@ def realpath(path):
...
+    curL = cur.lower()
+    matchedFileName = [f for f in os.listdir(rest) if f.lower() == curL]
+    assert len(matchedFileName) < 2, "Seems that we have a case sensitive filesystem"
+
+    # e.g path = "../.." would split into ".." and ".." but cannot find .. in dir ..
+    if len(matchedFileName) == 0:
+        return path
+    return os.path.join(
+        resolvePathCaseStyle(os.path.dirname(path)), matchedFileName[0]
+    ).replace("\\", "/")
```
Spacetown, line 106: In both cases the pathes should be with `/`.
```suggestion
    if len(matchedFileName) == 1:
        path = os.path.join(
            resolvePathCaseStyle(cur), matchedFileName[0]
        )
    return path.replace("\\", "/")
```

Spacetown:
Thank you for working on this.

PR #698, opened 2022-12-05
Use scrubbed data for update of reference

Spacetown:
This PR replaces the copy command used for updating the reference data and create the reference file with the scrubbed data.


latk (approved):
Looks fine, as far as I can tell!

Issue #212, opened 2018-02-12
lines considered as "non code"


goriy:
At the moment `gcovr` makes some addtional source code lines filtering to determine is the line contains code or not (for cases when gcov produces false positives?)

It's done by means of `is_non_code()` function. Particularly it filters out lines consist of single `}`, treat it as "non code".

I've got at least one case when it leads to errors in coverage calculations. I think it's very rare case.

There's some function which calls user function and kills running thread after that:
```C
static void thread_template (void *p)
{
  thread_data_t *dat = (thread_data_t *)p;
  dat->f();                /* call user function */
  (void)kill_thread(dat);  /* delete thread  */
}
```

Corresponding piece of gcov file looks like:
```gcov
function thread_template called 128 returned 0% blocks executed 100%
      128:  373:static void thread_template (void *p)
        -:  374:{
      128:  375:  thread_data_t *dat = (thread_data_t *)p;
      128:  376:  dat->f();                /* call user function */
call    0 returned 128
      128:  377:  (void)kill_thread(dat);  /* delete thread  */
call    0 returned 0
    #####:  378:}
```

Function `kill_thread` really never exits, so the end of `thread_template()` function is never executed. So, the statement coverage for that code could never be *100%*. `gcov` and some other coverage tools calculate it correctly, but `gcovr`, due to its `is_non_code()` function shows `100%` coverage. 

For this particular file `gcovr` shows:
```
------------------------------------------------------------------------------
File                                       Lines    Exec  Cover   Missing
------------------------------------------------------------------------------
../src/tmp.c                                 123     123   100%   
```

But `gcov` for the same file shows:
```
File '../src/tmp.c'
Lines executed:99.19% of 124
```

Cases like that are not very common, I suppose, so, there's no need to patch `gcovr` right now.

I think it needs to be discussed is there any sense to have additional "non code" filtration? Is there any real cases when `gcov` makes false positives in gcov files?

Spacetown:
Is the `kill_thread` a self written function?
I only find [pthread_kill](https://man7.org/linux/man-pages/man3/pthread_kill.3.html) and this function returns:
```
function thread_entry(void*) called 1 returned 100% blocks executed 100%
        1:    6:void* thread_entry(void* arg)
        -:    7:{
        1:    8:    pthread_detach(pthread_self());
        1:    9:    printf("Inside the thread\n");
        1:   10:    fflush(stdout);
        -:   11:
        1:   12:    pthread_kill(pthread_self(), 0);
[...]

PR #704, opened 2023-01-01
New option --no-exclude-noncode-lines to not exclude noncode lines

Spacetown:
This PR adds the flag `--no-exclude-noncode-lines` to turn off the exclude of noncode lines (opening and closing braces which are tagged as uncovered by gcov).

Only the test for clang has such lines, I haven't found a construct which works for GCC.

Related issue: #212


gcovr/configuration.py
```diff
@@ -1251,9 +1251,16 @@ def merge_options_and_set_defaults(
...
+        help="Exclude coverage from lines defining a function. Default: {default!s}.",
         action="store_true",
     ),
+    GcovrConfigOption(
+        "exclude_noncode_lines",
+        ["--no-exclude-noncode-lines"],
+        group="gcov_options",
+        help="Exclude coverage from lines which seem to be non-code. Default: {default!s}.",
+        action="store_false",
+    ),
```
latk, line 1263: This comes a bit late, but note that options that are phrased as negations should be avoided.

1. Negations can be more complicated to understand for users, especially here were we have a triple negation “no”, “exclude”, “noncode”.
2. This interacts badly with config files.

For example, a config file entry `no-exclude-noncode-lines = no` is incomprehensible to me. The more I think about it, the less sure I am what is happening. As a user, I would also be confused by the help message:

```none
--no-exclude-noncode-lines
    Exclude coverage from lines which seem to be non-code.
    Default: True.
```

Does this mean that the exclusion is enabled (true), or that the option (no-exclude) is enabled by default?

Here, this could be avoided by phrasing the option positively, and auto-generating a negation:

```python
    GcovrConfigOption(
        "exclude_noncode_lines",
        ["--exclude-noncode-lines"],
        group="gcov_options",
        help="Exclude coverage from lines that don't seem to contain any code. Default: {default!s}.",
        action="store_true",
        const_negate=False, # <-- autogenerate --no-FLAG option
    ),
```

Alternatively, if a `--exclude-noncode-lines` option should never be allowed, it would be sufficient here to explicitly name the config key via `config="exclude-noncode-lines"`.

Spacetown: I had also my problems with the name but the default is to remove lines which seems to have no code.
The easiest way would be to change our default to not remove them which can result in uncovered lines for the users. But then the user will get the line coverage which is also shown by gcov and this seems to be better.

Spacetown: @latk I've tried your suggestion but it's only possible if the noncode lines are not excluded by default anymore. Our Command line parser doesn't allow a default for `action="sore_true"` (`AssertionError: action=store_true and default conflict`):
```python
    GcovrConfigOption(
        "exclude_noncode_lines",
        ["--exclude-noncode-lines"],
        config="exclude-noncode-lines",
        group="gcov_options",
        help="Exclude coverage from lines which seem to be non-code. Default: {default!s}.",
        action="store_true",
        const_negate=False,
        default=True,
    ),
```

I can go with this because than the default coverage numbers are closed to the numbers of gcov. Shall I do this breaking change?

PR #712, opened 2023-02-05
Do not use realpath for DirectoryPrefixFilter

### LINKED ISSUE #635: gcovr not scanning files if source files located below virtual path

whart222:
A minor change to apply realpath() to the path provided to DirectoryPrefixFilter.  This ensures that both the pattern used to match directories and the directory paths are converted to realpaths.

Closes #635


Spacetown:
Hi William, this place is a little bit wired, see https://github.com/gcovr/gcovr/issues/635#issuecomment-1174264534.
This will fix problems if the root directory is a symlink but break all project configurations symlinks are used inside the project structure. See the failing tests.
For the first problem we've added a warning:
https://github.com/gcovr/gcovr/blob/dce78fcd7fe094c0a21d69d442f9e8b14751d170/gcovr/__main__.py#L271-L278

whart222:
I don't deny that this is an odd situation.  However, the real impact of this PR is to ensure consistency in the gcovr logic.  The pattern is updated using a call to realpath() [line 256 of utils.py], so then the user path should also be updated with a call to realpath().  Right?

Spacetown:
@whart222 I've checked how to fix the erros of the CI. The `linked` test can be fixed by a filter but the other tests I've not managed.

After thinking a little bit about this default filter with the root directory I was thinking about removing the `realpath` from the init instead of adding it to the match. The reason is that this filter should match all files in the root directory. I don't understand why using `realpath` here.

Can you check my branch https://github.com/Spacetown/gcovr/tree/whart222 which is based on yours? With this branch the test are still working and it should solve your configuration of the home directory described in #635. In this case we can also remove the warning about a symlinked root directory.

Spacetown:
@whart222 Did you check if your problem is solved with the current version?

Spacetown (approved):
LGFM

PR #719, opened 2023-02-16
Fix deadlog on gcov error

### LINKED ISSUE #673: gcovr5.2 stays stuck
**When I excute gcovr command, logs as follow:**

executing:time gcovr -r/repo .--exclude '.*/unitTest/.*' -exclude-directories '.*/unitTest.*' --exclude-throw-branches -exclude-unreachable-branches --gcov-executable app/RHEL7-x86_64/gcc/10.3.0/bin/gcov-j 8--html--html-details -o coverage.html 
(ERROR) Trouble processing'/repo/radioswci/coverage_dir/sw/drv/dl/src/testInitializer-dlFilterEngineSrcImpl.gcda' with working directory 
Stdout of gcov was>>File '/repo/radioswci/dfe/sw/drv/dl/src/dlFilterEngineSrcImpl.cc'
Lines executed:2.40% of 292 
Branches executed:2.58% of 466 
Taken at least once:1.29% of 466 
Calls executed:2.63% of 380
Creating 'dlFilterEngineSrcImpl.cc##787d9550729fffdf4f08cd3e9c5ab75a.gcov'

File '/repo/radioswci/coverage_dir/sw/unitTest/generated/krypton/inc/registerBlock/D1FbUc.h' 
No executable lines
No branches
No calls
Removing 'D1FbUc.h##d06b338430cb9f8ab7a14591f5f516b3.gcov'

File'/afs/sero.gic.ericsson.se/app/vbui1d/RHEL7-x86_64/gcc/10.3.0/include/c++/10.3.0/new' Lines executed:0.00% of 3
No branches
No calls
Creating 'new##lac56069d705793b2479ee92f58d172b.gcov' 

<<End of stdout
Stderr of gcov was>><< End of stderr

Current processed gcov file was'/repo/radioswci/dfe/dlFi1terEngineSrcImp1.cc##787d9550729fffdf4f08cd3e9c5ab75a.gcov'. 
Use option--verbose to get extended informations.

**Then gcovr dosen't exit and dosen't print other logs, just stays stuck.
Before updating to gcovr 5.2, We use gcovr 4.2 which works normally.**

Spacetown:
If `gcovr` is lcalled with `--jobs` and the `gcov` command fails we stay stuck. I reproduced this by adding a unknown option to the `gcov` execution.
As mentioned in https://github.com/gcovr/gcovr/issues/673#issuecomment-1255427992 the problem was the directory lock. This PR catches the exception which can occur in the yield, unlocks the directory and reraise the exception. 

Closes #673

### DIFF
--- a/gcovr/workers.py
+++ b/gcovr/workers.py
@@ -56,9 +56,14 @@ def locked_directory(dir_):
     """
     Context for doing something in a locked directory
     """
-    locked_directory.global_object.run_in(dir_)
-    yield
-    locked_directory.global_object.done(dir_)
+    try:
+        locked_directory.global_object.run_in(dir_)
+        yield
+        locked_directory.global_object.done(dir_)
+    except Exception:
+        # unlock the directory and reraise the exception.
+        locked_directory.global_object.done(dir_)
+        raise
 
 
 locked_directory.global_object = LockedDirectories()
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -40,6 +40,7 @@ Bug fixes and small improvements:
 - Fix problem in decision parser if open block brace is on same line. (:issue:`681`)
 - Add Python 3.11 to test matrix. (:issue:`717`)
 - Fix casing of files if filesystem is case insensitive. (:issue:`694`)
+- Fix deadlock if :option:`-j` is used and there are errors from ``gcov`` execution. (:issue:`719`)
 
 Documentation:
 


gcovr/workers.py
```diff
@@ -56,9 +56,14 @@ def locked_directory(dir_):
...
-    yield
-    locked_directory.global_object.done(dir_)
+    try:
+        locked_directory.global_object.run_in(dir_)
+        yield
+        locked_directory.global_object.done(dir_)
+    except Exception:
+        # unlock the directory and reraise the exception.
+        locked_directory.global_object.done(dir_)
+        raise
```
latk, line 66: Wow, great find, thank you for fixing this!

FYI the resource management pattern [suggested by the `@contextmanager` docs](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) would be:

```python
def locked_directory(dir_):
  locked_directory.global_object.run_in(dir_)
  try:
    yield
  finally:
    locked_directory.global_object.done(dir_)
```

Spacetown: I wasn't sure habout the finally here. I'll change this in a new PR.

PR #723, opened 2023-02-19
Update link to gcovr documentation in HTML.

Spacetown:
The existing reports always linked to the latest version of the documentation. This PR changes the link to point to the documentation of the version used to create the current page.

The version is also set to `master` on the development branch. If someone installs the current master branch and calls `pip freeze` the version is shown as `gcovr==5.2` which isn't the truth.
When bumping the new version we should change it back to `master` after the release.


latk:
Oh this is good.

But instead of using the label `master` as a version we should likely use a [local version identifier](https://peps.python.org/pep-0440/#local-version-identifiers) like `5.2+master`. Ideally, we'd use the Git commit as the local version, but I have no idea how that can be retrieved in a robust manner.

Would such versions require updates to the release checklist script? Could it work somewhat robustly by stripping off the local version part from a version number? In any case, it should check that a release proceeds only with a valid public version number.

Spacetown:
> Oh this is good.
> 
> But instead of using the label `master` as a version we should likely use a [local version identifier](https://peps.python.org/pep-0440/#local-version-identifiers) like `5.2+master`. Ideally, we'd use the Git commit as the local version, but I have no idea how that can be retrieved in a robust manner.
> 
> Would such versions require updates to the release checklist script? Could it work somewhat robustly by stripping off the local version part from a version number? In any case, it should check that a release proceeds only with a valid public version number.

Master doesn't work. I'm working on the checklist and will force a push if it's working.

PR #727, opened 2023-02-21
Replace support for keil with documentation entry

### LINKED ISSUE #697: Backwards compatibility for older gcov branch syntax to support uVision Keil coverage output.
We use gcovr to generate coverage reports from gcov files output by uVision Keil debugger, but gcovr after v4.2 cannot parse the branch syntax in the gcov files. It appears gcovr has deprecated support for older branch syntax in after 4.2.

We'd like to request backwards compatibility to support the legacy branch syntax, if possible.

The old branch syntax looks like this example (excerpt of a gcov file):
    19:   113:    if ((f_mode_g == f_mode_inner)||(f_mode_g == f_mode_outer))
branch 0 executed 1
branch 1 skipped 1
branch 2 executed 1
branch 3 skipped 1
branch 4 executed 1
branch 5 skipped 1
branch 6 executed 1
branch 7 skipped 1

The newer syntax is described in the comments of the gcovr v5.2 file, gcov_parser.py (line 850) as:
    # BRANCH
    #
    # Structure:
    # branch BRANCHNO never executed
    # branch BRANCHNO taken VALUE
    # branch BRANCHNO taken VALUE (ANNOTATION)

If you can't restore support for this syntax, then could you provide an explanation of how the older syntax maps to the new so we can implement a conversion script ourselves?

Spacetown:
Since the #623 only adds support for some markers and #697 showed that the files generated by Keil uVision compiler are not compatible with the [gcov specification](https://gcc.gnu.org/onlinedocs/gcc-11.1.0/gcc/Invoking-Gcov.html) this PR undo the changes from  #623 and adds a entry to the cookbook how the files can be transformed. 

Closes #697


doc/source/cookbook.rst
```diff
@@ -112,6 +112,27 @@ directory in ``BLD_DIR``, rather than next to the sources. Since
...
+``gcov`` specification.
+To support coverage data generated by this compiler you have to create the gcov files as documented
+in `Keil uVision documentation`_
+and process them before running ``gcov`` to get the correct format.
+
+.. code-block:: sh
+
+    # Untested because we have no access to Keil uVision compiler
+    find . -name '*.gcov' -exec sed -i '' -e 's/\(\#\#\#\#\#)\#/\1/; /^branch/ { s/executed 0/never executed/; s/executed .*/taken 1/; s/skipped .*/never executed/; }' {} \;
+    gcovr --use-gcov-files .
```
latk, line 134: Since the transformation is a bit more complex I'd suggest extracting the Sed commands into a separate script. Also, the script should cover all markers for uncovered code. Then:

```rst
Save the following Sed script as ``fix-gcov.sed``:

.. code-block::

   # fix markers for uncovered code:
   # match any of #=%$ repeated 6 times
   s/^\([#=%$]\)\(\1\{5\}\)/\2/

   # fix branch tags
   /^branch/ {
     s/executed 0/never executed/
     s/executed .*/taken 1/
     s/skipped .*/never executed/
   }

Then, apply this Sed script to all gcov files before invoking gcovr:

.. code-block:: sh

   find . -name '*.gcov' -exec sed -i -f fix-gcov.sed {} \;
   gcovr --use-gcov-files

.. warning:: Untested because we have no access to Keil uVision compiler
```

latk (approved):

PR #731, opened 2023-02-23
Add clang 14 to test suite

Spacetown:
Since XCode has updated from `clang-13` to `clang-14` I need to add this version to our test suite to be able to test faster.


latk (approved):
lgtm. Not 100% sure about the changes to the Dockerfile, but I think the chown had become superfluous anyway.

Spacetown:
> gtm. Not 100% sure about the changes to the Dockerfile, but I think the chown had become superfluous anyway.

Yes, it's not needed anymore because we don't generate a artifact anymore. Also the upload can be removed. The option to create the diff.zip was removed with the switch to nox in #516.

PR #749, opened 2023-03-10
Fix summary report

### LINKED ISSUE #745: `--print-summary` and `--decisions`
**Describe the bug**
`gcovr --decisions --print-summary` prints:

```
lines: 90.1% (13347 out of 14816)
functions: 91.2% (2338 out of 2565)
branches: 56.9% (14099 out of 24788)
calls: 0.0% (0 out of 0)
```

i.e., a summary of decision coverage is *not* printed.
On the other hand, the "calls:" line is always printed, regardless of the `--calls` option.

**Expected behavior**
* Show decision coverage summary when `--decisions` is specified.
* Do not print (nonsensical) calls coverage unless `--calls` is specified.

 - OS: Linux
 - GCC version 12.2.0
 - GCOVR version 6.0

Spacetown:
The decission summary is missing and the calls summary is printed even if the values aren't gathered.

This PR fixes the summary report by:

- Only print call coverage summary if option `--calls` is given.
- Print decision summary coverage if option `--decisions` is given.

Closes #745

### DIFF
--- a/gcovr/configuration.py
+++ b/gcovr/configuration.py
@@ -807,7 +807,7 @@ GCOVR_CONFIG_OPTIONS = [
         "show_decision",
         ["--decisions"],
         group="output_options",
-        help="Report the decision coverage. For HTML, JSON  and the summary report.",
+        help="Report the decision coverage. For HTML, JSON, and the summary report.",
         action="store_true",
     ),
     GcovrConfigOption(
--- a/gcovr/configuration.py
+++ b/gcovr/configuration.py
@@ -807,14 +807,14 @@ GCOVR_CONFIG_OPTIONS = [
         "show_decision",
         ["--decisions"],
         group="output_options",
-        help="Report the decision coverage. For HTML and JSON report.",
+        help="Report the decision coverage. For HTML, JSON  and the summary report.",
         action="store_true",
     ),
     GcovrConfigOption(
         "exclude_calls",
         ["--calls"],
         group="output_options",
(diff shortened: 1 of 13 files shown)


gcovr/configuration.py
```diff
@@ -807,14 +807,14 @@ def merge_options_and_set_defaults(
         "show_decision",
         ["--decisions"],
         group="output_options",
-        help="Report the decision coverage. For HTML and JSON report.",
+        help="Report the decision coverage. For HTML, JSON  and the summary report.",
```
Pesa, line 810:
```suggestion
        help="Report the decision coverage. For HTML, JSON, and the summary report.",
```

Pesa (commented):
Haven't tested the patch but looks like it should fix my issue. Thanks!

gcovr/configuration.py
Spacetown: This additional coma is wrong.

Pesa: It's and oxford comma, it's a stylistic choice so entirely up to you. I suggested adding it because I saw it used elsewhere in gcovr documentation, e.g. https://github.com/gcovr/gcovr/blob/2591d80014a2f7bb681863d1adf978c12d8560b0/gcovr/configuration.py#L1209

Spacetown: Didn't know the "Oxford comma", learned something new.

Pesa (approved):
Thank you!

PR #750, opened 2023-03-12
Fix nested report if only one directory in root is present.

### LINKED ISSUE #744: "Directory" in HTML report header is always `./`
**Describe the bug**
I would expect "Directory:" in the HTML report header (using `--html-nested` or `--html-details`) to use the value of the `--root` option, but it seems that it's always `./`

**To Reproduce**
1. just run `gcovr --html-nested out/ --root src build` or similar

**Screenshots**
Screenshot 2023-03-08 at 11 11 20 PM

 - OS: Linux
 - GCC version 12.2.0
 - GCOVR version 6.0

Spacetown:
The collapse of the directory wasn't working for the content of the root directory itself. This PR adds a test where in the root directory a single sub directory exists. In this case the content of the sub directory wasn't moved up to the root level.

For nested report the evaluating of the common path wasn't working because the root_filter wasn't applied to the directory statistics of the root directory.

Closes #744


macetw (approved):
Nice.

PR #766, opened 2023-04-08
Add support for GCOV JSON intermediate format

### LINKED ISSUE #282: Support GCOV intermediate format

Spacetown:
GCOV introduced a JSON intermediate format with version 9.1.
If the gcov help contains the option `--json-format` this option is used to get compressed JSON files (`.gcov.json.gz`) instead of the human readable files (`.gcov`). The compressed JSON files are only used if the calls are excluded (default) because the call information isn't present in the JSON format yet.

Closes #282


gcovr/formats/gcov/read.py
```diff
@@ -91,11 +105,13 @@ def read_report(options: Options) -> CovData:
 
 
 def find_existing_gcov_files(search_path, exclude_dirs):
-    """Find .gcov files under the given search path."""
+    """Find .gcov and .json.gz files under the given search path."""
     LOGGER.debug(f"Scanning directory {search_path} for gcov files...")
     gcov_files = list(
         search_file(
-            re.compile(r".*\.gcov$").match, search_path, exclude_dirs=exclude_dirs
+            re.compile(r".*(\.gcov|\.json\.gz)$").match,
```
Pesa, line 112: The `.json.gz` extension seems rather generic, there's a risk of matching other files unrelated to gcov.

The documentation of `--json-format` says it produces files with extension `.gcov.json.gz` so maybe you can use that instead?

fanfeilong:
.gcov files are useable when debug

Spacetown:
> .gcov files are useable when debug

???

Spacetown:
@Pesa I would be happy if you could review the change so that we can close this PR after a long time.

PR #773, opened 2023-04-24
Add option for decision and function threshold

### LINKED ISSUE #759: Support `fail_under_decision` option in CLI
I saw that https://github.com/gcovr/gcovr/pull/749 introduced the decisions metric produced in summary report, but that there does not seem to be a way to fail with an exit code specifically on the decision coverage percentage.

I would be interested in a `--fail-under-decision` option being added to the CLI. In addition to the CLI itself, this would require updating the [fail_under](https://github.com/gcovr/gcovr/blob/master/gcovr/__main__.py#L61) function to properly fail under a certain decision coverage amount. This would also require updating the error codes to maintain the ability to OR them together.

It's possible to extract this percentage out of the reported summary and run my own check to fail under a desired percent, but that seems less elegant than adding this as a feature that would potentially be useful for others.

Spacetown:
Add `--fail-under-decision` and `--fail-under-function` which will error under a given minimum coverage.

The exit code for an error of the reader module is changed from 8 to 64 and for a writer from 7 to 128 to have the coverage thresholds together.
The documentation also stated that the threshold exit statuses can be ORed. This wasn't true anymore since the writer error (7) sets also the bits for the minimum line (2) and branch coverage (4). Now this behavior is fixed. 

Closes #759


Codym48:
Thank you for doing this @Spacetown! Any idea when the next gcovr release is planned?

Spacetown:
Nothing planned yet. I think before end of this year should be possible.

Codym48:
Sounds good, let me know if there's anything we can do to help!

Spacetown:
@Codym48 Review open PRs. I don't like to merge without review.

PR #779, opened 2023-05-06
fix in ubuntu18 line format(COUNT:LINENO-block  BLOCKNO)

oathupdate:
image


Spacetown:
Sorryl, a little bit more context should be given. see also https://github.com/gcovr/gcovr/blob/master/CONTRIBUTING.rst.

Issue #783, opened 2023-05-12
TypeError Exception while decision analysis for switch-statement


genossefloyd:
I have got the following exception while running gcovr 6.0 on our project using the flags "--html-details" and "--decisions":
>   File "/home/build/.local/lib/python3.9/site-packages/gcovr/decision_analysis.py", line 228, in parse_one_line
>     lineno, max(lineno + 1, *self.coverage.lines.keys())
> TypeError: 'int' object is not iterable

One of the switch-statements, causing the exception was looking like this:
```
switch (step) {
    case Step::Start: return 0;
    case Step::DownloadDocuments: return 24.0;
    case Step::EncryptDocuments: return 10.0;
    case Step::SubmitDocuments: return 25.0;
    case Step::Finished: return 0;
}
```
Another one was:
```
switch(level) {
    case LogLevel::DEBUG:
        label = "DEBUG";
        break;
    case LogLevel::INFO:
        label = "INFO";
        break;
    case LogLevel::WARN:
        label = "WARN";
        break;
    case LogLevel::ERROR:
        label = "ERROR";
        break;
    default:
        break;
}
```

I have debugged the issue locally and found out, that the dictionary `self.coverage.lines` could be empty, which will cause the exception to be thrown by the *-operator on the empty key list.

I have fixed the issue locally by calculating the max line number beforehand and checking the dictionary length:
```
-            for next_lineno in range(
-                lineno, max(lineno + 1, *self.coverage.lines.keys())
-            ):
+            maxlines = max(lineno + 1, *self.coverage.lines.keys()) if len(self.coverage.lines) > 0 else lineno +1
+            for next_lineno in range(lineno, maxlines):
```

Spacetown:
Thanks for finding the fix. Can you raise a pull request for this?

genossefloyd:
#784 My very first contribution to foreign Github repo. Hope it is alright this way 😅

Spacetown:
PR is already merged.

PR #784, opened 2023-05-15
Fix type error in decision analysis

genossefloyd:
Proposal for fix of #783 
Avoiding TypeError exception if `self.coverage.lines` is empty during decision analysis


Spacetown:
Thank you for fixing this. Can you also extend the decision test to raise this exception without your fix?

genossefloyd:
I have extended the decisions test with a setup using a second cpp, that was raising the issue - but only for clang (14) and not for gcc (10).

The two failing test cases require an update of the json and html output reference. The "--update-reference" command did not work for me... I also do not have all the required build environments. Could you straigthen out these references yourself?

Spacetown:
> The "--update-reference" command did not work for me...

The switch is written with a `_`and not a `-` in the middle.

The change in the makefile is not working for us. Can you modify the test to use a enumeration in the same C-File without adding another compile unit?
I guess the problem is the scope operator in the case value.

genossefloyd:
I tried to different switch-variations (with break, return, additional brackets, ...) to make it come up with one cpp but it does not work. There have to be two cpp's.

As far as I understood it, the code in one cpp is creating some actual coverage of the switch statement, the other one only references the containing class but does not call the relevant function, so it ends up with 0 coverage and we end up with an empty list.

I have no real insights on how gcov works exactly but that is my best guess about whats happening after trying several setups half of the day.

As I am a bit out of time/resources to investigate the problem any further, I can not continue working on it - at least for the next couple of weeks. Sorry about that.

Spacetown (changes requested):

PR #785, opened 2023-05-16
Doxygen Awesome style HTML theme.

dani007200964:
I really like Doxygen-generated code documentation and I use a theme called [Doxygen Awesome](https://jothepro.github.io/doxygen-awesome-css/). I think it would be nice to add pre-made styles for the html output. This way the coverage report can be integrated with a more complex documentation system with the same style.

image

image

image

### DIFF
--- a/extras/html_styles/doxygen_awesome_dark_style.css
+++ b/extras/html_styles/doxygen_awesome_dark_style.css
@@ -1,0 +1,681 @@
+/*
+
+---- This is a Doxygen Awesome inspired theme. ----
+
+With this theme you can match the style of the coverage report
+with the Doxygen generated documentation( of course with Doxygen
+Awesome theme )
+
+Modified by: 
+Daniel Hajnal, hajnal.daniel96@gmail.com 2023.05.16
+
+*/
+
+:root {
+    font-family: sans-serif;
+    --tab_size: 4;
+}
+
+.theme-green, .theme-blue {
+    --unknown_color: lightgray;
+    --low_color: #ad2617;
+    --medium_color: #8a6800;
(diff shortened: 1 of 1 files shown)


Spacetown:
I'm not sure if we should provide such "specialized" themes. @latk what do you think about this?

There is no option to switch to this theme like `dark_green` and `dark_blue`in:
https://github.com/gcovr/gcovr/blob/ddc69302abd938ee5e19c4d86e028b74f6408882/gcovr/formats/html/write.py#L112

If we provide this theme I expect that it's added into the existing stylesheets.

Your css also contains the part of `pygments` which we always append:
https://github.com/gcovr/gcovr/blob/ddc69302abd938ee5e19c4d86e028b74f6408882/gcovr/formats/html/write.py#L372-L373
https://github.com/gcovr/gcovr/blob/ddc69302abd938ee5e19c4d86e028b74f6408882/gcovr/formats/html/write.py#L184-L189

This part needs also to be adapted if it's changed by the theme.

One thing this PR points out is that we missed the option to use pygments but provide the style in the user template customize it depending on the user stylesheet.

Spacetown:
> One thing this PR points out is that we missed the option to use pygments but provide the style in the user template customize it depending on the user stylesheet.

A solution for this can be a check for the comments which are added to the pygments stylesheet. If such a comment or the word `pygments` is found we do not append the pygments template.
https://github.com/gcovr/gcovr/blob/ddc69302abd938ee5e19c4d86e028b74f6408882/gcovr/tests/html-default/reference/gcc-5/coverage.css#L544

Spacetown:
No response.

Issue #769, opened 2023-04-13
gcovr 6.0 does not generate outputs with gcc8


MouhK:
**Describe the bug**
The issue occurs in our CI running ubuntu 18.04 and gcc8 (reproduced locally with docker). With previous version of gcovr (until 5.2) I had no issue, but since we updatated to gcovr 6.0, gcovr provides empty reports.

**To Reproduce**
1. Pull the project
2. Patch the script `doc/examples/example.sh` 

```
diff --git a/doc/examples/example.sh b/doc/examples/example.sh
index ba481487..c21be95e 100755
--- a/doc/examples/example.sh
+++ b/doc/examples/example.sh
@@ -13,7 +13,7 @@ g++ -fprofile-arcs -ftest-coverage -fPIC -O0 example.cpp -o program
 #END run
 
 #BEGIN gcovr
-gcovr
+gcovr -v
 #END gcovr
 
 rm -f program *.gc*
```

3. Launch the script `example.sh`

**Expected behavior**
gcovr reports data

***With gcc 7 (7.5.0)***
```
/example.sh 
(DEBUG) File system is case sensitive.
(DEBUG) Filters for --root: (1)
(DEBUG)  - re.compile('^/workspace/doc/examples/')
(DEBUG) Filters for --filter: (1)
(DEBUG)  - DirectoryPrefixFilter(/workspace/doc/examples/)
(DEBUG) Filters for --exclude: (0)
(DEBUG) Filters for --gcov-filter: (1)
(DEBUG)  - AlwaysMatchFilter()
(DEBUG) Filters for --gcov-exclude: (0)
(DEBUG) Filters for --exclude-directories: (0)
(DEBUG) Scanning directory . for gcda/gcno files...
(DEBUG) Found 2 files (and will process 1)
(DEBUG) Pool started with 1 threads
(DEBUG) Processing file: /workspace/doc/examples/example.gcda
(DEBUG) Running gcov: 'gcov --help' in '.'
(DEBUG) Running gcov: 'gcov --help-hidden' in '.'
(DEBUG) Running gcov: 'gcov /workspace/doc/examples/example.gcda --branch-counts --branch-probabilities --demangled-names --hash-filenames --object-directory /workspace/doc/examples' in '/workspace/doc/examples'
(DEBUG) Finding source file corresponding to a gcov data file
  gcov_fname   /workspace/doc/examples/example.cpp##9597a7a3397b8e3a48116e2a3afb4154.gcov
[...]

PR #789, opened 2023-05-30
Fix filtering wrong path for gcov file

ukyouz:
In Windows, the path include Disk Drive letter in the begining, which was error prone. And thus can not generate any coverage report as described in #769. This changes fix the issue.

* also fix bugs when source file may comes from different Disk Drive.

### DIFF
--- a/gcovr/coverage.py
+++ b/gcovr/coverage.py
@@ -91,7 +91,7 @@ def sort_coverage(
 
     def filename(key: str) -> str:
         return (
-            force_unix_separator(os.path.relpath(realpath(key), basedir))
+            force_unix_separator(realpath(key))
             if filename_uses_relative_pathname
             else key
         )
--- a/gcovr/gcov.py
+++ b/gcovr/gcov.py
@@ -124,7 +124,7 @@ def process_gcov_data(
     # Return if the filename does not match the filter
     # Return if the filename matches the exclude pattern
     filtered, excluded = apply_filter_include_exclude(
-        fname, options.filter, options.exclude
+        gcda_fname, options.filter, options.exclude
     )
 
     if filtered:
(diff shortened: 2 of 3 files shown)


gcovr/formats/html/write.py
```diff
@@ -290,7 +290,7 @@ def add_file(
         }
 
         display_filename = force_unix_separator(
-            os.path.relpath(realpath(cdata_fname), self.directory)
+            realpath(cdata_fname)
```
Spacetown, line 293: Now always absolute paths are displayed instead it relative paths to the root directory.

gcovr/formats/gcov/read.py
```diff
@@ -182,7 +182,7 @@ def process_gcov_data(
     # Return if the filename does not match the filter
     # Return if the filename matches the exclude pattern
     filtered, excluded = apply_filter_include_exclude(
-        fname, options.filter, options.exclude
+        gcda_fname, options.filter, options.exclude
```
Spacetown, line 185: The filter must use the source file and not the file name of the coverage data.

Spacetown (changes requested):
Source files coming from different drives seems to be a bug in the project setup. A project should always use one drive, the only thing coming from a different drive should be the compiler.

Spacetown:
I think the linked issue is wrong. The linked one is related to the example script and gcc-8 on Ubuntu.

ukyouz:
My original issue is that, there is **no coverage report after running gcovr** command. Maybe the patch is not correct, but I can get all the coverage report as I expected with such fix. Yay I also got the report from compiler sources, that is why I need to handle the disk drive issue.

I think somehow the default filter just **filters out** all the results, that is why I got nothing. I will consider the comment above and re-debug again with my project setup and fix the filter issue.

ukyouz:
Sure, I opened a issue at #790.

PR #792, opened 2023-06-03
📚🩹 Fix wrong command in 'How to create a standalone application' docs

s-weigand:
Thanks for adding the bundling option, this makes it a lot easier to add into docker images without the need to add python itself.

There is a small bug in the docs where the command is `python3 -m nox --session build_app` but should be `python3 -m nox --session bundle_app`.

https://github.com/gcovr/gcovr/blob/d8871de966391adb753bbcf5da3961ab5110231b/noxfile.py#L335


Spacetown:
Thanks for fixing this. Pleas can you add e entry in the changeling and add yourself to the contributors?

Spacetown (approved):
I've updated the authors file (Firstname Lastname) and replaced single quotes with RST inline verbatim in changelog.

s-weigand:
@Spacetown Thanks for fixing that.
Also looks like [cryptography had a new release 2 days back ](https://pypi.org/project/cryptography/), guess that is why the CI fails

Issue #791, opened 2023-06-03
HTML report github style test


mfont-bz17:
Hi,

I've made a test ([branch](https://github.com/mfont-bz17/gcovr/tree/feature/html-report-github-style)) using [github primer.css](https://primer.style/css/) to make the html reports with the style of github. 
You can see attached the screenshots. If you think than can be useful for the project, then I can continue working on it ;) some details are not finished yet...

**Only html report**

![Screenshot 2023-06-03 at 12-02-18 GCC Code Coverage Report](https://github.com/gcovr/gcovr/assets/121682800/4bd57f35-1a06-4db5-bba7-cb1ad89ccfae)

**Directory page**

![Screenshot 2023-06-03 at 11-55-58 GCC Code Coverage Report](https://github.com/gcovr/gcovr/assets/121682800/76502c62-a6c1-48aa-bac2-afa679df7926)

**Source page**

![Screenshot 2023-06-03 at 12-10-01 GCC Code Coverage Report](https://github.com/gcovr/gcovr/assets/121682800/d2ee41d2-74fe-437e-86ba-16d548b77454)

**Functions page**

![Screenshot 2023-06-03 at 11-57-30 GCC Code Coverage Report](https://github.com/gcovr/gcovr/assets/121682800/8d2064ee-8210-408e-8d20-51733366f83b)

**Functions page**

Spacetown:
Looks great but we've to think about the structure of the templates if we provide several themes. There is also a PR for a doxygen theme open (#785).
At the moment we have the option `--html-theme` which affects the class attribute in the HTML report  and we use the same stylesheet. I guess in your case it's better to exchange the whole stylesheet. It would also be great to provide a blue theme here.

mfont-bz17:
To avoid misunderstandings for me:

- theme: the layout change
- flavor: only the palette colors change

In my case not only the stylesheet must be changed, the html templates too. Is a new directory inside format/html. 
I think that each theme should have a specific folder and each theme can have different flavors, for now green and blue but can exists more.

- gcovr/formats/html/templates/    -> Default theme
- gcovr/formats/html/github/templates/ -> Github theme 
- gcovr/formats/html/doxygen/templates/ -> Doxygen theme 

As you pointed in #785 the part of highlight is important, one option should be that every theme has a mini config to overwrite the default values:

```
pygments-style: arduino
falvors: green, blue, dark_blue, dark_green
```

And then use the --html-theme option to choose the theme and flavor for example
```
[...]

PR #758, opened 2023-03-28
Template dir option

frankwiles:
This PR adds a `--template-dir` option to the CLI which allows the use of a directory of alternate Jinja2 templates for users who want to heavily customize their HTML output but don't want to fork gcovr to do so.  

I believe I've covered all the bases here, README, docs, etc but let me know if I've missed anything you would like to see. 

Closes #714

PR #793, opened 2023-06-07
Feature/html report GitHub style

mfont-bz17:
Here the PR created.

For now in the PR you can choose 2 themes and different colors for each theme.

`--html-theme: (choose from 'green', 'blue', 'github.blue', 'github.green', 'github.dark-green', 'github.dark-blue')`

Is still to be decided is where we configure the theme colors and theme pygment style highlight. 
For now is hardcoded in the code.

Related #791


Spacetown:
> Thinking about the highlight for me the best option is that the theme include the highlight css directly. The theme can manage the highlight based on the color too if necessary. We need to check if --html-syntax-highlighting or --html-details-syntax-highlighting enabled. If we include the theme in the css then PYGMENTS_CSS_MARKER wouldn't be necessary.
> 
> ```
> {% HIGHLIGHT %}
> {% if (theme == 'dark-blue' or theme == 'dark-green') %}
> {% include "highlight-dark.css" %}
> {% else %}
> {% include "highlight.css" %}
> {% endif %}
> ```
> 
> It's easy to create the highlight css from the command line and then include in style.css
> 
> `pygmentize -S default -f html > highlight.css`

But in the past the CSS changed if we update the pygments version. That's the reason why we pinn the version of pygments for our test.

What about adding a file with the highlight style since this is only a string. We can also use the template engine to get resolve the file and use the returned string for pygments styling.

mfont-bz17:
This is a good reason to keep inside.

> But in the past the CSS changed if we update the pygments version. That's the reason why we pinn the version of pygments for our test.

Then seems that this solution is better. I'm not sure if put only the file with the string or maybe is better put some kind of miniconfig file format (json, yaml, toml...) to have more flexibility in the future if necessary......

> What about adding a file with the highlight style since this is only a string. We can also use the template engine to get resolve the file and use the returned string for pygments styling.

Your option should be something like this (one file per theme if needs to overwrite the default style, if not the file is not necessary)

```
==> blue.highlight <==  (filename and content)
default
==> dark-blue.highlight <==
github-dark
==> dark-green.highlight <==
monokai
==> green.highlight <==
arduino
```
or with config files...

```
 - blue:
     highlight: default
     ...
 - green:
     highlight: arduino
     ...
 - dark-green:
     highlight: monokai
     ...
 - dark-blue:
     highlight: github-dark
     ...
     
```    
Both are ok for me

Spacetown:
> Your option should be something like this (one file per theme if needs to overwrite the default style, if not the file is not necessary)
> 
> ```
> ==> blue.highlight <==  (filename and content)
> default
> ==> dark-blue.highlight <==
> github-dark
> ==> dark-green.highlight <==
> monokai
> ==> green.highlight <==
> arduino
> ```

I prefer a filename `<colour>.pygments` because it contains the style for pygments. It should be mandatory for a theme.

> or with config files...
> 
> ```
>  - blue:
>      highlight: default
>      ...
>  - green:
>      highlight: arduino
>      ...
>  - dark-green:
>      highlight: monokai
>      ...
>  - dark-blue:
>      highlight: github-dark
>      ...
>      
> ```
> 
> Both are ok for me

We need to configure everything inside the directory of the theme. If we introduce a separate file which contains the configuration for all themes we will have a problem in #758.

mfont-bz17:
> We need to configure everything inside the directory of the theme. If we introduce a separate file which contains the configuration for all themes we will have a problem in https://github.com/gcovr/gcovr/pull/758.

Yes the idea is put a config file inside the directory of theme:

default/theme.conf
github/theme.conf
doxygen/theme.conf

Spacetown:
For now we only have the style used for pygments. I suggest to use a simple text file with the style only.

mfont-bz17:
> For now we only have the style used for pygments. I suggest to use a simple text file with the style only.

Pushed in the PR.  As you said previously the files  `<color>.pygments` are mandatory for theme.

gcovr/formats/html/write.py
```diff
@@ -69,16 +69,23 @@ def reuse_value(*args):
...
 # Only do this work if templates are actually used.
 # This speeds up text and XML output.
 @Lazy
-def templates():
+def templates(options):
     from jinja2 import Environment, PackageLoader
 
     return Environment(
-        loader=PackageLoader("gcovr.formats.html"),
+        loader=PackageLoader("gcovr.formats.html.%s" % get_theme_name(options.html_theme)),
```
Spacetown, line 88: Please use f-strings.
```suggestion
        loader=PackageLoader(f"gcovr.formats.html.{get_theme_name(options.html_theme)}"),
```

gcovr/formats/html/write.py
```diff
@@ -110,7 +117,9 @@ def load_user_template(template):
 
 class CssRenderer:
 
-    THEMES = ["green", "blue"]
+    THEMES = ["green", "blue",
+              "github.blue", "github.green", "github.dark-green", "github.dark-blue"
+    ]
```
Spacetown, line 122: Fix code style:
```suggestion
    THEMES = [
        "green",
        "blue",
        "github.blue",
        "github.green",
        "github.dark-green",
        "github.dark-blue",
    ]
```

gcovr/formats/html/write.py
```diff
@@ -146,12 +155,12 @@ def highlighter_for_file(filename):
 
 
 class PygmentHighlighting:
-    def __init__(self):
+    def __init__(self, style="default"):
```
Spacetown, line 158: There is no default style.
```suggestion
    def __init__(self, style: str):
```

gcovr/formats/html/write.py
```diff
@@ -69,16 +69,23 @@ def reuse_value(*args):
     def __call__(self, *args):
         return self.get(*args)
 
+# html_theme string is <theme_directory>.<color> or only <color> (if only color the use default)
+# examples: github.green github.blue or blue or green
+def get_theme_name(html_theme):
```
Spacetown, line 74: Add type annotations:
```suggestion
def get_theme_name(html_theme: str) -> str:
```

gcovr/formats/html/write.py
```diff
@@ -69,16 +69,23 @@ def reuse_value(*args):
...
+    return html_theme.split(".")[0]  if "." in html_theme else "default"
+
+def get_theme_color(html_theme):
```
Spacetown, line 77:
```suggestion
def get_theme_color(html_theme: str) -> str:
```

gcovr/formats/html/write.py
```diff
@@ -183,8 +192,9 @@ def highlighter_for_file(self, filename):
 
 @Lazy
 def get_formatter(options):
+    highlight_style = templates(options).get_template("%s.pygments" % get_theme_color(options.html_theme)).render()
```
Spacetown, line 195: Use f-strings. I think it's better to name the file `pygments.<colour>` to group the files in the file manager.
```suggestion
    highlight_style = templates(options).get_template(f"pygments.{get_theme_color(options.html_theme)}").render()
```

Spacetown (changes requested):
I've done a quick walk through and found some smaller issues.

Spacetown:
> Pushed changes.
> 
> I've added some online examples here:
> 
> [github.green](https://bz17.dev/gcovr/github-green/) [github.blue](https://bz17.dev/gcovr/github-blue/) [github-dark.green](https://bz17.dev/gcovr/github-dark-green/) [github-dark.blue](https://bz17.dev/gcovr/github-dark-blue/)

I'm not sure if it's better to remove the `- %` complete. It looks confusing for me because there is also `0.0 %`.

gcovr/formats/html/github/templates/directory_page.summary.html
```diff
@@ -11,48 +11,75 @@
                 <span>{{info.get_directory()}}</span>
             </div>
             <div>Date: {{info.date}}</div>
-            <div>
-                <span class="mr-2">Coverage:</span>
-                <span class="IssueLabel coverage-low color-fg-on-emphasis mr-1">low: &ge; 0%</span>
+            <div class="d-flex flex-wrap" style="margin-bottom: 2px;">
+                <div class="mr-2" style="min-width: 65px;">Coverage:</div>
```
Spacetown, line 15: The Lines row in https://bz17.dev/gcovr/legends/ is 65px long the other ones are about 67px on my screen. Maybe this here needs to be increased.

Spacetown (commented):
Looks good, now there is only a minor problem with the alignments.

gcovr/formats/html/github/templates/directory_page.summary.html
mfont-bz17: Pushed.
Yes you are right in my screen is almost 66px not 65px. I 've increased min-width to 67px.

Spacetown: On my screen it was 67,067. I think 70 will be better.

mfont-bz17: Changed

Spacetown:
Shall I add some test and the reference data?

mfont-bz17:
> Shall I add some test and the reference data?

This should be great. You prefer to do it or if you want I can do it. Maybe you can do faster or you have some script to make the reference data...

Spacetown:
If you have docker installed you can give it a try. The option `--generate_reference` is documented [here](https://gcovr.com/en/stable/contributing.html#run-and-filter-tests).

Spacetown:
You also need to run `nox -s black -- .` to format the files.

mfont-bz17:
Yes I've docker installed, I'll try it. Thanks for the info.

mfont-bz17:
Making the tests I've discovered some problems with pygments versions. Seems that when I was doing the examples and when run gcovr directly (installed via pip), is using my system version (python3-pygments 2.3.1) and now when I ran the --generate_reference was using pygments==2.7.4 and the css of pygments are different. 

Did you see this before? So you can fix the pygments version in gcovr package but in my case seems that is still using the other version, this can be a problem, I've to investigate a bit more....

pygments==2.7.4
```
 pre { line-height: 125%; }
 td.linenos pre { color: #000000; background-color: #f0f0f0; padding-left: 5px; padding-right: 5px; }
 span.linenos { color: #000000; background-color: #f0f0f0; padding-left: 5px; padding-right: 5px; }
 td.linenos pre.special { color: #000000; background-color: #ffffc0; padding-left: 5px; padding-right: 5px; }
 span.linenos.special { color: #000000; background-color: #ffffc0; padding-left: 5px; padding-right: 5px; }
```

imatge


python3-pygments 2.3.1
```
 pre { line-height: 125%; }
 td.linenos .normal { color: inherit; background-color: transparent; padding-left: 5px; padding-right: 5px; }
 span.linenos { color: inherit; background-color: transparent; padding-left: 5px; padding-right: 5px; }
 td.linenos .special { color: #000000; background-color: #ffffc0; padding-left: 5px; padding-right: 5px; }
 span.linenos.special { color: #000000; background-color: #ffffc0; padding-left: 5px; padding-right: 5px; }
```

imatge

Spacetown:
In the docker test the correct version should be used.

We pinned the version because of our reference data because after a pygments update all HTML tests where failing.

mfont-bz17:
I've investigated a bit more.

1.  The pygments version used in my computer is 2.14.0 installed via pip. The OS system is older 2.3.1 and is not used.
2.  The problem of different css between 2.7.4 and 2.14.0 . In the theme the linenos style generated by pygments is used and differs between pygments versions, I've solved adding this line in my css style: `td.linenos .normal { color: inherit; background-color: transparent; padding-left: 5px; padding-right: 5px; }`. 
3. During the generation of the tests reference I've discovered another error.  In tests the version of pygments is 2.7.4 and github-dark highlight style was added in [pygments 2.13.0](https://github.com/pygments/pygments/blob/88c23eaf98a5b86ba5ed666ba0f982588685e0b8/CHANGES#L232).  

```
(ERROR) Error occurred while printing reports:
Could not find style module 'github-dark'.
```
I think that should be a fallback style if the choosed one by the theme can not be applied due the pygments version and avoid the error. What do you think?

Spacetown:
You also need to run `nox -s black -- .` to format the fil

> I think that should be a fallback style if the choosed one by the theme can not be applied due the pygments version and avoid the error. What do you think?

We can update the pygments version for our test and update all our reference data in a separate PR. For the requirements we can configure a minimum pygments version and don't have a fallback.

The next 4 weeks I'm not able to do that because I'm on a trip.

mfont-bz17:
> We can update the pygments version for our test and update all our reference data in a separate PR. For the requirements we can configure a minimum pygments version and don't have a fallback.

Perfect. That's a better option. I'll try to prepare a new PR with these changes

> The next 4 weeks I'm not able to do that because I'm on a trip.

Enjoy your trip ;)

mfont-bz17:
I don't understand why is failing in **Check wheel** step...

```
nox > gcovr --html-details out.html-details
(ERROR) Error occurred while printing reports:
No module named 'gcovr.formats.html.default'nox > gcovr --html-details out.html-details
(ERROR) Error occurred while printing reports:
No module named 'gcovr.formats.html.default'
nox > Command gcovr --html-details out.html-details failed with exit code 7
nox > Command gcovr --html-details out.html-details failed with exit code 7
```

Spacetown:
> I don't understand why is failing in **Check wheel** step...
> 
> ```
> nox > gcovr --html-details out.html-details
> (ERROR) Error occurred while printing reports:
> No module named 'gcovr.formats.html.default'nox > gcovr --html-details out.html-details
> (ERROR) Error occurred while printing reports:
> No module named 'gcovr.formats.html.default'
> nox > Command gcovr --html-details out.html-details failed with exit code 7
> nox > Command gcovr --html-details out.html-details failed with exit code 7
> ```

It's not a python package. can you try to add a empty `__init__.py` to the templates directory?

Spacetown:
I fixed the problem. The theme should be used as `package_path` instead as part of the `package_name`.

mfont-bz17:
> I fixed the problem. The theme should be used as `package_path` instead as part of the `package_name`.

I pulled the changes and executing to test it and I still getting some error 
```
python3 -m nox --session 'docker_run_compiler(gcc-10)' -- -s build_wheel check_wheel
....
nox > Run all transformations to check if all the modules are packed
nox > cd /gcovr/.nox-containerized.gcc-10.uid_1000/check_wheel/tmp
nox > gcovr --cobertura out.cobertura
nox > gcovr --coveralls out.coveralls
nox > gcovr --csv out.csv
nox > gcovr --html-details out.html-details
(ERROR) Error occurred while printing reports:
style.css
nox > Command gcovr --html-details out.html-details failed with exit code 7
nox > Session check_wheel failed.
nox > Ran multiple sessions:
nox > * build_wheel: success
nox > * check_wheel: failed
nox > Command docker run --rm -it -e CC -e USE_COVERAGE -v /mnt/data/dev/gcovr/gcovr:/gcovr gcovr-qa-ubuntu_22.04-gcc-10-uid_1000 -s build_wheel check_wheel failed with exit code 1
nox > Session docker_run_compiler(gcc-10) failed.
```

Maybe I'm doing something wrong?

Spacetown:
I only checked the make test but here the error is in the wheel. The smoke test showed that not everything was packed. I've updated the paths in `setup.py` to contain the jinja2 templates.

mfont-bz17:
> I only checked the make test but here the error is in the wheel. The smoke test showed that not everything was packed. I've updated the paths in `setup.py` to contain the jinja2 templates.

Good. I've updated the windows reference tests and seems that all checks have passed ok :D

tehKaiN:
Looks like I just did almost the same by putting github's colors into generated css in a hacky way, only to discover that passing that css through command line results in pygments colors getting overridden by gcovr itself. :(

This solution is more complete and stylish than I managed to do, so I'm really looking forward to having it merged and released!

Spacetown:
@mfont-bz17 This PR is still a draft. Please publish if it's ready to merge.

@tehKaiN The CSS can already be set since end of May 2020 and was first released with version 5.0. Since may this year the pygments CSS is skipped if it's detected in the custom stylesheet but this isn't released yet..

Spacetown (approved):
LGFM

Spacetown:
Thanks for implementing this.

PR #799, opened 2023-06-19
pygments required >= 2.13.0

mfont-bz17:
* Fix pygments version 2.13.0 in tests enviroment
    * Update reference of tests using new pygments version

Related to #793


Spacetown:
You need to update the CHANGELOG.md, the authors will be updated with #793.

Spacetown:
The different files for Windows are attached as Artifacts to the run. You can download them and update your reference data.

mfont-bz17:
> The different files for Windows are attached as Artifacts to the run. You can download them and update your reference data.

I updated the reference data with the artifacts attached but still fails. Is there some way to test the windows parts locally?

Spacetown (approved):
LGFM

PR #805, opened 2023-06-30
Add support for importing Cobertura coverage file

### LINKED ISSUE #803: Add the ability to import existing coverage reports
I am working on a code base mainly consisting of C/C++, python and JS. I currently use 3 separate code coverage tools that each generate their own html. I prefer the gcovr/lcov html style, and would like to have a single place to view all coverage results.

I would like to add the ability for gcovr to import existing coverage reports. For example, I can generate Cobertura XML reports from all my code coverage tools. I would like gcovr to be able to read these reports and generate a single html output (or JSON, text, etc.)

Another project exists which is able to read an XML and generate HTML (pycobertura). However I much prefer the gcovr/lcov style.

**Additional context**
I have already done a basic proof of concept using the Cobertura XML format, so I know that it works. I am happy to do the work to integrate it into gcovr properly. For now I only need it to support XML, but there's no reason it couldn't support any other standard format.

jbloggz:
This PR adds the `--cobertura-add-tracefile` option to the CLI. This options works just like the existing `--json-add-tracefile` option, and both options can be used together.
A couple of things to note about importing a Cobertura XML

- The branch coverage doesn't contain the counts of each branch, it's either 1 or 0
- The function coverage doesn't align with how gcovr reports function coverage. Cobertura reports the percentage of the function that was covered, not the number of times it was called. For this reason, I excluded importing the function coverage.

resolves #803

This is a draft PR, as I still need to add test cases. However I am unfamiliar with the testing framework used and would appreciate some guidance in writing the tests.


gcovr/formats/cobertura/read.py
```diff
@@ -0,0 +1,141 @@
...
+    for filename in datafiles:
+        LOGGER.debug(f"Processing XML file: {filename}")
+
+        try:
+            root = etree.parse(filename).getroot()
+        except Exception:
+            raise RuntimeError(
+                "Bad --add-tracefile option.\n"
+                f"\tThe file {filename} is not a valid XML file."
+            )
```
Spacetown, line 75: Should the message of the original exception be added here? This is needed if there is a syntax error in the file.

gcovr/formats/cobertura/read.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def _line_from_xml(filename: str, xml_line) -> LineCoverage:
+    try:
+        lineno = int(xml_line.get("number"))
+        count = int(xml_line.get("hits"))
+        is_branch = xml_line.get("branch") == "true"
+        branch_msg = xml_line.get("condition-coverage")
+    except Exception:
+        LOGGER.warning(f"Invalid line information in file {filename}")
```
Spacetown, line 120: In this case I think we shouldn't catch the exception since we can't continue.

gcovr/formats/cobertura/read.py
```diff
@@ -0,0 +1,141 @@
...
+
+    line = LineCoverage(lineno, count=count)
+
+    if is_branch and branch_msg is not None:
+        try:
+            [covered, total] = branch_msg[branch_msg.find("(") + 1 : -1].split("/")
```
Spacetown, line 126: Better use
```suggestion
            [covered, total] = branch_msg[branch_msg.rfind("(") + 1 : -1].split("/")
```

Spacetown (changes requested):
Some comments. Please add yourself to the Authors, update the changeling and add a test case.

gcovr/formats/cobertura/read.py
jbloggz: Yes, you're right. I just checked it for various types of errors (missing file/corrupt file) and the message from lxml contains the file name as well, so I can use that instead.

gcovr/formats/cobertura/read.py
jbloggz: I think the raw exception message could be a bit vague. So I'll update it to rethrow with a nicer message, and provide the xml line that caused the issue.

Spacetown: Or add the raw message to the existing one.

gcovr/formats/cobertura/read.py
jbloggz: Are you suggesting this for effciency. It looks like `rfind` will have less work to do when 100% of the branches are covered. Otherwise `find` is as good or better in all other cases. Happy to change it, just curious about the reason.

gcovr/formats/cobertura/read.py
jbloggz: I mean this:
```suggestion
        raise RuntimeError(
            "Bad --covertura-add-tracefile option.\n"
            f"\tCorrupt/missing line coverage: {etree.tostring(xml_line).decode()}\n"
            f"\t{e}"
        )
```
So give a nicer looking message, with the line that caused it, but include the raw message at the end. How does that look?

Spacetown: You're right, here can several errors occur. E.g. the attribute doesn't exist or has a different type. Would it be better to read each attribute in a separate try block?

jbloggz: On closer inspection, only the 'int' conversion will throw here. A missing attribute will return None. I'll check those two conversions separately and give a specific message

gcovr/formats/cobertura/read.py
Spacetown: It's more for security. You need the content of the braced block at the end and with refund you're independent from the content before. With refund the content before can also contain a open brace without affecting the result.

gcovr/tests/test_args.py
```diff
@@ -494,6 +494,118 @@ def test_html_injection_via_json(capsys, tmp_path):
...
+    with tempfile.open("w+") as fp:
+        fp.write(xmldata)
+
+    filename = str(tempfile)
+    opts = merge_options_and_set_defaults(
+        [{"cobertura_add_tracefile": [filename], "filter": [re.compile(".")]}]
+    )
+    covdata = read_reports(opts)
+    assert covdata is not None
+    assert "/path/to/source/code.cpp" in covdata
```
Spacetown, line 549: Better use a variable for this string. this variable can also be interpolated into the here document.

gcovr/tests/test_args.py
```diff
@@ -494,6 +494,118 @@ def test_html_injection_via_json(capsys, tmp_path):
...
+    fcov: FileCoverage = covdata["/path/to/source/code.cpp"]
+    assert len(fcov.lines) == 10
+    assert fcov.lines[7].count == 1
```
Spacetown, line 552: Is it better to loop over a list with tuples `(lineno, hits, len branches or None)`?

Spacetown (changes requested):

gcovr/tests/test_args.py
jbloggz: I've updated the PR with your suggestion

gcovr/tests/test_args.py
Spacetown: Something like:
```python
    for line, count in [(7, 1), (9, 3), (16, 0)]:
        assert fcov.lines[line].count == count
```

jbloggz: I've updated the test with your suggestion

Spacetown (approved):
LGFM

Spacetown (changes requested):
Please can you add a test case for this?

jbloggz:
> Please can you add a test case for this?

I did already add a test case. Are you wanting another one? Or are you saying the current test case doesn't have enough coverage?

Spacetown:
Sorry, I missed this test.

Spacetown (approved):
LGFM

PR #817, opened 2023-08-03
Improve sorting of data in reports

### LINKED ISSUE #810: Show uncovered lines first in a report

Code coverage reports are need for elimination of places in a source code which coverage is worse.
Gcovr has options `--sort-percentage` and `--sort-uncovered`, but when these options are enabled report shows lines with better coverage first.

Sort lines in reverse order, show lines with worse coverage first.

lcov has JS code embedded to HTML report that can sort columns in both orders.

**Additional context**

Enabled `--sort-percentage` and `--sort-uncovered`:

image

Enabled `--sort-percentage`:

image

Spacetown:
- Sort file names alpha numerical (`file_10.c` comes after `file_0.c`).
- Always sort at the end by filename if line or branch coverage is identical for a file.
- Add `--sort-branches` to sort by branches instead of lines, this is the default if `--branch` is used.
- Add `--sort-decreasing` to start with the files with lower coverage values.

Closes #810


ligurio:
Patches version: 697eead9f704f0e5351cc7edc201c395214d896b, branch: `Spacetown:change_sorting_in_reports`

With `--sort-decreasing` sorting works in unexpected order:

image

With `--sort-decreasing` and `--sort-percentage`:

image

Spacetown:
@ligurio What do you think? Can this be merged or is a change needed?

ligurio:
This doesn't work for me at all.

`--sort-uncovered`:

image

`--sort-uncovered`:

image

Other used options:

```
        # See https://gcovr.com/en/stable/guide/configuration.html                 
        --root ${PROJECT_SOURCE_DIR}                                               
        --object-directory ${PROJECT_BINARY_DIR}                                   
        --filter ${PROJECT_SOURCE_DIR}/src                                         
        # Exclude DynASM files, that contain a low-level VM code for CPUs.         
        --exclude ".*\.dasc"                                                       
        # Exclude buildvm source code, it's the project's infrastructure.          
        --exclude ".*/host/"                                                       
        --print-summary                                                            
        --output ${COVERAGE_HTML_REPORT}                                           
        --cobertura ${COVERAGE_XML_REPORT}                                         
        --html                                                                     
        --html-title "Tarantool LuaJIT Code Coverage Report"                       
        --html-details                                                             
        --branches                                                                 
        --decisions                                                                
        -j ${CMAKE_BUILD_PARALLEL_LEVEL}  
```

Let me remind a goal in the original issue: sort uncovered lines (or branches) by descending order. I have tried all options in different combinations and all of them are not suitable for me and doesn't help to reach a goal.

You decided to introduce new options and this makes usage more complicated for users. Take a look a usage with sorting options, it is not trivial to get all of them, some of them are mutual exclusive:

```
  --sort-branches       Sort entries by branches instead of lines. Can only be
                        used together with --sort-uncovered or --sort-percent
                        is used. Config key(s): sort-branches.
  -u, --sort-uncovered  Sort entries by number of uncovered lines (if the
                        option --sort-branches is given the uncovered branches
                        are used). The default order is increasing and can be
                        changed by --sort-reverse. The secondary sort key (if
                        values are identical) is always the ascending
                        filename. For CSV, HTML, JSON and text report. Config
                        key(s): sort-uncovered.
  -p, --sort-percentage
                        Sort entries by percentage of uncovered lines (if the
                        option --sort-branches is given the percentage of
                        uncovered branches are used). The default order is
                        decreasing and can be changed by --sort-reverse.The
                        secondary sort key (if values are identical) is always
                        the ascending filename. For CSV, HTML, JSON and text
                        report. Config key(s): sort-percentage.
  --sort-reverse        Sort entries in reverse order. For CSV, HTML, JSON and
                        text report. Config key(s): sort-reverse.
  --sort-casefold       Sort file names with case folding. For CSV, HTML, JSON
                        and text report. Config key(s): sort-casefold.
```
I suggest removing options for separate sorting by branches/lines/functions etc. and left only option sort-by-ascending that will sort by ascending all columns and enable sorting by descending by default.

Spacetown:
In the command line you've the option `--branches` which switches from sorting by line coverage to branch coverage. This behavior only make sense for the text format since the CSV, JSON and HTML always contain branch information and should use the new option `--sort-branches` instead.
The option `--sort-casefold`can be the default, removing the option again.

The default sort order for `sort-uncovered` and `--sort-precentage` wasn't changed only a option for changing the direction was added.

Both of your screenshots have the same option mentioned and since you'r having the option `--branches` the sorting is done by branches instead of lines. 

If you use `--sort-uncovered`, your first screenshot is correct since the uncovered branches are sorted from 974-342=632 to 110-74=36.

The default sort order for `sort-uncovered` and `--sort-precentage` wasn't changed only a option for changing the direction was added.

Both of your screenshots have the same option mentioned and since you'r having the option `--branches` the sorting is done by branches instead of lines. 

If you use `--sort-uncovered`, your first screenshot is correct since the uncovered branches are sorted from 974-342=632 to 110-74=36.

Spacetown:
> I suggest removing options for separate sorting by branches/lines/functions etc. and left only option sort-by-ascending that will sort by ascending all columns and enable sorting by descending by default.

But which column shall than be the sorted one?

We need following options:
- One to select between default, line or branch coverage.
- One to select the absolute uncovered stuff or the percentage values.
- One to change between default order or reversed order. Here we have the problem that for the absolute uncovered values ascending is the default order and for the percentage descending is the default.

Spacetown:
@ligurio Any notes from you?

PR #818, opened 2023-08-04
Update html.rst to add --html-theme github.* screenshots

mfont-bz17:
I 've added some screenshots of github themes directly in documentation html.rst but maybe a new child could be added 

Output Formats 
    HTML Output
           Themes 

What do you think?


Spacetown:
I suggest to use the same file as for the default theme because it's smaller and you can see the differences better. Can you also inject the obfuscated date/time?

Spacetown (approved):
This looks good. The only thing I see now is that the boxes in the right upper corner should have a fixed size since the width depends on the needed digits of the value (`100.0%` is bigger). Do you want to fix it together with this PR or in a separate PR?

mfont-bz17:
Now the width is fixed using wrapper and the background depends on the digits, in all places.

**header** 
imatge

**list**
imatge

I created some examples with fixed width label. If 100.0% not appears we have always an extra 

**header**
imatge
imatge
imatge


**list**
imatge


In the list part I prefer keep as now, in the header part maybe can be changed, I'm not sure what is better....

Spacetown:
> I created some examples with fixed width label. If 100.0% not appears we have always an extra
> 
> **header** imatge imatge imatge
> 
> **list** imatge
> 
> In the list part I prefer keep as now, in the header part maybe can be changed, I'm not sure what is better....

I prefer this since it looks calmer.

mfont-bz17:
Pushed with the fixed size

Spacetown (approved):
LGFM

PR #830, opened 2023-09-19
Add lcov output format

### LINKED ISSUE #523: Support the LCOV-Info output format
The LCOV Info format is a widely used coverage format. It would be great if Gcovr can generate Info-files so that more downstream tools can be used. For example, #522 showed that the Coveralls Github Action expected LCOV-Info files, which currently makes it impossible to use gcovr together with that Action.

Expected solution: There is an `--lcov` output format.

Design constraint: being able to *parse* LCOV-Info would also be very nice, but is more complicated. The naming of command line options should make it possible to have both input + output LCOV files in the future.

The Lcov-Info format is documented in [`man 1 geninfo`](http://ltp.sourceforge.net/coverage/lcov/geninfo.1.php). It is a simple text-based format.

Spacetown:
This PR adds a writer for the LCOV info format as requested in #523.
The optional version line is filled with the MD5 of the file content.

Add scrubber for test suite to remove the root to the repository from output files.

Closes #523


Spacetown:
@latk In the linked man page [`man 1 geninfo`](http://ltp.sourceforge.net/coverage/lcov/geninfo.1.php) of #523 there is no `VER:...` but in the man page I used it's added. The VSCode plugin I doesn't understand this line. Shall I comment it since it'sto new and some tools wouldn't understand the file?

latk:
> In the linked man page [`man 1 geninfo`](http://ltp.sourceforge.net/coverage/lcov/geninfo.1.php) of #523 there is no `VER:...` but in the man page I used it's added. The VSCode plugin I doesn't understand this line. Shall I comment it since it'sto new and some tools wouldn't understand the file?

The optional `VER:` field was added to `man 1 geninfo` fairly recently (this January) in this commit: https://github.com/linux-test-project/lcov/commit/d98659aa42397760379ac948be1caadb1cf8eeb6

I think either approach for dealing with this is OK:

* It is perfectly reasonable to say that this field is optional, and for maximum compatibility with strict lcov-info consumers gcovr won't emit it.
* It is also perfectly reasonable to target a specific lcov-info format version – gcovr can't support all historical versions of that format, just like gcovr doesn't support ancient GCC versions. We might as well target the newest format existing *now*, emit the field, and the rest of the ecosystem will eventually catch up.

Personally, I'd tend towards the latter, for two reasons:

* Lcov itself will emit this field, so we should as well (if this can be done in a compatible manner).
* Tools that consume the lcov-info format should skip unrecognized lines, taking into account Postel's law. Similarly, gcovr can skip over unrecognized gcov lines.

Spacetown:
If there are problems with the field we can add an option to skip it.

Spacetown:
@latk  I've added an upload of a reference file to Codecov.

PR #832, opened 2023-09-21
Setup test case for coexisting object directories

### LINKED ISSUE #756: Failure when several gcovr instances are launched concurrently
**Describe the bug**
The issue occurs in our CI (jenkins), when trying to generate reports for several coverage data at the same time (one report per configuration).

**To Reproduce**

1. gitclone the project, the root directory contains (mainly) CMakeLists.txt, source folder
2. build the source for several configurations, the root directory now contains builds/conf1, builds/conf2, ...
3. launch tests in parallel  to produce coverage data, gcda/gcno files are in the builds/conf1, builds/conf2, etc...
4. launch gcovr in parallel to produce coverage reports, see it fails
  - command line (that can probably  be improved ?) for configX is like  
```bash
gcovr \
  --gcov-executable /usr/bin/gcov \
  --root /my/project/root \
  --object-directory /my/project/root/builds/configX \
  --exclude ".*/tests/src/.*" \
  --exclude ".*/unit_tests/src/.*" \
  --exclude ".*/func_tests/src/.*" \
  --branches \
  --exclude-unreachable-branches \
  --exclude-throw-branches \
  --sort-percentage \
  --html --html-details /my/project/root/builds/configX/coverage-reports/gcovr/index.html \
  --xml --xml-pretty --output /my/project/root/builds/configX/coverage-reports/gcov-results.xml \
  # end of args
```

**Expected behavior**
Calling gcovr on different object directory, at the same time, should not be an issue.

Gcovr 5.0 did not have the issue. Tracing behaviour changes (with `--verbose` call shows that gcovr 5.0 calls gcov in some temporary directory

```
Running gcov: '/usr/bin/gcov /my/project/root/builds/configX/src/lib1/CMakeFiles/mylib.dir/src/SomeSrc.cpp.gcda \
	--branch-counts \
	--branch-probabilities \
	--preserve-paths \
	--object-directory /my/project/root/builds/configX/src/lib1/CMakeFiles/mylib.dir/src' in '/tmp/tmp1sck169l'
```

while 5.1 and later calls gcov in the project source tree (that is quite ... unpretty ;)) ...
```
(DEBUG) Running gcov: '/usr/bin/gcov /my/project/root/builds/configX/src/lib1/CMakeFiles/mylib.dir/src/SomeSrc.cpp.gcda \
	--branch-counts \
	--branch-probabilities \
	--demangled-names \
	--hash-filenames \
	--object-directory /my/project/root/builds/configX/src/lib1/CMakeFiles/mylib.dir/src' in '/my/project/root'
```
Working directly in the (same) source tree should then trigger some race condition between gcov generated .gcov files 

**Screenshots**

 - OS: Linux Debian 11.6 (same behaviour on 9.13)
 - GCC version 10.3.0 
 - GCOVR version 5.1, 5.2, 6.0
 - Project directory layout:
```
.                                      <- root
├── builds 
│   ├── config1                        <- object-directory for config1
│   │   ├── cmake
│   │   ├── CMakeFiles
│   │   ├── coverage-reports
│   │   │   ├── gcovr
│   │   │   └── lcov
│   │   └── src
│   └── config2                        <- object-directory for config2
│       ├── cmake
│       ├── CMakeFiles
│       ├── coverage-reports
│       │   ├── gcovr
│       │   └── lcov
│       └── src
├── cmake
├── scripts
└── src
    ├── exe
    ├── lib1
    ├── lib2
    └── unit_tests
```  

**Additional context**
Note that enabling/disabling gcov parallel calls does not modify the behaviour.

oleurodecision:
The PR aims to present a basic setup for testing issue(s) related to coexisting object directories.

The _candidate_tests.bash_ script contains several valid or failing gcovr call variants (parallelised or non-parallelised calls).

Related to #756.
Closes #756.

Please ask for any piece of information needed on the subject.


Spacetown:
Thanks for the test. Please can you use the extension `.cpp` instead of `. cc` to be aligned to the other tests?

oleurodecision:
My bad! Twenty years of habits ;)

gcovr/tests/test_gcovr.py
```diff
@@ -107,7 +107,7 @@
 )
 
 RE_HTML_ATTRS = re.compile('((timestamp)|(version))="[^"]*"')
-RE_HTML_HEADER_DATE = re.compile(r"(<td)>\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d<(/td>)")
+RE_HTML_HEADER_DATE = re.compile(r"(<td)>\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d(\+\d\d:\d\d)?<(/td>)")
```
Spacetown, line 110: Use non captured group for better performance:
```suggestion
RE_HTML_HEADER_DATE = re.compile(r"(<td)>\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d(?:\+\d\d:\d\d)?<(/td>)")
```

gcovr/tests/test_gcovr.py
```diff
@@ -127,7 +127,7 @@ def scrub_xml(contents):
 
 def scrub_html(contents):
     contents = RE_HTML_ATTRS.sub('\\1=""', contents)
-    contents = RE_HTML_HEADER_DATE.sub("\\1>0000-00-00 00:00:00<\\2", contents)
+    contents = RE_HTML_HEADER_DATE.sub("\\1>0000-00-00 00:00:00<\\3", contents)
```
Spacetown, line 130: Offset isn't captured anymore:
```suggestion
    contents = RE_HTML_HEADER_DATE.sub("\\1>0000-00-00 00:00:00<\\2", contents)
```

gcovr/tests/coexisting_object_directories/candidate_tests.bash
Spacetown, line 1: What's the purpose of this script (it seems not to be called in test)?

Please add a description to the script.

Spacetown (changes requested):

Spacetown:
Please add a entry in the CHANGELOG.rst.

gcovr/tests/coexisting_object_directories/candidate_tests.bash
oleurodecision: The script was added as a reference of failing (and working) cases. The PR only set up what is required for exploring, testing these cases.
Do you want me to add several test cases ? If so, what is the way to do it ? One test folder per case ?

(BTW, thanks for the changelog update, I can only work on the subject on Thursday and Wednesday)

Spacetown: We've one folder per test case but this script is never executed. We only execute the makefiles.

oleurodecision: - script was removed
- dedicated folders were added for each test (the ones described in the removed script)

Thanks waiting for the changes

Spacetown (changes requested):
I thinkl `gcovr` isn't executed in parallel since the `-j` option is missing when calling make. I`ve adapted the test environment on the branch https://github.com/Spacetown/gcovr/tree/feature/coexisting_object_directories I've modified your tests and the tests coexisting_object_directories-from_build_dir-without_search_dir and coexisting_object_directories-from_root_dir-without_object_dir are failing because of concurency.

Please have a look.

oleurodecision:
Thanks for upgrading the test environment, this will avoid the dirty call I previously had to do in the makefiles themselves  :disappointed: 

The concurrency issue is now fixed using some dedicated rules (`run.%.done`) to avoid running the same binary several times.

Now (well, at least on my local tests), only the expected tests (the `-without` ones) fail (!)

Spacetown:
> The concurrency issue is now fixed using some dedicated rules (`run.%.done`) to avoid running the same binary several times.

I don't get what this additional rule is fixing. The result should be the same. I need to check this the next days.

Spacetown:
The additional rule doesn't change anything. If I understand you correct the tests "without" should fail because of the parallel execution. This we can't test here. For the test `coexisting_object_directories-from_root_dir-without_search_dir` there is a workaround with `GCOV_STRIP` and  `GCOVR_PREFIX`.

Spacetown (approved):
I've updated the tests to get them green.

@oleurodecision Please can you check again from your side?

Spacetown:
@oleurodecision Have you already checked it?

Spacetown:
@oleurodecision I'm waiting for your response to merge this.

oleurodecision:
Thanks for waiting, and sorry for the late, my wife has been (and is still) very illed, and I am facing lack of time to seriously work on the subject. For what I quickly reviewed, it seems ok, so let's go !

Once again, thanks for the tool and the time spent on it.

PR #840, opened 2023-10-11
Link function list by CSS to reduce size.

### LINKED ISSUE #837: Functions file can be huge, and of not much benefit
In our experience, with a large code base, the functions-html output is not worthy of the disk space that's required to save it. We would like the option in the config-options to prevent it from being written (or data-gathered).

In our case, with 1.7G of HTML files, we had 496Mb for this functions file. Our reports are eating a lot of disk. Meanwhile, the output looks like this:
```
    <td>
      <a href="index.dockets_graph_lib.cc.9eac23e5172e559aee87b169b08fb79b.html#l445">_Z11run_docketsRKN4Argo5tools7dockets20DocketsGraphSettingsE</a>
    </td>
```
... with filenames and call-count, and this isn't particularly valuable when this list is huge. It's too difficult to filter through, especially with mangled names. So we'd like the option to turn it off.

(maybe demangling the names could be another issue, but not desired by us)

In the command line arguments, we want a way to prevent this output. --no-functions-report or something like it. That should apply to each relevant generator format. We would also expect the link(in the HTML) to go away. 

There aren't good alternatives to this. This needs a code change to prevent the excessive disk use.

**Additional context**

Spacetown:
Add only one link per function and display it as block over the whole row. This reduces the size for each line in the function overview from 452 to 166 characters per function (excluding function and file name).

Also add the highlighting style of the current table row in the GitHub theme.

Closes #837


latk:
If I understand this change correctly, it makes the `<a>` link element empty, and then uses CSS to overlay an invisible click target over the contents of the table cell?

This feels fairly hacky, and might make things worse for some users. For example, we're trading file size versus accessibility.

Perhaps a compromise would be to only link one of the cells, but to keep using a normal link without CSS magic there. Right now we have multiple columns, all with their individual links:

* function name
* file name
* line number
* call count
* block coverage count

I would suggest to merge file name + line number into one cell, and to either link only the function name or only the location (file name + line number). Linking the counts seems a lot less important and could reasonably be removed.

It would also be possible to stop using a HTML table, and instead have a list of clickable entries, where each entry contains multiple types of information (perhaps on multiple lines). Mockup:

2023-10-12-function-list-mockup

<details><summary>mockup source code</summary>

```html
<html style="background: black;"><head></head><body style="display: flex;flex-flow: column;gap: 1em;font-family: sans-serif;margin: 1em auto;max-width: 40em;">
  <div class="card medium">
    <div class="fn">example::function()</div>
    <div class="loc">at <span class="file">example.cpp</span>:<span class="line">12</span></div>
    <div class="call">called 5 times</div>
    <div class="block">15 blocks</div>
  </div>
  <div class="card good">
    <div class="fn">another::example()</div>
    <div class="loc">at <span class="file">example.cpp</span>:<span class="line">102</span></div>
    <div class="call">called 100 times</div>
    <div class="block">2 blocks</div>
  </div>
</body><style type="text/css"></style></html>
```

To make it clickable, the `div class="card">` would become an `<a>` link.

```css
.card {
  border-radius: 1em;
  /*! margin: 1em; */
  background: #ddd;
  padding: 1em;
  display: flex;
  flex-flow: row wrap;
  gap: 0 1em;
}
.fn {
  width: 100%;
  font-size: 1.2em;
  font-family: "Source code pro";
}
.loc {
  color: #333;
  width: 100%;
}
.call, .block {
  /*! margin: ; */
  color: #333;
}
.medium {
  background: #dfdf27;
}
.good {
  background: #20df20;
}
.file, .line, .fn {
  font-family: "Source code pro";
}
```

</details>

Spacetown:
> f I understand this change correctly, it makes the `<a>` link element empty, and then uses CSS to overlay an invisible click target over the contents of the table cell?

Yes.

> This feels fairly hacky, and might make things worse for some users. For example, we're trading file size versus accessibility.

I thought that the accessibility would be better if the whole line is clickable and not only the text as it is now.

> I would suggest to merge file name + line number into one cell, and to either link only the function name or only the location (file name + line number). Linking the counts seems a lot less important and could reasonably be removed.

That's also an idea but the link style is removed and the user can't see which part is linked. We can also merge the function and whole location together and ling this part:

```html
<a href="...">example::function() defined at example.cpp:12</a>
```

> It would also be possible to stop using a HTML table, and instead have a list of clickable entries, where each entry contains multiple types of information (perhaps on multiple lines).

This looks great but there are many div and span elements which also cost size.

Do we need the file information in the function overview at all? In my opinion it's used to jump to the place where the function is defined, there you have the file information visible. Maybe we can remove this information from the overview.

Spacetown:
@macetw Please can you check how much the size is reduced by this branch?

An example for the overall list is:
Bildschirmfoto 2023-10-31 um 22 35 17

And an example of the list for a source file is:
grafik

macetw:
@Spacetown I am not going to be able to test this out until a couple weeks. Could I get back to you then?

Spacetown:
@macetw I want to create a release in December. The size is reduced for sure. Can you go with the change of the list forma in the screenshots above?

macetw:
@Spacetown The screenshots you provided look great!

In our case, with such a large project over so many directories and files, I can't even see the function table -- my browser choke on it with 625142 functions total. The Chrome browser can't handle creating a table that is so large. ... Hm, could a table-break be created after every 10,000 entries?

gcovr/formats/html/default/functions_page.content.html
```diff
@@ -1,36 +1,23 @@
...
-        {%- endif -%}
-      </a>
-    </td>
-    <td>
-      <a href="#l{{ entry['line'] }}">{{ entry["blocks"] }}%</a>
-    </td>
-</tr>
+    {#- #}<td><a href="{{ entry['html_filename'] }}#l{{ entry['line'] }}">{{ entry["name"] }} ({{ entry["filename"] }}:{{ entry["line"] }})</a></td>
+    {#- #}<td>
+            {%- if entry["excluded"] %} excluded
```
macetw, line 12: Suggest removing the whitespace before the word excluded.

Also, whitespace on lines 14 and 15.

gcovr/formats/html/default/source_page.navigation.html
```diff
@@ -4,32 +4,23 @@
...
-      </td>
-      <td>
-        <a href="#l{{ entry['line'] }}">{{ entry["blocks"] }}%</a>
-      </td>
+      {#- #}<td><a href="#l{{ entry['line'] }}">{{ entry["name"] }} (line {{ entry["line"] }})</a></td>
+      {#- #}<td>
+                {%- if entry["excluded"] %} excluded
+                {%- else %}
+                {%-  if entry["count"] == 0 %} not called
+                {%-  else %} called {{ entry["count"] }} time{% if entry["count"] > 1 %}s{% endif %}, returned {{ entry["returned"] }} time{% if entry["returned"] > 1 %}s{% endif %}
```
macetw, line 18:
```suggestion
                {%- if entry["excluded"] %}excluded
                {%- else %}
                {%-  if entry["count"] == 0 %}not called
                {%-  else %}called {{ entry["count"] }} time{% if entry["count"] > 1 %}s{% endif %}, returned {{ entry["returned"] }} time{% if entry["returned"] > 1 %}s{% endif %}
```

gcovr/formats/html/github/functions_page.content.html
```diff
@@ -1,36 +1,23 @@
...
-    <td>
-      <a href="#l{{ entry['line'] }}">{{ entry["blocks"] }}%</a>
-    </td>
-</tr>
+    {#- #}<td><a href="{{ entry['html_filename'] }}#l{{ entry['line'] }}">{{ entry["name"] }} ({{ entry["filename"] }}:{{ entry["line"] }})</a></td>
+    {#- #}<td>
+            {%- if entry["excluded"] %} excluded
+            {%- else %}
+            {%-  if entry["count"] == 0 %} not called
+            {%-  else %} called {{ entry["count"] }} time{% if entry["count"] > 1 %}s{% endif %}, returned {{ entry["returned"] }} time{% if entry["returned"] > 1 %}s{% endif %}
```
macetw, line 15:
```suggestion
            {%- if entry["excluded"] %}excluded
            {%- else %}
            {%-  if entry["count"] == 0 %}not called
            {%-  else %}called {{ entry["count"] }} time{% if entry["count"] > 1 %}s{% endif %}, returned {{ entry["returned"] }} time{% if entry["returned"] > 1 %}s{% endif %}
```

gcovr/formats/html/github/source_page.navigation.html
```diff
@@ -3,34 +3,25 @@
...
-      </td>
-      <td>
-        <a href="#l{{ entry['line'] }}">{{ entry["blocks"] }}%</a>
-      </td>
+      {#- #}<td><a href="#l{{ entry['line'] }}">{{ entry["name"] }} (line {{ entry["line"] }})</a></td>
+      {#- #}<td>
+                {%- if entry["excluded"] %} excluded
+                {%- else %}
+                {%-  if entry["count"] == 0 %} not called
+                {%-  else %} called {{ entry["count"] }} time{% if entry["count"] > 1 %}s{% endif %}, returned {{ entry["returned"] }} time{% if entry["returned"] > 1 %}s{% endif %}
```
macetw, line 19:
```suggestion
                {%- if entry["excluded"] %}excluded
                {%- else %}
                {%-  if entry["count"] == 0 %}not called
                {%-  else %}called {{ entry["count"] }} time{% if entry["count"] > 1 %}s{% endif %}, returned {{ entry["returned"] }} time{% if entry["returned"] > 1 %}s{% endif %}
```

macetw:
@Spacetown Please consider the linked PR. I cannot base it of this branch, because the forked repo could not be chosen as a base in this PR.

PR #849, opened 2023-11-11
add support for nvpl-lapack, solved three problems:

gaosongyun:
1. since gcov would issue warnings like the following /gitlab/nvpl-lapack/build/nvpl-ci/src/iface/CMakeFiles/nvpl_lapack_lp64_gomp.dir/larfg.cpp.gcno:version 'B32*', prefer 'B23*' /gitlab/nvpl-lapack/build/nvpl-ci/src/iface/CMakeFiles/nvpl_lapack_lp64_gomp.dir/larfg.cpp.gcda:version 'B32*', prefer version 'B23*' and may return errorcode=3 because of this, gcovr should not throw exception upon such "success with info"
2. we used cross-build for our nvpl-lapack project, which caused the gcov artifacts(*.gcno, gcda) to be detached from the source code (i.e. they are not in the same directory), such an situation can not be handled properly by gcovr, so we change the filter from options.filter(DirectoryPrefixFilter) to options.gcov_filter(AlwaysMatchFilter)
3. removed unnecessary blocking in the worker(...) function, because python's native Queue class can already handle the concurrency correctly by using condition varialbles. we can further remove the workers.lock for the same reason, but I decided to leave it as is for the time being.

the command line to generate the coverage report for nvpl-lapack:
	gcovr -r /gitlab/nvpl-lapack/build/nvpl-ci/src/ --verbose --html-details -o coverage.html --html-title 'nvpl_lapack GCC Code Coverage Report'gcov --gcov-ignore-parse-errors negative_hits.warn --merge-mode-functions=separate

### DIFF
--- a/gcovr/gcov.py
+++ b/gcovr/gcov.py
@@ -124,7 +124,7 @@ def process_gcov_data(
     # Return if the filename does not match the filter
     # Return if the filename matches the exclude pattern
     filtered, excluded = apply_filter_include_exclude(
-        fname, options.filter, options.exclude
+        fname, options.gcov_filter, options.exclude
     )
 
     if filtered:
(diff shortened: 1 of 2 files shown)


latk:
Hi gaosongyun,

I'm a bit confused by this because the code does not match your pull request description.

But all things you mentioned would be great things to improve – if you find a way, please do improve them and create a new PR. For example, better support for cross-compilation would be really nice.

Issue #857, opened 2023-12-01
Large function lists break a browser


macetw:
In my case, I use gcov with a large project over  many directories and files. With so many functions to list, I can't even see the function table -- my browser chokes on it. Chrome is trying to render a table with 625142 (+1) lines. The Chrome browser can't handle rendering an HTML table that is so big.

Could a table-break be created after every 1,000 or 10,000 entries? This would give a browser the breathing room to render tables.

(eventually, in my case, Chrome does this ...)
![Screenshot from 2023-12-01 09-30-17](https://github.com/gcovr/gcovr/assets/3484933/517ad0ed-db28-421e-bd2d-a401114670a2)

_Originally posted by @macetw in https://github.com/gcovr/gcovr/issues/840#issuecomment-1836505320_

I am using gcovr 6.0 and Chrome 119.0.6045.159 on Ubuntu Linux 64bit

Spacetown:
Closed because #858 is merged.

PR #858, opened 2023-12-01
Prevent Chrome from crashing on large function lists.

macetw:
Designed to fix this issue: https://github.com/gcovr/gcovr/issues/857

Designed to base on top of this PR: https://github.com/gcovr/gcovr/pull/840

Work remaining: update reference HTML files. (which was also needed on commits from PR 840)


Spacetown:
@macetw I'll add this suggestion to the existing PR.

Spacetown (approved):
LGFM. Please can you check the change from `{% if loop.first or loop.index % 10000 == 1 %}` to `{% if loop.index0 % 10000 == 0 %}`.

macetw:
> LGFM. Please can you check the change from `{% if loop.first or loop.index % 10000 == 1 %}` to `{% if loop.index0 % 10000 == 0 %}`.

I will test. Honestly, my last test didn't actually test it in Chrome -- I tested with very small sample sizes. Today, I realized how to test into our huge coverage, it will take a little time to do this test to get all the data.

Spacetown:
@macetw If this PR is merged I want to create a release.

Spacetown:
@macetw Have you found time to test this?

gcovr/formats/html/default/functions_page.content.html
```diff
@@ -1,11 +1,12 @@
 {# -*- engine: jinja -*- #}
+{% for entry in all_functions %}
+{% if loop.index0 % 10000 == 0 %}
```
latk, line 3: Instead of checking the index and emitting `<table>`/`</table>` tags as necessary, this is probably much clearer to express with the `batch` filter: https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.batch

Roughly:

```jinja
{% for function_batch in all_functions | batch(10000) %}
<table ...>
{% for entry in function_batch %}
...
{% endfor %}
</table>
{% endfor %}
```

Spacetown: I didn't know this filter. It's really clearer.

Spacetown:
Thanks for working on this.

PR #864, opened 2023-12-27
[Decisions] Added optional txt-summary print of decision coverage

Cirromulus:
Decision coverage is not printed to txt currently, although the description of the `--decisions` switch mention it:

> "Report the decision coverage. For HTML, JSON, and the summary report."

This PR changes that by adding a decision flag to the TXT writer.


gcovr/coverage.py
```diff
@@ -66,6 +69,12 @@ def sort_coverage(
 
     returns: the sorted keys
     """
+
+    if by_branch and by_decision:
+        raise RuntimeWarning("Sorting by coverage and by decision is mutually exclusive")
```
Spacetown, line 74:
```suggestion
        raise RuntimeWarning("Sorting by branch coverage and by decision coverage is mutually exclusive")
```

gcovr/coverage.py
```diff
@@ -337,6 +348,20 @@ def is_uncovered(self) -> bool:
...
+
+        if isinstance(self.decision, DecisionCoverageUncheckable):
+            return False
+
+        if isinstance(self.decision, DecisionCoverageConditional):
+            return self.decision.count_true == 0 or self.decision.count_false == 0
+
+        if isinstance(self.decision, DecisionCoverageSwitch):
+            return self.decision.count == 0
+
```
Spacetown, line 364:
```suggestion

raise RuntimeError("Decision type needs to be added.")

```

gcovr/formats/txt/write.py
```diff
@@ -54,8 +57,15 @@ def write_report(covdata: CovData, output_file: str, options: Options) -> None:
...
-        title_total = "Branches" if options.txt_use_branch_coverage else "Lines"
-        title_covered = "Taken" if options.txt_use_branch_coverage else "Exec"
+        title_total = "Lines"
+        title_covered = "Exec"
+        if options.txt_use_branch_coverage:
+            title_total = "Branches"
+            title_covered = "Taken"
+        elif options.show_decision:
+            title_total = "Decisions"
+            title_covered = "Taken"
```
Spacetown, line 67: Unneeded initialization possible.
```suggestion
        if options.txt_use_branch_coverage:
            title_total = "Branches"
            title_covered = "Taken"
        elif options.show_decision:
            title_total = "Decisions"
            title_covered = "Taken"
        else:
            title_total = "Lines"
            title_covered = "Exec"
```

gcovr/coverage.py
```diff
@@ -48,16 +48,19 @@
...
     covdata: CovData,
-    by_branch: bool,
-    by_num_uncovered: bool,
-    by_percent_uncovered: bool,
     reverse: bool,
+    *, # end of positional arguments
+    by_branch: bool = False,
+    by_decision: bool = False,
+    by_num_uncovered: bool = False,
+    by_percent_uncovered: bool = False,
```
Spacetown, line 56: The forced keyword arguments is a good idea but the intention here is that all arguments need to be supplied and this would not be possible anymore if we use default values.
```suggestion
    by_branch: bool,
    by_decision: bool,
    by_num_uncovered: bool,
    by_percent_uncovered: bool,
```

Spacetown (changes requested):

Spacetown:
@Cirromulus The description of the option is correct, it's used in the summary report.
https://github.com/gcovr/gcovr/blob/8e1a41074f7e2966cd436588ff134b75cb3bc483/gcovr/configuration.py#L472
https://github.com/gcovr/gcovr/blob/8e1a41074f7e2966cd436588ff134b75cb3bc483/gcovr/formats/txt/write.py#L111-L112

Cirromulus:
> @Cirromulus The description of the option is correct, it's used in the summary report.

Ah sorry, then I misunderstood the help text.

Spacetown:
@latk What do you think about this change?

gcovr/coverage.py
Cirromulus: This would mean that a lot of other writers will need a change. But still if this is desired, I can do this.

latk:
> What do you think about this change?

This is a logical change. The txt-report doesn't really care what metric it uses, so line/branch/decision should all work. Thank you @Cirromulus for working on this!

Instead of using multiple mutually exclusive boolean flags to select metric type, could use a single variable like `metric: Literal["line", "branch", "decision"] = "line"`. Perhaps this would also be a better CLI interface.

Not a fan of reusing a global option that affects multiple report formats. Going forward, namespacing such options under the report format name might make more sense.

Personally, I don't like the decision coverage analyzer. It is a feature that many users seem to want, but it is incredibly buggy.

Spacetown:
@Cirromulus I suggest that I first change the options to specify the desired metric type and then you can merge this into your branch.

Cirromulus:
> @Cirromulus I suggest that I first change the options to specify the desired metric type and then you can merge this into your branch.

This is fine with me. Thank you.

Spacetown:
@Cirromulus Please can you also take a look at the linked PR?

Spacetown (approved):

Spacetown:
@Cirromulus Thanks for working on this.

PR #867, opened 2024-01-08
Add report type enumeration

Spacetown:
This PR is a preparation for #864 to use a literal definition for sorting and also for the values used for the TXT report.

The option flags `--sort-uncovered`, `--sort-percentage`, `-b`, `--txt-branches` and `--branches` are marked as deprecated and will be removed in one of the next versions.


latk:
I like this general direction. Thank you for doing this refactoring!

Not sure though about passing `options` directly to the sort method. Yes, fewer arguments, but this also hides which fields are needed by that function.

Spacetown:
> Not sure though about passing `options` directly to the sort method. Yes, fewer arguments, but this also hides which fields are needed by that function.

Yes they are hidden now but the user always needs to give the same arguments and can't change them since they are documented globally.

Another solution can be to change the sorting to be a literal option and deprecate `--sort-uncovered`, `--sort-percent`:
```python
            GcovrConfigOption(
                "sort",
                ["--sort"],
                config="sort",
                group="output_options",
                help=(
                    "The sort order for the output."
                ),
                choices=["filename", "uncovered-number", "uncovered-percent"],
                default="line",
            ),
```

The `--sort-reverse` I would keep.

Spacetown:
@latk Please can you check the last updates?

gcovr/configuration.py
```diff
@@ -490,36 +506,56 @@ def merge_options_and_set_defaults(
...
+        nargs=0,
+        action=UseSortUncoveredNumberAction,
     ),
     GcovrConfigOption(
-        "sort_percent",
+        "sort_key",
         ["-p", "--sort-percentage"],
         group="output_options",
         help=(
+            "Deprecated, please use '--sort-key uncovered-percent' instead."
```
Cirromulus, line 543:
```suggestion
            "Deprecated, please use '--sort-key uncovered-percent' instead. "
```

gcovr/configuration.py
```diff
@@ -490,36 +506,56 @@ def merge_options_and_set_defaults(
...
+    GcovrConfigOption(
+        "sort_key",
         ["-u", "--sort-uncovered"],
         group="output_options",
         help=(
+            "Deprecated, please use '--sort-key uncovered-number' instead. "
             "Sort entries by number of uncovered lines or branches (if the option "
             "--sort-branches is given). "
-            "The default order is increasing and can be changed by --sort-reverse. "
+            "The default order is increasing and can be changed by --sort-reverse."
```
Cirromulus, line 531:
```suggestion
            "The default order is increasing and can be changed by --sort-reverse. "
```

gcovr/configuration.py
```diff
@@ -490,36 +506,56 @@ def merge_options_and_set_defaults(
...
     GcovrConfigOption(
-        "sort_percent",
+        "sort_key",
         ["-p", "--sort-percentage"],
         group="output_options",
         help=(
+            "Deprecated, please use '--sort-key uncovered-percent' instead."
             "Sort entries by percentage of uncovered lines or branches (if the option "
             "--sort-branches is given). "
             "The default order is increasing and can be changed by --sort-reverse."
```
Cirromulus, line 546:
```suggestion
            "The default order is increasing and can be changed by --sort-reverse. "
```

Cirromulus (commented):
The change makes the code much better to read; I like this.
Only found minor things with missing spaces.

CHANGELOG.rst
```diff
@@ -27,17 +27,23 @@ New features and notable changes:
...
 - Add support for specifying files for :option:`search_paths`. (:issue:`834`)
 - Use different color for partial covered lines in HTML report. (:issue:`839`)
 - Add support to generate LCOV info files. (:issue:`830`)
 - Add support for FIPS enabled OS when used with Python 3.9. (:issue:`850`)
 - Reduce file size for detailed HTML reports by using CSS to link the function lists. (:issue:`840`)
 - Ignore all negative hits if :option:`--gcov-ignore-parse-errors` is used. (:issue:`852`)
+- Use literal options for sorting and TXT metric. (:issue:`867`)
+
+  - The :option:`-b`, :option:`--txt-branches` and :option:`--branches` are deprecated, use :option:`--txt-metric` instead.
+  - The :option:`--sort-uncovered` and :option:`--sort-percentage` are deprecated, use :option:`--sort` instead.
```
latk, line 46: I don't remember if we have a policy on what we mean by "deprecation". Perhaps it would be better to explicitly state our intention here.

* is it "will be removed in gcovr x.0.0"? I think this would cause unnecessary breakage, especially since these options are very old and will be part of some ancient scripts.
* or is it "hey FYI, we have something better now" and we are just *discouraging* use of the old options?

latk (approved):
great work, thank you!

gcovr/coverage.py
```diff
@@ -38,7 +38,18 @@
 from collections import OrderedDict
 import os
 import re
+import sys
 from typing import Any, List, Dict, Iterable, Optional, TypeVar, Union
+
+if sys.version_info >= (3, 8):
```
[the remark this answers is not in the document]
Spacetown: If we official drop the support for 3.7 we can remove this (and others). This I'll prefer to do in a separate PR.

gcovr/options.py
```diff
@@ -19,11 +19,12 @@
 
 from __future__ import annotations
 from argparse import ArgumentTypeError
+import argparse
 import logging
 from typing import Any, List, Optional, Union, Callable
 import os
 
-
+GET_TYPE = type
```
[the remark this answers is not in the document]
Spacetown: Removed.

gcovr/options.py
```diff
@@ -188,6 +189,26 @@ def get(self, name: str) -> Any:
...
+    def __call__(self, parser, namespace, values, option_string=None):
+        # This part is used when merging configurations
+        if isinstance(namespace, dict):
+            namespace[self.dest] = values
+        # We are called from argparse
+        else:
+            LOGGER.warning(
+                f"Deprecated option {option_string} used, please use '{self.option} {self.value}' instead."
+            )
+            setattr(namespace, self.dest, self.value)
```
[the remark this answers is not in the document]
Spacetown: I'll add a test for this.

CHANGELOG.rst
Spacetown: `Discouraged option which can be removed in newer versions´ is maybe better. I'm not sure when to drop the support for such option.
After this PR I want to extend our logger to support the pipeline tasks:
- [Azure](https://github.com/MicrosoftDocs/azure-devops-docs/blob/main/docs/pipelines/scripts/logging-commands.md#logissue-log-an-error-or-warning)
- [GitHub](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message)

If a pipeline run is detected we can modify the format to get the errors listed.

Spacetown: Deprecated is also used in the documentation and is handled in `bump_version.py`. In this case I'll stick to this. I think we should keep it for the next few versions/year.

gcovr/options.py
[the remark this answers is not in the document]
Spacetown: I need to extend the custom class to also be usable for configuration keys. Tests are also added.

PR #877, opened 2024-01-28
Add session to generate screenshots from html

Spacetown:
The used tool wkhtmltoimage was hard to get running on my Mac and I discovered that the tool can't handle CSS variables. The tool isn't developed any further and I searched for something else.
This PR adds a session to generate a JPEG from a HTML by using a server running a Chrome in a headless mode. The server needs to be started manually by the user if it's not started yet. The server wasn't able to provide PNG therefore I switched to JPEG.
I updated the docs and the examples that they fit together.


Spacetown:
@latk Please can you check if this dependency is OK? I want to get rid of the manual capturing of the HTML for the documentation. Maybe we can try to capture the screenshots during the build of the documentation, if there is nothing changed in the HTML the file is binary identical.

noxfile.py
```diff
@@ -379,6 +383,98 @@ def check_bundled_app(session: nox.Session) -> None:
...
+
+        content = re.sub(
+            r'<link rel="stylesheet" href="([^"]+)"/>',
+            lambda match: f'<style type="text/css">{read_file(os.path.join(os.path.dirname(html), match[1]))}</style>',
+            read_file(html),
+        )
+        data = {
+            "html": content,
+            "export": {
+                "type": "jpeg",
```
latk, line 458: The documentation claims that `"type": "png"` should work, which would be preferable for screenshots: https://github.com/bedrockio/export-html

noxfile.py
```diff
@@ -379,6 +383,98 @@ def check_bundled_app(session: nox.Session) -> None:
...
+    def exporter_running():
+        with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as sock:
+            if sock.connect_ex(("localhost", port)) == 0:
+                return True
+            return False
+
+    if not exporter_running():
+        session.error(
+            f"HTML exporter not running. Please run 'docker run  -p {port}:2305 bedrockio/export-html'"
+        )
```
latk, line 401: How quickly does the container boot up? If it's not too expensive, we could start the container automatically, figure out it's IP address, and then send requests its way. This would also avoid the need to hardcode a port number.

Something like:

```python
with ExitStack() as defer:
    container_id = subprocess.check_output(["docker", "run", "--rm", "--detach", "IMAGE"])
    defer.callback(subprocess.run, ["docker", "stop", container_id])
    container_config = json.loads(subprocess.check_output(["docker", "inspect", container_id]))
    container_ip = container_config[0]["NetworkSettings"]["IPAddress"]
    url = f"http://{container_ip}:2305/1/screenshot"

    ...
```

noxfile.py
```diff
@@ -379,6 +383,98 @@ def check_bundled_app(session: nox.Session) -> None:
...
+
+    for html, jpeg, size in [
+        (
+            "doc/examples/example_html.html",
+            "doc/images/screenshot-html.jpeg",
+            [800, 290],
+        ),
```
latk, line 408: Might be preferable to turn this loop into function invocations.

```python
for args in [(...), (...), ...]:
    ...
```

&darr;

```python
def screenshot(args):
   ...

screenshot(...)
screenshot(...)
...
```

noxfile.py
```diff
@@ -379,6 +383,98 @@ def check_bundled_app(session: nox.Session) -> None:
...
+            "-d",
+            json.dumps(data),
+            "-H",
+            "Content-Type: application/json",
+            "--output",
+            jpeg,
+            "-XPOST",
+            f"http://localhost:{port}/1/screenshot",
+            external=True,
+        )
```
latk, line 475: Can use the `requests` module from PyPI instead of shelling out to `curl`

latk:
> if this dependency is OK?

sure, wkhtml2pdf seems obsolete, and https://github.com/bedrockio/export-html looks OK (open source, easy to use via scripts)

> The server needs to be started manually by the user if it's not started yet.

Not the biggest fan of this because that's a manual step. We already automatically build and launch Docker containers for tests, so starting the Chrome container automatically could make sense as well.

> The server wasn't able to provide PNG therefore I switched to JPEG.

That's tolerable if there's no other way, but nice crisp PNGs would be nicer. The container's docs mention PNG export – what challenges did you encounter when using that feature?

> if there is nothing changed in the HTML the file is binary identical.

If that's the case that's great. Previously, the screenshots would include metadata like the creation date (I think), which made it impossible to check if the screenshots were up to date.

doc/examples/example.sh
```diff
@@ -13,7 +13,7 @@ g++ -fprofile-arcs -ftest-coverage -fPIC -O0 example.cpp -o program
 #END run
 
 #BEGIN gcovr
-gcovr
+gcovr -o example.txt
 #END gcovr
```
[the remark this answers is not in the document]
Spacetown: The  `-o` should of course be changed to `--html` but the file needs to be given or the documentation text needs to be updated because the output file name in mentioned:
https://github.com/gcovr/gcovr/blob/3e121a212d1219684b28165701af2a70112d0bbf/doc/source/output/html.rst?plain=1#L14-L16

doc/examples/example_csv.csv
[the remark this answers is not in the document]
Spacetown: The default line endings for the CSV writer is `"\r\n"`. If the file is stdout this seems to use `"\n"`.  Should we use always `"\n"` for writing the CSV or use .gitattributes to set the line endings?

noxfile.py
Spacetown: There is an [issue](https://github.com/bedrockio/export-html/issues/1) that the quality is always set but not supported for PNG.

noxfile.py
Spacetown: I thought also about this but that won't work inside the container without DIND. I'll try to modify the session to start the container for doing the conversion and stop it afterwards.

Spacetown: For me getting the IP address of the container is working but the container isn't accessible. I changed to get a free port from the OS and use this port to start the docker container.

PR #883, opened 2024-02-07
Add support for more than 9999 lines

### LINKED ISSUE #882: UnknownLineType thrown when parsing coverage data from 10K+ line file
**Describe the bug**
When the source code file is larger than 10K lines the .gcov format appears to drop the space between %%%%%: and the line number. This causes _RE_BLOCK_LINE to not match since it requires the space to be present.

Slightly redacted .gcov extract:
```
    %%%%%: 9998-block  1
call    6 never executed
    #####: 9999:    child_->set_access([...]);
call    0 never executed
branch  1 never executed (fallthrough)
branch  2 never executed (throw)
    #####:10000:    setReadAccess(*child_);
    %%%%%:10000-block  0
call    0 never executed
branch  1 never executed (fallthrough)
branch  2 never executed (throw)
    #####:10001:    child_ = child_->add_child([...]);
    %%%%%:10001-block  0
```

**To Reproduce**
1. Generate coverage data for a file with at least 10000 lines of code
2. gcovr -r . --html-details -o coverage/index.html (or similar)
3. gcovr dumps all -block lines with line number 10000 or higher, then fails with `gcovr.formats.gcov.parser.UnknownLineType:     %%%%%:10000-block  0`

**Expected behavior**
_RE_BLOCK_LINE should matches lines 10000 and later

 - OS: Debian testing
 - GCC version: Debian 13.2.0-13
 - GCOVR version 7.0

**Additional context**
When I deleted the .gcno file for the large .cpp file the report for the other files was generated successfully

Spacetown:
If more than 9999 code lines are present there is no space in front of the block number printed by gcov. Add support for optional spaces.

Closes #882


latk:
Instead of using an end-to-end test case with a file containing over 10k lines, it might be sufficient to just test this on the gcov parser level and add an example to the parser's doctests. Something like:

```pycon
>>> _parse_line("    %%%%%:12345-block  0")  # see https://github.com/gcovr/gcovr/issues/882
_BlockLine(hits=0, lineno=12345, blockno=0, extra_info=NONE)
```

As a general point, the gcovr test suite is fairly detailed but also extremely slow, and maintaining the "golden master" files is tedious (especially for XML and HTML formats). Wherever we can check something via a cheap unit test, we should prefer that.

One thing I've wanted to do (when I have the energy, so probably never) would be to review the test suite, remove tests for formats that aren't needed (so probably preferring JSON tests over HTML tests when testing reader functionality, preferring JSON inputs over running gcov when testing writer-side functionality), porting Makefiles to Python test modules to reduce re-building and process spawning overhead, and maybe even finding a way to use pytest-xdist for parallelization. Once there are fewer golden master files, it would also be easier to drop ancient GCC versions.

Spacetown:
> Instead of using an end-to-end test case with a file containing over 10k lines, it might be sufficient to just test this on the gcov parser level and add an example to the parser's doctests. Something like:
> 
> ```
> >>> _parse_line("    %%%%%:12345-block  0")  # see https://github.com/gcovr/gcovr/issues/882
> _BlockLine(hits=0, lineno=12345, blockno=0, extra_info=NONE)
> ```

I'll update this.

> As a general point, the gcovr test suite is fairly detailed but also extremely slow, and maintaining the "golden master" files is tedious (especially for XML and HTML formats). Wherever we can check something via a cheap unit test, we should prefer that.
> 
> One thing I've wanted to do (when I have the energy, so probably never) would be to review the test suite, remove tests for formats that aren't needed (so probably preferring JSON tests over HTML tests when testing reader functionality, preferring JSON inputs over running gcov when testing writer-side functionality), porting Makefiles to Python test modules to reduce re-building and process spawning overhead, and maybe even finding a way to use pytest-xdist for parallelization. Once there are fewer golden master files, it would also be easier to drop ancient GCC versions.

In this case we should also consider to remove the makefiles and use a python based test configuration. I was also thinking about this change.

PR #899, opened 2024-02-29
Create tag before publishing the distribution

Spacetown:
Because the tag pushed with the pipeline token is not triggering an additional workflow (see https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow) the tag is created and the package is published after the tag is pushed.


latk:
Sorry, I don't have the capacity to review this. But I trust that a better release process makes sense.

PR #918, opened 2024-04-19
Sort files on actual uncovered percentage

hoxell:
This fixes:
* Files with covered = 0 and total = 1 and 100% get equal sorting value (1), thus alphabetically interleaving files with 0% and 100% coverage.

* Files with covered = 0 and total > 1 end up after files with 100% coverage (or before, if reverse sorting is enabled), but before files that have 100% coverage because total = 0.

In terms of coloring, this change leads to color-consistent sorting (red-yellow-green, or green-yellow-red for reverse), but not the current interleaving of colors and coverage percentages.


Spacetown:
The intention of the logic was to have the uncovered branches sorted by the total number of branches instead of alphabetical order. So it's more a change than a fix. To you have a screenshot before/after your modification?

Spacetown:
Taking a look at the test results and thinking about it I don't get the reason for having the special case for uncovered. @latk can you recall this?

```
  --- reference/gcc-8/coverage.txt
  +++ coverage.txt
  @@ -4,11 +4,11 @@
  +file4.cpp                                      2       0     0%   1,3
   file3.cpp                                      9       4    44%   8,10-12,14
   file2.cpp                                      7       4    57%   8,10-11
   file1.cpp                                      4       3    75%   4
   main.cpp                                       5       5   100%
  -file4.cpp                                      2       0     0%   1,3
   TOTAL                                         27      16    59%
```

gcovr/coverage.py
```diff
@@ -90,6 +90,15 @@ def coverage_stat(key: str) -> CoverageStat:
...
 
+    def key_num_lines(key: str) -> int:
+        stat = coverage_stat(key)
+        # When sorting in ascending order, a heuristic not uncommonly
+        # employed when looking to improve coverage, is to browse
+        # the list from worst to best, so it makes sense to place
+        # files with many lines higher up in groups that otherwise share
+        # the same (primary) sorting key.
+        return -stat.total
+
```
Spacetown, line 101: I do not agree with this. If the numbers are the same the alphabetical order is better than sorting them by the total number of branches or lines.

gcovr/coverage.py
```diff
@@ -100,20 +109,23 @@ def key_percent_uncovered(key: str) -> float:
...
+        # By default, we sort by filename alphabetically
         return sorted(covdata, key=key_filename, reverse=sort_reverse)
 
-    # First sort filename alphabetical and then by the requested key
-    return sorted(sorted(covdata, key=key_filename), key=key_fn, reverse=sort_reverse)
+    return sorted(
+        covdata, key=lambda key: (
+            key_fn(key), key_num_lines(key), key_filename(key)
+        ), reverse=sort_reverse
+    )
```
Spacetown, line 128: Using a single sort also reverse the alphabetical order. If I'm sorting reverse by branch coverage is it required to sort also the filenames in reverse order? I don't think so.

hoxell:
Here's a stripped down example of the type of sorting the current logic
image

The various groups resulting from the branching in `key_percent_uncovered(...)` are sorted among themselves, but not with respect to each other, which leads to the interleaving behavior we're seeing.

It occurred to me that it would be convenient to sort the files based on the number of lines instead of the file name, but I didn't want to change the behavior.  Hearing that that's the intended behavior, though, I've updated this PR such that for custom sorting, the selected sorting key is used as the primary key, `total` is used as the secondary key, and the filename is used as the tertiary key. This leads the sorting being correct with respect to coverage and between the previously mentioned "groups", but I believe that it also makes the overall sorting more intuitive within the groups with the same primary sorting key value.

Here's an example:
image


Reverse sorting example:
image

gcovr/coverage.py
hoxell: I was on the fence about this one, thinking that reversing the sorting should mirror it, but I do agree with you on this one, it should only be applied to the primary key.

gcovr/coverage.py
hoxell: I think there are two camps here, or at least it depends on what your intentions with the report are.  
I'd find the sorting useful if I were looking for where adding tests would give most bang for the buck (especially those files with zero coverage) or understand which files contribute most to the overall coverage, but I can also see how sorting files based on their path/name is useful for other use cases.  
That said, I'll leave the final decision to you.

Spacetown: I would remove this level of sorting again. Imagine someone uses `--sort uncovered-number` and the second key is the total number of lines if the uncovered number of lines is the same, the file with more lines will come first but has a better coverage in %.

latk:
> I don't get the reason for having the special case for uncovered. @latk can you recall this?

Not sure either :)

It seems that in 0418545 the sign of a sorting key got flipped, unclear to me whether that was an accident or intentional.

I think at some point the intention of that code was to ensure that 0/1 branches and 0/123 branches are not both sorted as "0%", but effectively as -1% and -123% – more uncovered is worse. However, 0/0 branches should look more like 100%. Under no circumstances should there be NaNs or ZeroDivisionErrors.

I suspect that this inconsistency wasn't discovered previously because gcovr depends almost exclusively on end-to-end tests, which are high-fidelity but make it difficult to spot relevant changes in behaviour. Perhaps this PR could add unit tests to demonstrate the expected sort orders under different sort modes.

Spacetown:
I think the `-1` was removed because of the introduction of the `reversed`. Before this commit the sort order was from highest to lowest percentage by using this `-1`, see https://github.com/gcovr/gcovr/pull/817#issuecomment-1673719557:
```
                           GCC Code Coverage Report
Directory: .
main.cpp                                       6       6   100%   
file1.cpp                                      4       3    75%   4
file2.cpp                                      7       4    57%   8,10-11
file3.cpp                                      9       4    44%   8,10-12,14
file4.cpp                                      2       0     0%   1,3
TOTAL                                         28      17    60%
```
was changed to:
```
                           GCC Code Coverage Report
Directory: .
file3.cpp                                      9       4    44%   8,10-12,14
file2.cpp                                      7       4    57%   8,10-11
file1.cpp                                      4       3    75%   4
main.cpp                                       6       6   100%
file4.cpp                                      2       0     0%   1,3
TOTAL                                         28      17    60%
```

After thinking about it more closely, the specific case of covered doesn't make any sense to me. The same for the usage of the total number of branches if uncovered.

Spacetown:
@hoxell Can you update the reference data by running the docker containers local? I'll update the description for the sort options since the still reference the deprecated options and push this to your branch..

Spacetown (approved):
LGFM. Thanks for fixing this.

tests/html-nested-sort-percentage/reference/clang-10/coverage.A.7fc56270e7a70fa81a5935b72eacbe29.html
```diff
@@ -23,7 +23,7 @@ <h1>GCC Code Coverage Report</h1>
     </tr>
     <tr>
       <th scope="row">Date:</th>
-      <td>0000-00-00 00:00:00</td>
+      <td>00-00-00 00:00:00</td>
```
Pesa, line 26: is this change expected?

Spacetown: I also saw this when creating the single file report and fixed the error in the scrubber there. See https://github.com/gcovr/gcovr/pull/916/files#diff-14fc10b334e44e45733e71ed3f0d748422b8f8dedc6078c92f2ebca731072ccc.

Issue #913, opened 2024-04-05
Add support for GCC 14 condition coverage (MC/DC)


sebhub:
GCC 14 will support condition coverage:

https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fcondition-coverage

The corresponding GCC commit is:

https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=08a52331803f66a4aaeaedd278436ca8eac57b50

Support was added to the gcov tool to show the condition coverage information:

https://gcc.gnu.org/onlinedocs/gcc/Invoking-Gcov.html

```
gcov --conditions:

        3:   17:void fn (int a, int b, int c, int d) {
        3:   18:    if ((a && (b || c)) && d)
conditions covered 3/8
condition  0 not covered (true false)
condition  1 not covered (true)
condition  2 not covered (true)
condition  3 not covered (true)
        1:   19:        x = 1;
        -:   20:    else
        2:   21:        x = 2;
        3:   22:}

gcov --conditions --json-format:

"conditions": [
    {
        "not_covered_false": [
            0
        ],
        "count": 8,
        "covered": 3,
        "not_covered_true": [
            0,
            1,
            2,
            3
        ]
    }
],
```
It would be nice if gcovr could also present this information.

Spacetown:
We're waiting already for the release of version 14 to finish #766

Spacetown:
@latk How shall we continue with the decision coverage? With gcc 14 this wouldn't be needed anymore.

latk:
> With gcc 14 this wouldn't be needed anymore.

It will be quite some time until GCC 14 is widely used. Gcovr claims support for GCC 5.x (!), and Clang is also a thing.

It would be great to use this new GCC feature whenever available, but to continue supporting a reasonable range of other compilers.

whentojump:
Hi, is there any update on this? Thanks!

Spacetown:
There is still an open bug for the JSON. See #282.

Spacetown:
@sebhub Can we assume that condition 0 is branch 0 and 1, condition 1 is branch 2 and 3, ...?

jokva:
Unfortunately, in general, no. From the gcov manual:

[...]

PR #935, opened 2024-07-16
Remove function returned count

Spacetown:
The new JSON format from `gcov` doesn't provide the function return count, therefore this PR is removing it from the internal data model, the HTML report and the JSON format.


Spacetown:
This PR isn't needed anymore. I saw that the #913 introduced this information into the JSON:
```json
                   "calls": [
                        {
                            "source_block_id": 4,
                            "destination_block_id": 1,
                            "returned": 1
                        }
                    ]
```

Also the `block_ids` are now added to the JSON output.

PR #941, opened 2024-07-30
✨ Upload standalone applications as release artifacts

### LINKED ISSUE #940: ✨ Upload standalone applications as release artifacts
Since `6.0` `gcovr` has a nox session to create a standalone application, however, users still need to build it themselves. It would be nice to simply download the standalone application directly from the release page, which would enable users to easily add them to their own docker images.
The problem with building it locally in the CI is that `bundle_app` nox session also invokes `check_bundled_app` which requires `gcov` and `bash` to be installed in the same environment.
However in a typical docker build environment that isn't always the case since a c++ build image won't necessarily have `python` installed and a `python` build image won't necessarily have `gcov`.
And most Windows docker images won't have `bash` readily available.

Downloading the `gcovr` standalone applications directly from the release page.

Currently, we are building the `gcovr.exe` with a python docker image and just regex remove `session.notify("check_bundled_app")` at the end of the `bundle_app` nox session so it won't fail due to the missing bash executable.

**Additional context**
If the idea of uploading the standalone applications as release artifacts is fine with you I can make a PR to update the workflow.

s-weigand:
With this change the bundled standalone apps will be uploaded as artifacts to the release page.
The bundled apps will be uploaded as artifacts to the test workflow (default retention time 90 days) each CI run and only uploaded to the release page when the triggering event is a tag (release will be created for the tag if it didn't exist before).
Each app binary will be suffixed with the corresponding OS for easy identification (`linux`, `macos`, `win`).

As I understand the current release workflow the following happens:
Bump version on main -> deploy workflow creates tag -> deploy and test workflow run on the created tag
image
So the upload should happen naturally as a result of the current release process and create a release [similar to this test one](https://github.com/s-weigand/gcovr/releases/tag/release-artifact-upload-test). 

closes #940 

Additional notes:
- `gcovr-linux` won't work on musl-linux (i.e. apline) due to ABI incompatibility
- `gcovr-macos` should work on apple silicon but not tested


Spacetown (changes requested):
The MacOs executable from your branch can't be executed on a M1 Mac. Maybe because the option was not used in the noxfile.

 One major point: We do not trigger the deployment on tag push. We push the tag in the deployment pipeline and use the pipeline credentials which will not trigger other pipeline runs. Maybe we need to rethink this and use a separate token for pushing the tag.

.github/workflows/test.yml
```diff
@@ -118,23 +118,26 @@ jobs:
             python-version: '3.9'
           - os: ubuntu-22.04
             gcc: gcc-11
-            python-version: '3.11'
+            python-version: "3.12"
```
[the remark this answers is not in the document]
s-weigand: Just for consistency of the python version used to build the standalone apps across OS's. The changes where the extra matrix entries were added were quite some time apart.
image

noxfile.py
```diff
@@ -334,11 +335,15 @@ def bundle_app(session: nox.Session) -> None:
     # with the needed data
     session.install(".")
     os.makedirs("build", exist_ok=True)
+    additional_build_args = []
```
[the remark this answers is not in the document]
s-weigand: Ups my bad, I hope this fixes the app on M1 Mac.
Alternatively one could build a `x86` (`macos-12` image) and `ARM64` (`macos-14` image) version for mac. See [GHA-runner specs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories).

s-weigand:
> One major point: We do not trigger the deployment on tag push. We push the tag in the deployment pipeline and use the pipeline credentials which will not trigger other pipeline runs. Maybe we need to rethink this and use a separate token for pushing the tag.

Personally, I like this workflow:
- Create a "Prepare release x.y.z" PR
- Merge PR
- Create a tag and release from the release page

Except for the tag creation, this seems to mostly fall in line with your current release workflow.

Spacetown:
From the release page the tags are not annotated and our pipeline will fail.
I need t check how to register an app to get a token for pushing the tag which will then trigger the deployment.

noxfile.py
```diff
@@ -215,7 +218,7 @@ def doc(session: nox.Session) -> None:
     if not GCOVR_ISOLATED_TEST and not (
         # Github actions on MacOs can't use Docker
         platform.system() == "Darwin"
-        and "GITHUB_ACTION" in os.environ
+        and CI_RUN is True
```
Spacetown, line 221:
```suggestion
        and CI_RUN
```

noxfile.py
```diff
@@ -334,11 +337,15 @@ def bundle_app(session: nox.Session) -> None:
...
     session.install(".")
     os.makedirs("build", exist_ok=True)
+    additional_build_args = []
     with session.chdir("build"):
         if platform.system() == "Windows":
             executable = "gcovr.exe"
         else:
             executable = "gcovr"
+        # See: https://pyinstaller.org/en/stable/feature-notes.html#macos-multi-arch-support
+        if platform.system() == "Darwin" and CI_RUN is True:
```
Spacetown, line 347:
```suggestion
        if platform.system() == "Darwin" and CI_RUN:
```

noxfile.py
```diff
@@ -373,6 +381,20 @@ def check_bundled_app(session: nox.Session) -> None:
                 f"./gcovr --{format} $TMPDIR/out.{format}",
                 external=True,
             )
+    if CI_RUN is True:
```
Spacetown, line 384:
```suggestion
    if CI_RUN:
```

noxfile.py
```diff
@@ -537,7 +559,7 @@ def docker_build_compiler(session: nox.Session, version: str) -> None:
     set_environment(session, version)
     container_id = docker_container_id(session, version)
     cache_options = []
-    if "GITHUB_ACTIONS" in os.environ:
+    if CI_RUN is True:
```
Spacetown, line 562:
```suggestion
    if CI_RUN:
```

noxfile.py
```diff
@@ -276,7 +279,7 @@ def tests(session: nox.Session) -> None:
     if "--" not in args:
         args += ["--"] + DEFAULT_TEST_DIRECTORIES
 
-    running_locally = os.environ.get("GITHUB_ACTIONS") is None
+    running_locally = CI_RUN is False
```
Spacetown, line 282:
```suggestion
    running_locally = not CI_RUN
```

Spacetown (changes requested):
Instead of using the tag here we should use the [release event](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release):
```yaml
on:
  release:
    types:
      - released
```
Is the separate step for uploading all artefacts in one step needed or can we upload the artefacts directly when they are created?

noxfile.py
s-weigand: Even so, the truthy checks are shorter and maybe more pythonic, I learned to hate them over the years because of all the bugs I saw them cause (especially `if obj.method_i_forgot_to_call:` :sweat_smile: ), so I started to use an explicit style. But that is a taste issue, so I will fix it.

s-weigand:
The artifacts could also be uploaded directly to the release page after creation, without the intermediate push/PR artifacts. The advantage of the intermediate artifacts is that they can be tested on push and/or PR and an upload would only happen if all tests pass.
I don't think you need an app token, a [PAT](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#about-personal-access-tokens) should be sufficient to trigger consecutive workflows, since the PAT owner will be actor of the event. Didn't try it out yet but there is also a [beta for orga PATS](https://docs.github.com/en/organizations/managing-programmatic-access-to-your-organization/setting-a-personal-access-token-policy-for-your-organization).
For ease of use you could use [github-push-action](https://github.com/marketplace/actions/github-push), but manually rewriting the remote URL to include the PAT is ofc also an option.
My main reasoning of adding the upload to the test workflow was to avoid duplication of the GCC setup for Windows and Mac, since that would add additional maintenance burden to sync between a test and release workflow.

Also, note that [`softprops/action-gh-release`](https://github.com/softprops/action-gh-release) will create a release for a tag if no release already exists.

noxfile.py
Spacetown: I want to add pylint in the future. And I thought even flake8 gives an error with the explicit check.

Spacetown:
For now the release is created by hand as draft and published after the tag is created and pushed in the pipeline. The deployment is done in a separate pipeline and I think we need to merge the two pipelines to get the artefacts added to the release during this deployment. In that case we can also generate the release within the pipeline by extracting the information from the changelog.

In my opinion we should first get all the artefacts created in this PR and merging the pipelines for a fully automated release process should be done in a separate PR.

noxfile.py
s-weigand: I think what you mean is [E712](https://docs.astral.sh/ruff/rules/true-false-comparison/) where one should not use an `==` to compare against singletons like `True`, `False` or `None` but rather use a truthy check or `is` since `is` check for identity and not just equality (`__eq__`).
Personally, I followed the examples of `pydantic` and `FastAPI` to change all my linter+formatter tooling to [ruff](https://docs.astral.sh/ruff/) which replaced `flake8`, `yesqa`, `black`, and `isort` in my python projects. Compared to `pylint` it has the advantages that it runs A LOT faster and it has editor support to autofix some issues.
The other tooling I couldn't live without anymore is [`pre-commit`](https://pre-commit.com/) to manage+run all the code-QA tooling (added bonus is that you won't have things like `run black` or `fix linter issue` in you commit history :sweat_smile: ).

s-weigand:
I changed the test workflow to also run on the `release->released` event and the artifact upload to the release page to only happen on release (see [demo release](https://github.com/s-weigand/gcovr/releases/tag/only-upload-release-artifacts-on-release-test) and [corresponding CI run](https://github.com/s-weigand/gcovr/actions/runs/10311419701)).
I guess now it should fit nicely in the current release flow :smile: 
>  I think we need to merge the two pipelines to get the artefacts added to the release during this deployment

Another option would be to [call the deploy workflow from the test workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) after all tests have passed. That way the artifacts from the test workflow already exist and can be collected in the deploy workflow.

Spacetown:
Instead of calling one workflow from another I prefer a separate job which depends on the previous one. I already tried a little bit with the reusable job and step workflows.

Spacetown (approved):
LGFM

Spacetown:
@s-weigand Thanks again for this great improvement. Now alle things for the release are created by the pipeline and I can also remove the Milestones. 
Creating a new release by pushing a commit is incredible easy and fast.

PR #956, opened 2024-09-04
Don't loose stderr and stdout of gcovr for non-zero return code

NeitrinoK:
I try to fix next message
```
(ERROR) GCOV produced the following errors processing /home/gitlab-runner/disk/builds/HafKdfP1/0/PT/hiper100re/build/contrib/libre2/CMakeFiles/re2.dir/util/strutil.cc.gcda:
	Trouble processing '/home/gitlab-runner/disk/builds/HafKdfP1/0/PT/hiper100re/build/contrib/libre2/CMakeFiles/re2.dir/util/strutil.cc.gcda' with working directory '/home/gitlab-runner/disk/builds/HafKdfP1/0/PT/hiper100re'.
Stdout of gcov was >>None<< End of stdout    (<- Here)
Stderr of gcov was >>None<< End of stderr   (<- Here)
Exception was >>GCOV returncode was 5.<< End of stderr
```
We lose stdout and stderr when return code of gcov is not 0 or 6, so all useful information lost ( I used strace tool to get messages).

Now, stdout and stderr are saved in object for sure

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,6 +30,7 @@ Bug fixes and small improvements:
 - Fix bundle of standalone executable with Python 3.12. (:issue:`924`)
 - Fix merging of function coverage data. (:issue:`925`)
 - Fix inefficient regular expression. (:issue:`933`)
+- Fix missing output of gcov if execution fails. (:issue:`956`)
 
 Documentation:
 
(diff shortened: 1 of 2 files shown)


Spacetown:
Now we only see the last output. I think a better solution is to keep the local variables and add them to the exception text.

NeitrinoK:
> Now we only see the last output. I think a better solution is to keep the local variables and add them to the exception text. In the error output they should not be used if set to None.
NP, fixed.

Spacetown:
Can you print the output after the error message?

Spacetown (commented):
LGFM. Thanks for fixing this.

Spacetown (approved):
LGFM. Thanks for fixing this.

PR #961, opened 2024-09-10
Add comment to ignore branch source

### LINKED ISSUE #937: Branch coverage exclusion does not work
**Describe the bug**
Specific branches cannot be excluded from branch coverage calculation.

**To Reproduce**
`main.c`:

```c
enum myEnum {a};

int main() {
    enum myEnum foo = a;

    switch (foo) {
    case a:
        break;
    default:   /* GCOVR_EXCL_LINE */
        break; /* GCOVR_EXCL_LINE */
    }
    return 0;
}
```

Code compilation and report generation:

```bash
gcc --coverage -g -O0 main.c -o"main.exe"
main.exe
py -3.10 -m gcovr --html-details coverage.html
```

**Expected behavior**
Accordingly to <https://gcovr.com/en/7.2/guide/exclusion-markers.html> it should be possible to also exclude branches from coverage calculation.
The code above shall have 100% line **and** branch coverage, when `GCOVR_EXCL_BR` is placed in the `default:`-line.

However, what combinations of `GCOVR_EXCL_BR_*` I used around or in the `default` case I could not remove the `default` branch from coverage calculation.

**Screenshots**
line-coverage

 - OS: Windows 10
 - GCC version:
   - gcc x86_64-8.1.0-release-posix-seh-rt_v6-rev0
   - gcc.exe (Rev6, Built by MSYS2 project) 13.2.0
 - GCOVR version: 7.2
 - Project directory layout: not used

Spacetown:
Add exclusion marker `GCOVR_EXCL_BR_SOURCE` to exclude source branches which have one of the block ids of the line where the marker occurs as destination block.

```C
#include <iostream>

int foo(int param) {
   switch (param) {
   case 0:
      return 1;
   case 1:
      return 2;
   case 2:
      return 3; // GCOVR_EXCL_BR_SOURCE: Excludes the branches jumping to this line
   case 3:
      return 4;
   default:
      return 0; // GCOVR_EXCL_BR_SOURCE: Excludes the branch jumping to this line
   }
}

int main(int argc, char* argv[]) {
   if (foo(0) && foo(3)) {
      std::cout << "True" << std::endl;
   }
   else {
      std::cout << "False" << std::endl; // GCOVR_EXCL_BR_SOURCE: Excludes the branches jumping to this line
   }

   return 0;
}
```

Closes #937


Spacetown:
@latk Can you have a short look on this?

PR #976, opened 2024-09-30
Assume block 0 if block is missing

### LINKED ISSUE #973: gcovr with --lcov output does not seem compliant with lcov
# Describe the bug

When creating an lcov output with --lcov, it seems to not be LCOV compliant because `lcov --list coverage.lcov` returns failures.

# To Reproduce

1. Generate an lcov file with --lcov output. `gcovr --lcov coverage.lcov`
2. Try to use lcov to output the file in lcov format `lcov --list coverage.lcov`
3. Step 2 fails

# Expected behavior

LCOV should output fine and correctly parse the file output given by gcovr

- OS: [e.g. iOS] Mac
- GCC version [e.g. MinGW 8.1.0]:  13.2
- GCOVR version [e.g. 4.2]: 7.2

# Additional context

The issues seem to be the following:
unexpected file record 'VER:29b9cd73f3659f1b9236a52412291289'
unexpected  file record 'BRDA:1242,None,1,-'

When I run `sed -i '/VER:/d;s/None,/0,/g' coverage.lcov` it seems to work fine.

Spacetown:
For some branches the block number is missing in gcov test output file. Assume block 0 in this case to prevent `None` as block number in LCOV files.

Closes #973


gcovr/formats/gcov/parser.py
```diff
@@ -512,7 +512,9 @@ def _gather_coverage_from_line(
     elif isinstance(line, _FunctionLine):
         # Defer handling of the function tag until the next source line.
         # This is important to get correct line number information.
-        return state._replace(deferred_functions=[*state.deferred_functions, line])
+        return state._replace(
+            deferred_functions=[*state.deferred_functions, line]
+        )._replace(blockno=0)
```
latk, line 517: Can replace multiple fields in one call.

```suggestion
        return state._replace(
            deferred_functions=[*state.deferred_functions, line],
            blockno=0,
        )
```

However, it might be preferable to apply the block reset at the point where the deferred function is applied?

gcovr/formats/gcov/parser.py
```diff
@@ -453,7 +453,7 @@ def _reconstruct_source_code(tokens: Iterable[_Line]) -> List[str]:
 class _ParserState(NamedTuple):
     deferred_functions: List[_FunctionLine] = []
     lineno: int = 0
-    blockno: int = None
+    blockno: int = 0
```
latk, line 456: I'm not totally sure if we want to change this default. Are there scenarios where it might be helpful to show in our internal data format (and the JSON report) that we didn't get any block-level coverage data?

I absolutely agree that replacing `None` with `0` is important in the lcov-writer, but that could also be done without changes to our data model.

latk (approved):

gcovr/formats/gcov/parser.py
Spacetown: You're right here. Maybe it's better to keep the parser as it is and add a property `ensure_blockno` which returns `0` if it is `None`.
In that case we can also print a info message.

gcovr/formats/gcov/parser.py
Spacetown: I`ll undo the changes in the parser and add a property to the data model to get always a valid blockno.

PR #985, opened 2024-10-09
add conditions reporting in --txt-summary

dlegaultbbry:
Add support for reporting the conditions metrics in the text summary output if there are any.


dlegaultbbry:
I saw there is no gcc14 test for text summary so I could add one with the proper updates to the code if necessary to cover this addition. Anything else you feel would be required for testing?

Spacetown:
There are tests for the summaries but if there are no conditions they are removed. I suggest to add a summary to the decision test.

tests/decisions/Makefile
```diff
@@ -16,7 +16,7 @@ html: coverage.json
 	$(GCOVR) --verbose -a coverage.json --decisions --html-details -o coverage.html
 
 txt: coverage.json
-	$(GCOVR) --verbose -a coverage.json --txt-metric decision -o coverage.txt
+	$(GCOVR) --verbose -a coverage.json --txt-metric decision --txt-summary > coverage.txt
```
Spacetown, line 19: Instead of changing the current output a new one should be added..
```suggestion
	$(GCOVR) --verbose -a coverage.json --txt-metric decision -o coverage.txt
	$(GCOVR) --verbose -a coverage.json --txt-metric decision --txt-summary -o coverage_summary.txt
```

Spacetown (changes requested):
For the decision coverage we should compare both text outputs. The detailed and the summary one.

dlegaultbbry:
Sorry for not being quick enough, it's a holiday long week-end in Canada.

PR #998, opened 2024-10-14
Add empty file coverage

### LINKED ISSUE #986: Generating json for files which have no coverage
I'm not sure this would qualify as an idea or not, but here's what I'd like to do even if it means writing separate tool(s).

Assuming I can generate the list of source files which have no coverage (because no tests defined for them), how would I go to generate a gcovr output file using the current version of gcovr's json format such that later I can simple use `--json-add-tracefile` with 'empty' coverage stats combined with the actual json files which have real coverage from tests in them to get a combined report showing 0 for all the missing files.

Spacetown:
Add a option to scan the search paths for files which ar added to the report with zero coverage by default to see files which are without coverage/test for now.

Closes #986


dlegaultbbry:
Tested and it works fine.

Should the files report their total lines so that the total % would be lower than what it is for just tested files?

Spacetown:
> Should the files report their total lines so that the total % would be lower than what it is for just tested files?

I had this in the beginning but we don't know which lines contain code. If you wish this you can compile the files with coverage and we use then the gcno file if no gcda file is found.

dlegaultbbry:
I agree it's not a blocker. At least we can see the signal that a file is not covered which is sufficient in my use case. Thanks.

gcovr/configuration.py
```diff
@@ -611,6 +611,21 @@ class UseSortUncoveredPercentAction(GcovrDeprecatedConfigOptionAction):
...
     ),
+    GcovrConfigOption(
+        "include",
+        ["-i", "--include"],
+        group="filter_options",
+        help=(
+            "Include source files that match this filter. "
+            "This is to ensure that files are in report even "
+            "if no coverage data is found. Files are searched "
+            "recursive in root directory. "
```
dlegaultbbry, line 622: in -> from the

gcovr/formats/__init__.py
```diff
@@ -84,6 +93,34 @@ def read_reports(options) -> CovData:
...
+                exclude_dirs=options.gcov_exclude_dirs,
+            ):
+                # Return if the filename does not match the filter
+                # Return if the filename matches the exclude pattern
+                filtered, excluded = apply_filter_include_exclude(
+                    fname, options.filter, options.exclude
+                )
+
+                if filtered:
+                    LOGGER.debug(f"  Filtering coverage data for file {fname}")
```
dlegaultbbry, line 111: This is more of a debug question but it came up while I was trying to debug filters/excludes/etc. when using the tool. Could it be possible to know which pattern was matched when something was excluded/included/filtered? this would help when you mix and match several of these things together to know which ones had an effect.

gcovr/configuration.py
```diff
@@ -619,7 +619,7 @@ class UseSortUncoveredPercentAction(GcovrDeprecatedConfigOptionAction):
             "Include source files that match this filter. "
             "This is to ensure that files are in report even "
             "if no coverage data is found. Files are searched "
-            "recursive in root directory. "
+            "recursive from th --root directory. "
```
dlegaultbbry, line 622: missing an 'e' for the

gcovr/formats/__init__.py
Spacetown: That's not possible for now because the filter function uses the `any` function of Python. This needs to be refactored to a `for` loop.

Spacetown: I'll give it a try.

Spacetown: Can you check the logging? I added a debug message to the filter class itself and refactored the filtering function where for every call some logging was always done afterwards.

dlegaultbbry: Works great and very informative now!

PR #1009, opened 2024-10-22
Conditions mismatch

dlegaultbbry:
This adds on @Spacetown's changes here: https://github.com/Spacetown/gcovr/tree/improve_logging

This adds a new option called `--merge-mode-conditions=[strict|fold]` which allows to bypass aborts when gcov conditions do not line up across data files.

When 'fold' is given, a warning is printed and the merging data is shrunk to match the lowest count between left and right. This allows to bypass cases where gcov generates differing condition count sizes.

To fit the new merge options, I had to rework the option handling and split the options into function and condition options and handle both separately. Adjusted existing code for checking the proper values from sub data types.

Tested on the code which causes the issue currently seen:
- using strict2 = invalid option reported
- using strict = same abort as before with more information
- using fold = process goes through and proper coverage is generated. Checked both txt and HTML and they look the same as expected.

Missing:
- test case for this if we can makeup some data files to fake this condition as I don't have a fake code example to reproduce the issue

Testing:
- Some failures due to missing changes the logging commits before the one I'm adding
- test_html_template_dir: this one seems to be problematic as it only has gcc-5 and clang in the testing but it still runs on my machine which is ubuntu 22.04 and gcc11 so perhaps some filtering is missing to skip over


doc/source/guide/merging.rst
```diff
@@ -42,7 +42,17 @@ With the :option:`--merge-mode-functions` you can change this:
 - ``merge-use-line-max``: Allow same function on different lines, in this case use maximum line.
 - ``separate``: Allow same function on different lines. Instead of merging keep the functions separate.
 
+If you have differing number of conditions on the same line, the default behaviour is to abort.
+With the :option:`--merge-mode-conditions` you can change this:
+
+- ``strict``: Abort if condition count differs for the same line (old behavior).
```
Spacetown, line 48:
```suggestion
- ``strict``: Abort if condition count differs for the same line (default).
```

doc/source/guide/merging.rst
```diff
@@ -42,7 +42,17 @@ With the :option:`--merge-mode-functions` you can change this:
...
+- ``fold``: Narrow the conditions to match the lowest count between the two coverage counts.
+
 .. versionadded:: 6.0
 
    The :option:`gcovr --json-base` option.
    The :option:`gcovr --merge-mode-functions` option.
+
+.. versionadded:: 8.3
```
Spacetown, line 56: We replace the `NEXT` with the version if a new release is prepared.
```suggestion
.. versionadded:: NEXT
```

gcovr/configuration.py
```diff
@@ -655,6 +655,21 @@ class UseSortUncoveredPercentAction(GcovrDeprecatedConfigOptionAction):
             "Default is '{default!s}'."
         ),
     ),
+    GcovrConfigOption(
+        "merge_mode_conditions",
+        ["--merge-mode-conditions"],
+        metavar="COND_MERGE_MODE",
```
Spacetown, line 661: This value is only used in the help, we do not refer to it so we can use the same meta variable for all merge options.
```suggestion
        metavar="MERGE_MODE",
```

gcovr/formats/gcov/parser.py
```diff
@@ -345,6 +345,7 @@ def parse_coverage(
     *,
     filename: str,
     ignore_parse_errors: Set,
+    data_filename: str = None,  # Only for tests
```
Spacetown, line 347:
```suggestion
    data_filename: str = None,  # Default is only for tests
```

gcovr/formats/json/write.py
```diff
@@ -165,11 +165,19 @@ def _json_from_file(file: FileCoverage, options) -> dict:
     filename = presentable_filename(file.filename, options.root_filter)
     if options.json_base:
         filename = "/".join([options.json_base, filename])
-    return {
+    json_file = {
         "file": filename,
         "lines": _json_from_lines(file.lines),
         "functions": _json_from_functions(file.functions),
     }
+    # Only write data in verbose mode
```
Spacetown, line 173:
```suggestion
    # Only add the data sources in verbose mode
```

CHANGELOG.rst
```diff
@@ -15,6 +15,7 @@ New features and notable changes:
 - Add condition coverage to text summary report. (:issue:`985`)
 - Add :option:`--include` to search files in search paths which should be added to report. (:issue:`998`)
 - Add option to generate LCOV format produced by version 1.x of LCOV tool. (:issue:`1001`)
+- Add condition coverage merge mode option `--merge-mode-conditions` (:issue:`1005`)
```
Spacetown, line 18:
```suggestion
- Add condition coverage merge mode option :option:`--merge-mode-conditions` (:issue:`1005`)
```

gcovr/merging.py
```diff
@@ -468,7 +521,33 @@ def merge_condition(
...
+                ]
+                right.not_covered_false = right.not_covered_false[
+                    : len(left.not_covered_false)
+                ]
+                right.count = left.count
+        else:
+            raise AssertionError(
+                f"The number of conditions must be equal, got {left.count} and {right.count}.\n"
+                "\tYou can run gcovr with --merge-mode-conditions=COND_MERGE_MODE.\n"
+                "\tThe available values for COMD_MERGE_MODE are described in the documentation."
```
Spacetown, line 549: Now it's tricky because I added the line information at the end of the exception text. This would look like:
```
The number of conditions must be equal, got x and y.
\tYou can run gcovr with --merge-mode-conditions=COND_MERGE_MODE.
\tThe available values for COMD_MERGE_MODE are described in the documentation. while merging line z.
```
The only thing would be to split the line when extending the message and join it again.

Spacetown (changes requested):
Some small adaptions are needed. Do you have any suggestion how to solve the issue with the line information? Maybe we can add a parameter with the context information to the merge functions.

gcovr/merging.py
dlegaultbbry: Yes I see what you are saying. Will have to fiddle with things slightly.

dlegaultbbry:
Do you want to submit the logging improvements in a separate PR? Then once those merged in I can rebase on them and build these changes on top?

Spacetown:
> Do you want to submit the logging improvements in a separate PR? Then once those merged in I can rebase on them and build these changes on top?

I think this would be a better way. Will raise the PR in a few minutes.

dlegaultbbry:
Yes, adding the line information to the Condition coverage seems like a reasonable choice if you are ok with that. I'm fine adding this change in this set or you can add it to the logging change you will do separately.

Spacetown:
I'll add it to the logging improvement.

Spacetown:
@dlegaultbbry I've rebased your branch and you can continue.

tests/different-conditions-fold/Makefile
```diff
@@ -0,0 +1,38 @@
...
+
+run: json txt lcov cobertura html sonarqube coveralls
+
+coverage.json:
+	cd build1 ; ./testcase
+	cd build2 ; ./testcase
+	$(GCOVR) --verbose --merge-mode-conditions=fold --json-pretty --json coverage.json
+
+json: coverage.json
+	$(GCOVR) --verbose --json-add-tracefile $< > coverage.log 2>&1 ; test $$? -eq 0 | cat coverage.log
```
Spacetown, line 16: Would first check strict and test for the expected exit code and then check the redirected log file. 
```suggestion
	cd build1 ; ./testcase
	cd build2 ; ./testcase
	$(GCOVR) --verbose --merge-mode-conditions=fold --json-pretty --json coverage.json

json: coverage.json
```

tests/different-conditions-strict/Makefile
```diff
@@ -0,0 +1,20 @@
...
+	mkdir -p build1 build2
+	cd build1 ; cat ../main_hdr.c ../main1.c ../main_ftr.c > ../main.c ; $(CC) $(CXXFLAGS) ../main.c -o testcase
+	cd build2 ; cat ../main_hdr.c ../main2.c ../main_ftr.c > ../main.c ; $(CC) $(CXXFLAGS) ../main.c -o testcase
+
+run: json
+
+json:
+	cd build1 ; ./testcase
+	cd build2 ; ./testcase
+	$(GCOVR) --verbose --merge-mode-conditions=strict --json-pretty --json > coverage.log 2>&1 ; test $$? -eq 64 | cat coverage.log
```
Spacetown, line 13: Check the default.
```suggestion
	$(GCOVR) --verbose --json-pretty --json > coverage.log 2>&1 ; test $$? -eq 64 | cat coverage.log
```

tests/different-conditions-strict/Makefile
```diff
@@ -0,0 +1,20 @@
...
+	cat coverage.log
+	touch empty
```
Spacetown, line 15: For what is this needed?

Spacetown (changes requested):
Minor changes in test needed.

tests/different-conditions-fold/Makefile
dlegaultbbry: So combine the 2 tests into 1 test and check failing first (strict), then check that the merge is good (fold)?

tests/different-conditions-strict/Makefile
dlegaultbbry: I could not figure out how to run a test without any reference data intially.

I assume it runs the test against all compilers then and only checks against reference data if some is defined for the test?

Spacetown: Yes. But it searches the reference data also in older versions. If e.g. `coverage.json` isn't found for `gcc-14-Darwin` it is searched in `gcc-14`, `gcc-13-Darwin`, `gcc-13`, ...

tests/different-conditions-fold/Makefile
dlegaultbbry: I'm not sure I understand what you are suggesting here.

Spacetown: Sorry, the comment was old. I would not use `--merge-mode-conditions=fold` to check if the value `fold` is used as default.

Spacetown: Also the second gcovr call isn't needed.

dlegaultbbry: So delete all the other formats from the fold test except generating the coverage.json file? Sorry again, just trying to make sure I get what you are asking, but I'm not sure what you mean with `I would not use --merge-mode-conditions=fold to check if the value fold is used as default.`.

Also, what can I do to resolve the test issues which seem to be from environment differences? Like the mac test?

The gcc-14 one I'm not sure I get what differs and why there's variation in the data.

Spacetown: Have you used docker for testing? The GCC data can be generated for all compilers with the docker by running the command `nox -rs 'docker_run_compiler(all)' -- --s tests -- --update_reference` (I need to update the documentation here).
If there are differences for Windows or MacOs the differences are uploaded to the pipeline run. You can download them, put them to the root of your clone and run `not -rs import_reference -- *.zip`.

Spacetown:
Shall I update the reference for you?

dlegaultbbry:
Sure you can, it will be faster. I'll learn some more about everything for the enhancement I'll work on next. Thanks for the patience!

Spacetown (approved):
LGTM

dlegaultbbry:
Seems like the gcc-14 output from gcov can vary with subsequent executions. How do you deal with these normally in other tests?

Spacetown:
This is the first time we have this. Instead of Compling we can checkin the gcov files and use he existing once.

Spacetown:
Shall I merge it or do you want to change the test?

dlegaultbbry:
Really up to you. My fear is that the next PR someone works on will generate a failure on this test which is unrelated to their changes. If you want to check in the gcov files for this test, I'll let you do that so I can see how it is done for the next time it happens. You can also decide to leave it as is, merge and see if it will break again later and fix it then.

Spacetown:
@dlegaultbbry Please can you check my last 4 commits to solve the problem with the different order and the changed data of gcov?

dlegaultbbry:
LGTM!

Spacetown:
@dlegaultbbry I'm adding the merge for the branches and was rethinking the merge option `fold` and if we should use one option for condition and for branches since they are related.
I think it's better to use either left or right depending on the count (min/max) of branches/condition. If the count is different the data isn't related to each other and we shouldn't the values.

dlegaultbbry:
If you wish. My fear is that some day you'll have a different value for one of these items and you'll need to split them again. The potential reason is that a user may only want to do something with one or the other and if regrouped then they won't be able to. I can't say I can anticipate all the possible use cases that could fall out from all of these changes.

Spacetown:
I understand you concerns but conditions always have related branches and with GCC-14 we have additional informations for branches which need to be checked when merging. I need to check if I find a way to merge the branches depending on the source and target block ids if they are available because now I get an error if the branches with the same index has different information.

Spacetown:
I checked the JSON output of GCC-14 and found a block where the folding is wrong. When using lambda functions there are several entries for the same line and folding this will give a wrong result. E.g. The following code:
```C
#define LAMBDA_SORT \
   { \
      std::array<int, 10> arr = { 0, 9, 1, 8, 2, 7, 3, 6, 4, 5 }; \
      std::sort( \
         std::begin(arr), \
         std::end(arr), \
         [](int a, int b) { \
            if (a > b) \
               return true; \
            return false; \
         } \
      ); \
   }

// Function in line with exclude
void sort_excluded(void) /* GCOVR_EXCL_FUNCTION */ { LAMBDA_SORT /* THIS is not excluded*/
   LAMBDA_SORT
}
```

Gives following output (definition of sort_excluded is on line 44):
```JSON
{
    "line_number": 44,
    "function_name": "_Z13sort_excludedv",
    "count": 27,
    "unexecuted_block": false,
    "block_ids": [
        2,
        4,
        6
    ],
    "branches": [
        {
            "count": 9,
            "throw": false,
            "fallthrough": true,
            "source_block_id": 6,
            "destination_block_id": 7
        },
        {
            "count": 0,
            "throw": true,
            "fallthrough": false,
            "source_block_id": 6,
            "destination_block_id": 13
        }
    ],
    "calls": [
        {
            "source_block_id": 6,
            "destination_block_id": 1,
            "returned": 9
        }
    ],
    "conditions": []
},
{
    "line_number": 45,
    "function_name": "_Z13sort_excludedv",
    "count": 27,
    "unexecuted_block": false,
    "block_ids": [
        7,
        9,
        11
    ],
    "branches": [
        {
            "count": 9,
            "throw": false,
            "fallthrough": true,
            "source_block_id": 11,
            "destination_block_id": 12
        },
        {
            "count": 0,
            "throw": true,
            "fallthrough": false,
            "source_block_id": 11,
            "destination_block_id": 14
        }
    ],
    "calls": [
        {
            "source_block_id": 11,
            "destination_block_id": 1,
            "returned": 9
        }
    ],
    "conditions": []
},
{
    "line_number": 46,
    "function_name": "_Z13sort_excludedv",
    "count": 9,
    "unexecuted_block": false,
    "block_ids": [
        12
    ],
    "branches": [],
    "calls": [],
    "conditions": []
},
{
    "line_number": 44,
    "function_name": "_ZZ13sort_excludedvENKUliiE_clEii",
    "count": 369,
    "unexecuted_block": false,
    "block_ids": [
        2,
        3,
        4
    ],
    "branches": [
        {
            "count": 225,
            "throw": false,
            "fallthrough": true,
            "source_block_id": 2,
            "destination_block_id": 3
        },
        {
            "count": 144,
            "throw": false,
            "fallthrough": false,
            "source_block_id": 2,
            "destination_block_id": 4
        }
    ],
    "calls": [],
    "conditions": [
        {
            "count": 2,
            "covered": 2,
            "not_covered_true": [],
            "not_covered_false": []
        }
    ]
},
{
    "line_number": 45,
    "function_name": "_ZZ13sort_excludedvENKUliiE0_clEii",
    "count": 369,
    "unexecuted_block": false,
    "block_ids": [
        2,
        3,
        4
    ],
    "branches": [
        {
            "count": 225,
            "throw": false,
            "fallthrough": true,
            "source_block_id": 2,
            "destination_block_id": 3
        },
        {
            "count": 144,
            "throw": false,
            "fallthrough": false,
            "source_block_id": 2,
            "destination_block_id": 4
        }
    ],
    "calls": [],
    "conditions": [
        {
            "count": 2,
            "covered": 2,
            "not_covered_true": [],
            "not_covered_false": []
        }
    ]
},
```

With this information we should remove the merge mode and find another solution.

dlegaultbbry:
I'm getting the impression that this is getting closer to mission impossible! Seems worse with C++ than C too.

Spacetown:
I'll give my best but we need to extend our data model which is at the moment line based with dicts for the conditions and branches where the key is an incremented integer. Here I want to use a table with more information like function name.

Spacetown:
@latk The extending of the dict keys seem to work. How shall we count a line which is covered in one entry and not covered in the other one? Do we need partial covered lines and how shall this be aggregated in the stats?

Spacetown:
@dlegaultbbry Can you check the PR #1069?

PR #1010, opened 2024-10-22
Extend logging for data merge errors with info about the data sources

Spacetown:
If there is an error when merging the data it's hard to find the root cause. This PR extends the error message with the information about the data source where the error occurred.


Spacetown:
@dlegaultbbry Can you check this?

dlegaultbbry (approved):
Looks good to me.

Issue #988, opened 2024-10-10
Show metrics for excluded items in reports


dlegaultbbry:
I'd like to show a full report in one go, but have an additional number beside each metric show the stats on excluded items (branch, line, function, etc.).

Something along the lines of `gcovr --no-markers --marker-stats  ...`

This would turn something like
```
lines: 82.6% (161 out of 195)
functions: 100.0% (37 out of 37)
branches: 49.5% (52 out of 105)
conditions: 51.2% (44 out of 86)
```

Into

```
lines: 77.8% (161 out of 207): excl 5.7% (12/207)
functions: 100.0% (37 out of 37): excl 0% (0/37)
branches: 48.6% (52 out of 107): excl 1.8% (2/107)
conditions: 50.0% (44 out of 88): excl 2.2% (2/88)
```

Similar thought for all the other generated formats. 

Not sure if there is any combination of options which can do this today or not. I can work on it and submit a PR if there is nothing like this and it would be a nice addition.

Spacetown:
That sounds good for me but for this we need to extend the data model and change the exclusion handling. For excluded branches we remove them from the data model but for this we need to have an exclusion flag as we do it for lines.
I suggest to do the change in two steps:
1. Add the exclusion flag to the data model and adapt the writers respect it. --> Reference data must not be changed.
2. Add an option (e.g. `--exclusion-stats`) to show the excluded stats. This is not related to markers.

In your  example the percent values are confusing for me because the overall number of elements is inclusive the excluded elements but the percentage is evaluated without it.

dlegaultbbry:
Yes. I wasn't sure if the actual metrics included the lines that were excluded or not. Sounds good on the plan, will work on it.

dlegaultbbry:
I'll be working on this next. Will wait for all the cleanup and linting changes to go in.

dlegaultbbry:
Started working on last part of this feature.

dlegaultbbry:
Had a question regarding the behavior of the option for reporting.

If I understand the stats properly, right now the total for items (branches, etc.) is only for non excluded items.

This means that if I want to report excluded stats, the total now needs to include excluded items in the count.

So something like for branches:
real total = 12
non excluded total = 6
covered = 4

4/6 66.6% -> 4/12 33.3% (excl 6/12 50%)

1) Do you want to change the default reporting to always have total include all items including excluded? 
[...]

PR #1016, opened 2024-10-29
Rework exclusion handling to publish metrics

dlegaultbbry:
- Added excluded flags to some coverage metrics
- Changed code to set the flag instead of wiping things for branches, calls and conditions
- Adjusted reporting code to ignore now present but excluded items

Ran all docker tests, a few reference files were updated from the changes.

#988

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,6 +38,7 @@ Internal changes:
 - Replace ``black`` and ``flake8`` with ``ruff`` and move configuration of ``pytest`` to ``pyproject.toml``. (:issue:`1007`)
 - Add ``pylint`` for testing code. (:issue:`1014`)
 - Align variable names across the files. (:issue:`1015`)
+- Rework exclusion handling to keep information about excluded coverage. (:issue:`1016`)
 
 8.2 (13 October 2024)
 ---------------------
--- a/gcovr/coverage.py
+++ b/gcovr/coverage.py
@@ -188,9 +188,19 @@ class BranchCoverage:
         return self.blockno
 
     @property
+    def is_excluded(self) -> bool:
+        """Return True if the branch is excluded."""
+        return self.excluded
+
+    @property
+    def is_reportable(self) -> bool:
+        """Return True if the branch is reportable."""
+        return not self.excluded
+
+    @property
     def is_covered(self) -> bool:
         """Return True if the branch is covered."""
-        return self.count > 0
+        return self.is_reportable and self.count > 0
 
 
 class CallCoverage:
@@ -201,22 +211,31 @@ class CallCoverage:
             The number of the call.
         covered (bool):
             Whether the call was performed.
+        excluded (bool, optional):
+            Whether the call is excluded.
     """
 
-    __slots__ = "callno", "covered"
+    __slots__ = "callno", "covered", "excluded"
 
     def __init__(
         self,
         callno: int,
         covered: bool,
+        excluded: Optional[bool] = False,
     ) -> None:
         self.callno = callno
         self.covered = covered
+        self.excluded = excluded
+
+    @property
+    def is_reportable(self) -> bool:
+        """Return True if the call is reportable."""
+        return not self.excluded
 
     @property
     def is_covered(self) -> bool:
         """Return True if the call is covered."""
-        return self.covered
+        return self.is_reportable and self.covered
 
 
 class ConditionCoverage:
@@ -228,12 +247,14 @@ class ConditionCoverage:
         covered (int):
             Whether the call was performed.
         not_covered_true List[int]:
-            The conditions which where not true.
+            The conditions which were not true.
         not_covered_false List[int]:
-            The conditions which where not false
+            The conditions which were not false.
+        excluded (bool, optional):
+            Whether the condition is excluded.
     """
 
-    __slots__ = "count", "covered", "not_covered_true", "not_covered_false"
+    __slots__ = "count", "covered", "not_covered_true", "not_covered_false", "excluded"
 
     def __init__(
         self,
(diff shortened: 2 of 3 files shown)


gcovr/coverage.py
```diff
@@ -201,18 +201,21 @@ class CallCoverage:
...
     """
 
-    __slots__ = "callno", "covered"
+    __slots__ = "callno", "covered", "excluded"
 
     def __init__(
         self,
         callno: int,
         covered: bool,
+        excluded: Optional[bool] = None,
```
Spacetown, line 214: Use always a default value `False` (same as for `LineCoverage`).

gcovr/coverage.py
```diff
@@ -230,17 +233,20 @@ class ConditionCoverage:
...
-    __slots__ = "count", "covered", "not_covered_true", "not_covered_false"
+    __slots__ = "count", "covered", "not_covered_true", "not_covered_false", "excluded"
 
     def __init__(
         self,
         count: int,
         covered: int,
         not_covered_true: List[int],
         not_covered_false: List[int],
+        excluded: Optional[bool] = None,
```
Spacetown, line 249: Also default of `False`.

gcovr/coverage.py
Spacetown, line 1: I would use the same functions as for the line coverage to the other coverage types:
```python
    @property
    def is_excluded(self) -> bool:
        """Return True if the line is excluded."""
        return self.excluded

    @property
    def is_reportable(self) -> bool:
        """Return True if the line is reportable."""
        return not self.excluded

    @property
    def is_covered(self) -> bool:
        """Return True if the line is covered."""
        return self.is_reportable and self.xxx

    @property
    def is_uncovered(self) -> bool:
        """Return True if the line is uncovered."""
        return self.is_reportable and not self.xxx
```

tests/excl-branch/reference/gcc-5/coverage.main.cpp.118fcbaaba162ba17933c7893247df3a.html
```diff
@@ -116,9 +116,16 @@ <h1>GCC Code Coverage Report</h1>
...
+          <summary class="linebranchSummary">1/0</summary>
+          <div class="linebranchContents">
+            <div class="excludedBranch">&ndash; Branch 0 excluded.</div>
+            <div class="excludedBranch">&ndash; Branch 1 excluded.</div>
+          </div>
+        </details>
       </td>
-      <td class="linecount coveredLine">1</td>
-      <td class="src coveredLine"><span class="w">   </span><span class="k">if</span><span class="w"> </span><span class="p">(</span><span class="n">param</span><span class="p">)</span><span class="w"> </span><span class="p">{</span><span class="w"> </span><span class="c1">// GCOVR_EXCL_BR_LINE</span></td>
+      <td class="linecount partialCoveredLine">1</td>
```
Spacetown, line 127: Why is this changed? I think this change shouldn't affect the reference files.

CHANGELOG.rst
```diff
@@ -36,6 +36,7 @@ Internal changes:
 - Fix sporadic timestamp mismatch in development build package. (:issue:`1006`)
 - Replace ``black`` and ``flake8`` with ``ruff`` and move configuration of ``pytest`` to ``pyproject.toml``. (:issue:`1007`)
 - Add ``pylint`` for testing code. (:issue:`1014`)
+- Rework exclusion handling to allow publishing metrics (:issue:`1016`)
```
Spacetown, line 39:
```suggestion
- Rework exclusion handling to keep information about excluded coverage. (:issue:`1016`)
```

Spacetown (changes requested):
I think this PR should not affect any reference data because only the information is preserved and should be ignored for the report.

gcovr/exclusions/__init__.py
```diff
@@ -141,9 +141,7 @@ def remove_internal_functions(filecov: FileCoverage):
                     linecov.function_name is not None
                     and linecov.function_name == function.name
                 ):
-                    linecov.excluded = True
-                    linecov.branches = {}
-                    linecov.count = 0
+                    linecov.exclude()
```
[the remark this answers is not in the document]
Spacetown: Than I suggest to exclude only branches and do the other things with the next PR.

dlegaultbbry:
I'd propose we discuss the behavior before approving and doing the rest.

If I exclude a line, does that mean all branches, calls and conditions under it should report 0 as if never executed? This is what was happening earlier with the first patch set. Right now they are wiped from existence so the metrics never report them because they are deleted before the coverage output is produced. 

There is a slight inconsistency with calls and conditions are they are not emptied out like branches when the line is excluded. This happens in `remove_internal_functions` and `apply_exclusion_ranges`. Maybe it would be better to fix this inconsistency before making other changes to reduce the complexity of reviewing test diff changes.

Spacetown:
I think it's better to fix the inconsistency first by also clearing the data and then continue on this PR which then shouldn't affect the test data.

Spacetown (commented):
LGFM except the change in the noxfile.

Spacetown (approved):

noxfile.py
```diff
@@ -404,6 +404,7 @@ def tests(session: nox.Session) -> None:
     set_environment(session)
     session.log("Print tool versions")
     session.run("python", "--version")
+    session.run("gcovr", "--version")
```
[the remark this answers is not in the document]
dlegaultbbry: Wanted to see the gcovr version used printed when running tests. I found it useful, but I can revert the change.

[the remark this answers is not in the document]
Spacetown: GCOVR is installed editable in the virtual environment, no need to print the version.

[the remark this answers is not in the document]
Spacetown:
```suggestion
```

PR #1018, opened 2024-10-31
Fix line exclusion

dlegaultbbry:
- When lines are excluded, they should remove coverage for branches, calls and conditions.
- This ensures consistency of reporting for future enhancements to exclusion metrics

#988

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,7 @@ Bug fixes and small improvements:
 - Check format version of external generated ``gcov`` JSON files. (:issue:`999`)
 - Fix crash on Windows when trying to fix the case of the files. (:issue:`1000`)
 - Fix LCOV report. Excluded lines where added with a count of 0. (:issue:`1012`)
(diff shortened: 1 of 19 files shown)


Spacetown (approved):
LGFM

dlegaultbbry:
One thing I noted while downloading the data files for mac and windows.

The Windows 2019, gcc, 3.8 and 3.12 files are the same name. Should they?

Spacetown:
They can. In that case you only need to import the 3.8.

dlegaultbbry:
Seems like there is no test hitting that exclusion condition. I'll see about creating one.

Spacetown:
> Seems like there is no test hitting that exclusion condition. I'll see about creating one.

This code is only executed with GCC 14. I'm not sure if such functions will have assigned lines.

```python
def _function_can_be_excluded(name: str) -> bool:
    """special names for construction/destruction of static objects will be ignored"""
```

The exclusion was added with #991 but I was not able to test it, I'll bypass the gate here.

PR #903, opened 2024-03-20
Detect suspicious counters

Spacetown:
A suspicious counter is a counter value of 2^32 or greater.

@JohanBertrand Please can you check this branch with your code?

Closes #898

PR #1021, opened 2024-11-02
Add option for suspicious hits

Spacetown:
https://github.com/gcovr/gcovr/pull/903#issuecomment-2444895806 pointed out that the threshold of 2**32 needs to be increased. This PR adds an option to increase it or to turn off the detection at all.


Spacetown:
@tvoskui Please can you check the PR if it's working for you?

tvoskui:
Yep, tested it on my large test set with no flag (parse error from the suspicious hits check) and with the threshold set to both 0 and 2**48 (both work with no errors or warnings). Thanks!

PR #1023, opened 2024-11-06
Add coverage container

Spacetown:
Replace the dict type with a coverage container class. The sorting routine is now a member function of the coverage class.

The coverage statistics are available as property of the coverage container and the file coverage class instead of a static function of the summary statistics.

Remove the reference to the parent directory from the file coverage and create the directory coverage as list of the coverage container.


Spacetown:
@macetw You've added the nested report with the summarised directory coverage. Are you willing to review this PR here where I've added container classes for the coverage.

dlegaultbbry:
Had a look at it, didn't notice anything obviously wrong.

PR #1030, opened 2024-11-25
Drop support of python 3.8

Spacetown:
Drop support and replace type annotations tor set, list, tuples and dict.


src/gcovr/exclusions/__init__.py
```diff
@@ -59,7 +59,7 @@ class ExclusionOptions:
     """
 
     respect_exclusion_markers: bool = True
-    exclude_functions: List["re.Pattern[str]"] = field(default_factory=lambda: [])
+    exclude_functions: list["re.Pattern[str]"] = field(default_factory=lambda: [])
```
latk, line 62: quotes around the type expression can be removed now, I think

src/gcovr/utils.py
```diff
@@ -31,16 +31,14 @@
 
 LOGGER = logging.getLogger("gcovr")
 
-MD5_KWARGS = {"usedforsecurity": False} if sys.version_info >= (3, 9) else {}
-
```
latk, line 35: :+1: excellent for spotting + removing this now-unnecessary check!

admin/Dockerfile.qa
```diff
@@ -29,12 +29,37 @@ apt-get install -y sudo \
...
+  # cspell:enable
+  cd /tmp
+  PythonVersion=3.9.20
+  wget https://www.python.org/ftp/python/$PythonVersion/Python-$PythonVersion.tgz
+  tar xzf Python-$PythonVersion.tgz
+  cd Python-$PythonVersion
+  ./configure --enable-optimizations
+  # cspell:ignore altinstall
+  make altinstall
+  update-alternatives --install /usr/local/bin/python3 python3 /usr/local/bin/python3.9 10
```
latk, line 56: This part is absolutely awful but I'm glad you found a way.

I have a minor concern that installing the `build-essential` package implies a GCC version that might differ from the one we request with `$CC`/`$CXX`, but that shouldn't be able to interfere.

An alternative to building Python ourselves could be download an ancient version from some binary archive. For example, the `uv` package manager (alternative to `pip`) downloads pre-built Pythons from the https://github.com/indygreg/python-build-standalone project. But doing that requires extra platform detection (e.g. aarch64 vs x86_64) and a step to unpack the archives.

latk (approved):

src/gcovr/utils.py
Spacetown: I always search for `version_info` if dropping support for a specific version.

admin/Dockerfile.qa
Spacetown: I removed the build essentials and it still works., using our gcc version. It would be hard to build the filename for a prebuilt version and building it on our own doesn't take much time. The build for the whole container takes about 160s.
I played with the layering to first build the common parts and then install GCC and Python but the time benefit is worse against the effort for maintaining such a Dockerfile.

Spacetown: We can improve this in a separate PR and maybe build the Python in a separate stage `FROM ... AS PythonBuild` and then install it in the final container.

PR #1050, opened 2025-01-08
Update link now sources have been moved to src/ folder

gilprime:
Point to real folder now the sources files have been moved to `src/`.

[no changelog]


Spacetown (commented):
LGFM. Thanks for fixing this broken link.

Spacetown (approved):

PR #1059, opened 2025-01-27
Fix function names in json report

### LINKED ISSUE #1058: AssertionError: Function name must be equal.
# Describe the bug

In some cases, generated json data from gcov traces will have:
- name: None
- demangled_name: <true function name>

When merging the 2 data sets with --json-add-tracefile I get an exception on https://github.com/gcovr/gcovr/blob/main/src/gcovr/merging.py#L388

If I remove this condition and let it run everything seems fine and the proper data is generated.

# To Reproduce

None so far, trying to figure out why the name is empty. I can see there are 2 cases where the data is added with a None name:

https://github.com/gcovr/gcovr/blob/main/src/gcovr/formats/gcov/parser/text.py#L352
https://github.com/gcovr/gcovr/blob/main/src/gcovr/formats/gcov/parser/text.py#L429

# Expected behavior

Things should merge since the function names do align.

# Screenshots

- OS: Linux
- GCC version: 14
- GCOVR version: 8.3

# Additional context

None

Spacetown:
This PR adds support to merge tracefiles from different compiler versions, one with mangled and the other with demangled names.
It also fixes the issue that always `demangled_name` was set even if the compiler doesn't support demangled names in text output (e.g. clang-10).  

Closes #1058.


doc/source/guide/merging.rst
```diff
@@ -42,6 +42,14 @@ With the :option:`--merge-mode-functions` you can change this:
 - ``merge-use-line-max``: Allow same function on different lines, in this case use maximum line.
 - ``separate``: Allow same function on different lines. Instead of merging keep the functions separate.
 
+If you have the mangled or demangled name defined only in some files (mainly if you user cross compiling
+with ``gcc-14`` and older ``gcc`` versions) the mangled names can be defined only in one input file.
+With the :option:`--merge-mode-function-names` you can change this:
+
+- ``strict``: Abort if same (de)mangled name is only defined in one file or is different (old behavior).
+- ``ignore-single-definition``: Allow that function name is only defined in one file.
```
dlegaultbbry, line 50: Allow the case of a function name (demangled or not) only being present in one file.

src/gcovr/merging.py
```diff
@@ -382,15 +411,40 @@ def merge_function(
...
-        raise AssertionError("Function demangled name must be equal.")
-    if left.name != right.name:
-        raise AssertionError("Function name must be equal.")
+    if left.demangled_name is not None and right.demangled_name is not None:
+        if (
+            not options.func_name_opts.ignore_function_name
+            and left.demangled_name != right.demangled_name
+        ):
+            raise AssertionError(
+                f"Demangled name must be equal, got {left.demangled_name} and {right.demangled_name}."
```
dlegaultbbry, line 420: Demangled function names ...

Should these exceptions also point to the relevant option to use to fix things?

src/gcovr/merging.py
```diff
@@ -382,15 +411,40 @@ def merge_function(
...
+            )
+    elif left.demangled_name is not None or right.demangled_name is not None:
```
dlegaultbbry, line 422: I think this logic doesn't factor in the option to ignore the name completely

```
if (left.dm is None or right.dm is None) and (not options.single and not options.ignore_name):
   raise AssertionError(
       f"Demangled function name {left.demangled_name or right.demangled_name} only defined in one file."
   )
```

src/gcovr/merging.py
```diff
@@ -382,15 +411,40 @@ def merge_function(
...
+            left.demangled_name = right.demangled_name
+        if not options.func_name_opts.ignore_function_name_single_definition:
+            raise AssertionError(
+                f"Demangled function name {left.demangled_name} only defined in one file."
+            )
+
+    if left.name is not None and right.name is not None:
+        if not options.func_name_opts.ignore_function_name and left.name != right.name:
+            raise AssertionError(
+                f"Mangled name must be equal, got {left.name} and {right.name}."
```
dlegaultbbry, line 433: Do we want to say Mangled here and not Just "Function names must be equal"

src/gcovr/merging.py
```diff
@@ -382,15 +411,40 @@ def merge_function(
...
+            )
+    elif left.name is not None or right.name is not None:
```
dlegaultbbry, line 435: Same as my other comment on this logic block in the demangled section. Since you are repeating the exact same code for both, perhaps it would make sense to refactor in a helper function?

src/gcovr/merging.py
```diff
@@ -382,15 +411,40 @@ def merge_function(
...
+        if right.name is not None:
+            left.name = right.name
+        if not options.func_name_opts.ignore_function_name_single_definition:
+            raise AssertionError(
```
dlegaultbbry, line 439: Should all these exceptions also be GcovrMergeAssertionError if I see the use below?

.github/workflows/gcovr-ci-job.yml
```diff
@@ -119,10 +124,9 @@ jobs:
         run: |
           nox --non-interactive $NOX_CONTAINER_ARGUMENTS --session tests --  --archive_differences
           # There must not be any modifications to get the correct versions later
-          Status=$(git status --porcelain)
-          if [ -n "$Status" ] ; then
+          if [ -n "$(git status --porcelain)" ] ; then
             echo "Following files are modified:"
-            echo "$Status"
+            git diff
```
dlegaultbbry, line 129: `git diff --name-status` 

if you only want to see which files were changed without the full diff in each one.

.github/workflows/gcovr-ci-job.yml
```diff
@@ -146,10 +150,8 @@ jobs:
           nox --non-interactive $NOX_CONTAINER_ARGUMENTS --session doc || exit 1
 
           # Check if files are modified (outdated screenshots)
-          Status=$(git status --porcelain)
-          if [ -n "$Status" ] ; then
-            echo "Following files are modified:"
-            echo "$Status"
+          if [ -n "$(git status --porcelain)" ] ; then
+            git diff
```
dlegaultbbry, line 154: same here

.github/workflows/gcovr-ci-job.yml
Spacetown: I want to see the differences here because on local system there was no change. The diff pointed out that the version 0.0+main was used in CI because the tags are not fetched anymore (see https://github.com/actions/checkout/issues/2041).

Spacetown:
> I'm also unsure about ignoring names entirely and if that is desirable. I assume you stumbled on a use case requiring this.

You're right here, since the function is the key of the dict used for merging this should not be the case and I remove this switch.

Spacetown:
I've got the same issue as described in https://github.com/linux-test-project/lcov/issues/79#issuecomment-1347146853 when using the demangled names as keys for the function with GCC-14. Virtual destructors have an internal function which results in the same demangled name. I don't want to treat them as different functions but how shall they be detected?
@dlegaultbbry , @latk Can someone suggest how to solve this?

dlegaultbbry:
> I've got the same issue as described in [linux-test-project/lcov#79 (comment)](https://github.com/linux-test-project/lcov/issues/79#issuecomment-1347146853) when using the demangled names as keys for the function with GCC-14. Virtual destructors have an internal function which results in the same demangled name. I don't want to treat them as different functions but how shall they be detected? @dlegaultbbry , @latk Can someone suggest how to solve this?

From the lcov thread it seems like they are saying that code should fold the functions/stats as if it were all the same. So basically, if they all end in `.*[C|D]<num>Ev`, then you could trap this and replace them all into `.*[C|D]0Ev` for mangled name handling. I don't know if you want to make this behaviour default or behind a command line option.

Spacetown:
Without support of demangled names I'll not fold it because this are two different functions which won't be sorted. 
With support of demangled names they are folded without doing anything. 
With GCC-14 they are also folded but the mangled name is different and the exception is raised. In that case we can suppress the check if the mangled name is ending on this pattern.

latk:
I agree that it *might* make sense to parse enough of the GCC name mangling scheme to collapse all destructors into a single coverage data record. This would be a special case for destructors and maybe constructors, not for other functions.

This is a destructive operation, but it seems to make the problem easier to solve, and aligns with what users tend to expect. This could make gcovr less useful for people who parse the JSON output, but that tradeoff might be worth it.

It seems that GCC currently uses the Itanium C++ ABI for name mangling, which has [5 different ctor and 3 different dtor flavors](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling-special-ctor-dtor). So it is likely that the regex `D[012]Ev$` will cover all relevant cases.

CHANGELOG.rst
```diff
@@ -10,8 +10,12 @@ Known bugs:
 
 Breaking changes:
 
+- Function properties ``name`` and ``demangled_name`` in JSON are changed. Both properties are only available for ``gcc-14``. (:issue:`1059`)
```
dlegaultbbry, line 13: Should this say, `... gcc-14 or newer.`?

CHANGELOG.rst
```diff
@@ -10,8 +10,12 @@ Known bugs:
...
+
 New features and notable changes:
 
+- Add :option:`--merge-mode-function-names` to support merging of JSON trace files from ``gcc-14`` with older versions. (:issue:`1059`)
```
dlegaultbbry, line 17: Same here `... gcc-14 or newer with older versions.`

doc/source/guide/merging.rst
```diff
@@ -42,6 +42,13 @@ With the :option:`--merge-mode-functions` you can change this:
 - ``merge-use-line-max``: Allow same function on different lines, in this case use maximum line.
 - ``separate``: Allow same function on different lines. Instead of merging keep the functions separate.
 
+If you have the mangled or demangled name defined only in some files (mainly if you user cross compiling
```
dlegaultbbry, line 45: If you have the mangled or demangled name only defined in some files (if you mix reports generated using ``gcc-14`` and older ``gcc`` versions) the mangled names could only be present in one input file but not the other. When combining the reports, the default behavior is to detect the situation where the names do no match.

doc/source/guide/merging.rst
```diff
@@ -42,6 +42,13 @@ With the :option:`--merge-mode-functions` you can change this:
...
+with ``gcc-14`` and older ``gcc`` versions) the mangled names can be defined only in one input file.
+With the :option:`--merge-mode-function-names` you can change this:
```
dlegaultbbry, line 47: `... change this behavior:

doc/source/guide/merging.rst
```diff
@@ -42,6 +42,13 @@ With the :option:`--merge-mode-functions` you can change this:
...
+
+- ``strict``: Abort if same (de)mangled name is only defined in one file or is different (old behavior).
```
dlegaultbbry, line 49: Abort if a (de)mangled name is only defined in one file or they differ (old behavior).

doc/source/guide/merging.rst
```diff
@@ -42,6 +42,13 @@ With the :option:`--merge-mode-functions` you can change this:
...
+- ``ignore-single-definition``: Allow the case of a function name (demangled or not) only being present in one file.
```
dlegaultbbry, line 50: Allow the case of a function (de)mangled name being only present in one file.

src/gcovr/configuration.py
```diff
@@ -669,6 +669,21 @@ class UseSortUncoveredPercentAction(GcovrDeprecatedConfigOptionAction):
...
+        ["--merge-mode-function-names"],
+        metavar="MERGE_MODE",
+        group="gcov_options",
+        choices=(
+            "strict",
+            "ignore-single-definition",
+        ),
+        default="strict",
+        help=(
+            "The merge mode for function names from different gcov files for same sourcefile. "
```
dlegaultbbry, line 683: `... for the same ...`

dlegaultbbry (commented):
Rest looks good to me.

CHANGELOG.rst
Spacetown: There is no newer version released yet.

dlegaultbbry: I was kind of trying to make the comment forward looking as we know newer releases will have this support.

Spacetown: I checked the previous changing entries and we used "... since gcc-xx".

doc/source/guide/merging.rst
Spacetown: > `... change this behavior:

With this button you can add a suggestion:
image

```suggestion
With the :option:`--merge-mode-function-names` you can change this behaviour:
```

dlegaultbbry (approved):

Spacetown:
I wanted to increase coverage and found out that due to multi threading the output can change for each execution when merging files with virtual destructors generated from legacy text format and new JSON format of gcov. Need to dig into this and find a solution.

Spacetown:
The merging is horrible and I'm thinking bout to not support merging data from different GCC versions.
After merging the line coverage for the virtual destructors I now have problems with lambda functions:
```
RuntimeError: GCOV produced the following errors processing /gcovr/tests/excl-function/testcase-main.gcda:
        Function name in /gcovr/tests/excl-function/main.cpp:44 must be _Z13sort_excludedv, got _ZZ13sort_excludedvENKUliiE_clEii.
        (gcovr could not infer a working directory that resolved it.)
```

I think we first need to refactor the coverage classes and fix the merging problems from same GCC versions which I found during implementing this PR. E.g. For a line we have one function name in our data model but with GCC-14 there can be more functions in one line.
After the refactoring we can try again to add support for merging data from different gcc versions.

dlegaultbbry:
I am totally fine if you want to drop this entirely. This is an unusual use case that most people would probably not have in their systems. I realize that C++ coverage seems to be much more painful to handle.

Spacetown:
This approach is not followed anymore. The new approach is to keep the data separate by extending the dict key from a line number to a list of line information.

#1069 
#1092

PR #1075, opened 2025-03-12
Add filename and line number in warning log for negative hits

### LINKED ISSUE #1071: Add file name and line in warning log for ignoring negative hits

Currently the warning for encountering negative hits during parsing doesn't have the information of what file and what line is the code located. It would be better for debugging if there are the file name and the line number for large projects.

It's just as:
```sh
(WARNING) Ignoring negative hits in line '    std::cout << "hello world" << std::endl;'.
```

```sh
(WARNING) Ignoring negative hits in sample.cpp line 97: '    std::cout << "hello world" << std::endl;'.
```

# Additional context

kenchung285:
Closes #1071


src/gcovr/formats/gcov/parser/common.py
```diff
@@ -77,7 +78,7 @@ def raise_if_not_ignored(
             if "negative_hits.warn_once_per_file" in persistent_states:
                 persistent_states["negative_hits.warn_once_per_file"] += 1
             else:
-                LOGGER.warning(f"Ignoring negative hits in line {line!r}.")
+                LOGGER.warning(f"Ignoring negative hits in {os.path.basename(filename)}, line {lineno}: {line!r}.")
```
kenchung285, line 81: important change

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -135,7 +136,7 @@ def raise_if_not_ignored(
             if "suspicious_hits.warn_once_per_file" in persistent_states:
                 persistent_states["suspicious_hits.warn_once_per_file"] += 1
             else:
-                LOGGER.warning(f"Ignoring suspicious hits in line {line!r}.")
+                LOGGER.warning(f"Ignoring suspicious hits in {os.path.basename(filename)}, line {lineno}: {line!r}.")
```
kenchung285, line 139: important change

kenchung285:
Hi @Spacetown , I have open this PR for the feature we have discussed in #1071   
I found that not only warning logs for negative hits but also logs for suspicious hits having the same issue.  
I fix them together.

Could you please review this PR? Because of my first-time contribution to `gcovr`, if there's any problem, please tell me!  
Appreciate your review and comments!

Spacetown:
Thanks for working on this but I want to wait for #1069 to be merged because the whole configuration tracking is reworked there.

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -148,23 +149,25 @@ def check_hits(
...
+    lineno: int,
 ) -> int:
     """
     Check if hits count is negative or suspicious, if the issue is ignored returns 0
-    >>> check_hits(1, "", {}, 10, {})
+    >>> check_hits(1, "", {}, 10, {}, "", 5)
     1
-    >>> check_hits(-1, "", {"all"}, 10, {})
+    >>> check_hits(-1, "", {"all"}, 10, {}, "", 5)
     0
```
Spacetown, line 160: Can you add a test also checking the exception?

src/gcovr/formats/gcov/parser/text.py
```diff
@@ -527,6 +528,7 @@ def _report_lines_with_errors(
 
 def _parse_line(
     line: str,
+    filename: str = "",
```
Spacetown, line 531: Putting the filename as first parameter seems to be more logical for me.

src/gcovr/formats/gcov/parser/text.py
```diff
@@ -759,6 +765,7 @@ def _parse_line(
 
 def _parse_tag_line(  # pylint: disable=too-many-return-statements
     line: str,
+    filename: str,
```
Spacetown, line 768: Also changing order of arguments.

Spacetown (changes requested):
I suggest to use as first argument the filename and then the line content.
You need to update the AUTHORS and CHANGELOG.

kenchung285:
Hi, @Spacetown, I'm fixing the errors occurs in pytest and doctest.  
But I have no idea of how to fix errors like the following in `test_gcovr.py::test_build` in my local development environment?  
```sh
________________________________________________ test_build[simple1-stdout-cobertura] _________________________________________________

compiled = 'simple1-stdout', current_format = 'cobertura'
available_targets = {'all': set(), 'clean': set(), 'cobertura': set(), 'coveralls': set(), ...}, generate_reference = False
update_reference = False, archive_differences = False

    ......

>       assert diff_is_empty, "Diff output:\n" + "".join(whole_diff_output)
E       AssertionError: Diff output:
E         -- reference/clang-13/cobertura.xml
E         ++ cobertura.xml
E           <coverage branch-rate="0.5" branches-covered="2" branches-valid="4" complexity="0.0" line-rate="..." lines-covered="8" lines-valid="11" timestamp="..."
E         -   version="gcovr 8.3+main"
E         +   version="gcovr 0.0+main"
E           >
E           ...
E           </coverage>
E         
E       assert False

/Users/ken/Documents/Side-Projects/gcovr/tests/test_gcovr.py:613: AssertionError
-------------------------------------------------------- Captured stdout call ---------------------------------------------------------
STDOUT - START ['make', '-j', '4', 'cobertura']
./testcase
/Users/ken/Documents/Side-Projects/gcovr/.nox/tests/bin/python -m gcovr -d -x > cobertura.xml
(INFO) Reading coverage data...
(INFO) Writing coverage report...
STDOUT - END
```

```sh
FAILED tests/test_gcovr.py::test_build[add_coverages-html] - AssertionError: Diff output:
FAILED tests/test_gcovr.py::test_build[add_coverages-cobertura] - AssertionError: Diff output:
FAILED tests/test_gcovr.py::test_build[bad++char-html] - AssertionError: Diff output:
FAILED tests/test_gcovr.py::test_build[bad++char-cobertura] - AssertionError: Diff output:
FAILED tests/test_gcovr.py::test_build[calls-html] - AssertionError: Diff output:
FAILED tests/test_gcovr.py::test_build[cmake_oos-html] - AssertionError: Diff output:
FAILED tests/test_gcovr.py::test_build[cmake_oos-cobertura] - AssertionError: Diff output:
FAILED tests/test_gcovr.py::test_build[cmake_oos_ninja-html] - AssertionError: Diff output:
FAILED tests/test_gcovr.py::test_build[cmake_oos_ninja-cobertura] - AssertionError: Diff output:
FAILED tests/test_gcovr.py::test_build[config-output-html] - AssertionError: Diff output:
FAILED tests/test_gcovr.py::test_build[config-toml-html] - AssertionError: Diff output:
......
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -148,23 +161,44 @@ def check_hits(
...
+    >>> check_hits(1000, "code", {}, 10, {}, "", 5)
+    Traceback (most recent call last):
+        ...
+    gcovr.formats.gcov.parser.common.SuspiciousHits: Got suspicious hit value in gcov line 'code' caused by a
+    bug in gcov tool, see
+    https://gcc.gnu.org/bugzilla/show_bug.cgi?id=68080. Use option
+    --gcov-ignore-parse-errors with a value of suspicious_hits.warn,
+    or suspicious_hits.warn_once_per_file or change the threshold
+    for the detection with option --gcov-suspicious-hits-threshold.
+    >>> check_hits(-1, "code", {"all"}, 10, {}, "", 5)
```
kenchung285, line 188: Add tests for checking the exception when encountering negative hits or suspicious hits

src/gcovr/formats/gcov/parser/text.py
```diff
@@ -552,118 +554,118 @@ def _parse_line(
...
+    >>> _parse_line(line=' #####:10000:foo += 1;')  # see https://github.com/gcovr/gcovr/issues/882
     _SourceLine(hits=0, lineno=10000, source_code='foo += 1;', extra_info=NONE)
-    >>> _parse_line(' =====: 13:foo += 1;')
+    >>> _parse_line(line=' =====: 13:foo += 1;')
     _SourceLine(hits=0, lineno=13, source_code='foo += 1;', extra_info=EXCEPTION_ONLY)
-    >>> _parse_line('   12*: 13:cond ? f() : g();')
+    >>> _parse_line(line='   12*: 13:cond ? f() : g();')
     _SourceLine(hits=12, lineno=13, source_code='cond ? f() : g();', extra_info=PARTIAL)
-    >>> _parse_line(' 1.7k*: 13:foo();')
+    >>> _parse_line(line=' 1.7k*: 13:foo();')
```
kenchung285, line 569: Because the argument `filename` has been moved to the first argument, we should specify the argument name `line` in these tests.

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -41,7 +42,7 @@ def __init__(self, line: str) -> None:
 
     @staticmethod
     def raise_if_not_ignored(
-        line: str, ignore_parse_errors: set[str], persistent_states: dict[str, Any]
+        line: str, ignore_parse_errors: set[str], persistent_states: dict[str, Any], filename: str, lineno: int
```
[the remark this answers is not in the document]
Spacetown: Yes, that would be more logical for me.

Spacetown:
> Hi, @Spacetown, I'm fixing the errors occurs in pytest and doctest. But I have no idea of how to fix errors like the following in `test_gcovr.py::test_build` in my local development environment? These errors are the only errors I haven't fixed.

You need to pull at least the last tag to get the correct version from `setuptools_scm`.

src/gcovr/formats/gcov/parser/text.py
Spacetown: You can use `"file", 1` in the test.

src/gcovr/formats/gcov/parser/common.py
Spacetown: Why do you add comments on your own code line?

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -42,27 +42,31 @@ def __init__(self, line: str) -> None:
 
     @staticmethod
     def raise_if_not_ignored(
-        line: str, ignore_parse_errors: set[str], persistent_states: dict[str, Any], filename: str, lineno: int
+        line: str,
+        ignore_parse_errors: set[str],
+        persistent_states: dict[str, Any],
+        filename: str,
+        lineno: str,
```
Spacetown, line 49: Should be moved at the begin of the arguments.

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -78,7 +82,9 @@ def raise_if_not_ignored(
             if "negative_hits.warn_once_per_file" in persistent_states:
                 persistent_states["negative_hits.warn_once_per_file"] += 1
             else:
-                LOGGER.warning(f"Ignoring negative hits in {os.path.basename(filename)}, line {lineno}: {line!r}.")
+                LOGGER.warning(
+                    f"Ignoring negative hits in {os.path.basename(filename)}, line {lineno}: {line!r}."
```
Spacetown, line 86: Why using base name and not the filename itself?

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -136,7 +146,9 @@ def raise_if_not_ignored(
             if "suspicious_hits.warn_once_per_file" in persistent_states:
                 persistent_states["suspicious_hits.warn_once_per_file"] += 1
             else:
-                LOGGER.warning(f"Ignoring suspicious hits in {os.path.basename(filename)}, line {lineno}: {line!r}.")
+                LOGGER.warning(
+                    f"Ignoring suspicious hits in {os.path.basename(filename)}, line {lineno}: {line!r}."
```
Spacetown, line 150: Why using base name and not the filename itself?

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -150,19 +162,38 @@ def check_hits(
     suspicious_hits_threshold: int,
     persistent_states: dict[str, Any],
     filename: str,
-    lineno: int,
+    lineno: str,
```
Spacetown, line 165: Why is linen a string?

src/gcovr/formats/gcov/parser/text.py
```diff
@@ -237,7 +237,7 @@ def parse_metadata(
         if not line:
             continue
 
-        parsed_line = _parse_line(line, suspicious_hits_threshold)
+        parsed_line = _parse_line("", line, suspicious_hits_threshold)
```
Spacetown, line 240: Would add a filename argument to the function and give the name of the gcov file here.

src/gcovr/formats/gcov/parser/text.py
```diff
@@ -543,7 +544,8 @@ def _report_lines_with_errors(
 
 
 def _parse_line(
-    line: str,
+    filename: str = "",
+    line: str = "",
```
Spacetown, line 548: Remove the default values here.

src/gcovr/formats/gcov/parser/text.py
```diff
@@ -840,6 +853,8 @@ def _parse_tag_line(  # pylint: disable=too-many-return-statements
                 ignore_parse_errors,
                 suspicious_hits_threshold,
                 persistent_states,
+                filename,
+                lineno="",
```
Spacetown, line 857: `lineno` is an integer which should be kept in the persistent states.

Spacetown (changes requested):
It's unclear for me why adding the defaults and the named parameter in the doctests.

src/gcovr/formats/gcov/parser/common.py
kenchung285: I originally want to tell the reviewers like you where's the important modification or new update after your first review. I hope this can help you review my code.

src/gcovr/formats/gcov/parser/common.py
kenchung285: I test the warning log manually in my own project. I found that the `filename` object actually includes the path to the file, like `/home/ken/project/include/sub-module-name/example.cpp`.  
This makes the warning log been too long and too complicated to read. Therefore, I recommend to use just base name like `example.cpp`, which is good enough for finding out where is the code line that triggers the warning.

src/gcovr/formats/gcov/parser/common.py
kenchung285: Same reason as https://github.com/gcovr/gcovr/pull/1075#discussion_r1999983497

src/gcovr/formats/gcov/parser/common.py
kenchung285: I found this line of code and thought the project store `lineno` as a `str`. 
https://github.com/gcovr/gcovr/blob/b8c8995c4e500373de15642c6b54cfe38cf3cca1/src/gcovr/formats/gcov/parser/text.py#L700
I'll update to use `int` instead

src/gcovr/formats/gcov/parser/common.py
Spacetown: I always check all changes and can select the commits till your last change. The comments are more confusing.

src/gcovr/formats/gcov/parser/common.py
Spacetown: I wouldn't do so. The same name can also be present in different directories. For errors the log shouldn't be shortened.

Spacetown: I wouldn't do so. The same name can also be present in different directories. For errors the log shouldn't be shortened.

src/gcovr/formats/gcov/parser/common.py
Spacetown: The referenced location is from matching the raw line. After the matching it should be converted into an integer.

src/gcovr/formats/gcov/parser/text.py
Spacetown: I think you should add a tuple `("filename", "linen")` to the persistent states and update it whenever a new line is read. This can then be used if a error occurs.

Spacetown: Here you can update the persistent state:
https://github.com/gcovr/gcovr/blob/820b1810d3f4c12aabfb8a9c6ebd11b449def13d/src/gcovr/formats/gcov/parser/text.py#L706

kenchung285: Did you mean to add a line after the reference line like
```python
    hits_str, lineno, source_code = match.groups()
    persistent_states[filename] = lineno
```
But this occurs at line 706 while we are discuss on `_parse_tag_line` which is called at line 681

tests/test_gcov_parser.py
```diff
@@ -760,7 +760,10 @@ def test_negative_branch_count_ignored_json(
     for index in range(0, number_of_warnings):
         message = messages[index]
         assert message[1] == logging.WARNING
-        assert message[2].startswith("Ignoring negative hits in line ")
+        if index == 0:
+            assert message[2].startswith("Ignoring negative hits in <stdin>, line 2: ")
+        else:
+            assert message[2].startswith("Ignoring negative hits in <stdin>, line 4: ")
```
Spacetown, line 766:
```suggestion
        assert message[2].startswith(f"Ignoring negative hits in <stdin>, line {2 if index == 0 else 4}: ")
```

tests/test_gcov_parser.py
```diff
@@ -811,7 +814,14 @@ def test_negative_line_count_ignored(
...
         assert message[1] == logging.WARNING
-        assert message[2].startswith("Ignoring negative hits in line ")
+        if index == 0:
+            assert message[2].startswith(
+                "Ignoring negative hits in example.cpp, line 2: "
+            )
+        else:
+            assert message[2].startswith(
+                "Ignoring negative hits in example.cpp, line 4: "
+            )
```
Spacetown, line 824:
```suggestion
        assert message[2].startswith(
            f"Ignoring negative hits in example.cpp, line {2 if index == 0 else 4}: "
        )
```

tests/test_gcov_parser.py
```diff
@@ -925,7 +935,14 @@ def test_suspicious_line_count_ignored(
...
         assert message[1] == logging.WARNING
-        assert message[2].startswith("Ignoring suspicious hits in line ")
+        if index == 0:
+            assert message[2].startswith(
+                "Ignoring suspicious hits in example.cpp, line 2: "
+            )
+        else:
+            assert message[2].startswith(
+                "Ignoring suspicious hits in example.cpp, line 4: "
+            )
```
Spacetown, line 945:
```suggestion
        assert message[2].startswith(
            f"Ignoring suspicious hits in example.cpp, line {2 if index == 0 else 4}: "
        )
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -151,20 +163,42 @@ def check_hits(
...
+    for the detection with option --gcov-suspicious-hits-threshold.
+    >>> check_hits(-1, "code", {"all"}, 10, {"reading_line": ("file", 5)})
     0
-    >>> check_hits(1000, "", {"all"}, 10, {})
+    >>> check_hits(1000, "code", {"all"}, 10, {"reading_line": ("file", 5)})
     0
     """
+    filename = persistent_states["reading_line"][0]
+    lineno = persistent_states["reading_line"][1]
+
```
Spacetown, line 192:
```suggestion
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -151,20 +163,42 @@ def check_hits(
...
     if hits < 0:
-        NegativeHits.raise_if_not_ignored(line, ignore_parse_errors, persistent_states)
+        NegativeHits.raise_if_not_ignored(
+            filename, lineno, line, ignore_parse_errors, persistent_states
+        )
```
Spacetown, line 196:
```suggestion
        NegativeHits.raise_if_not_ignored(line, ignore_parse_errors, persistent_states)
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -151,20 +163,42 @@ def check_hits(
...
         hits = 0
 
     if suspicious_hits_threshold != 0 and hits >= suspicious_hits_threshold:
         SuspiciousHits.raise_if_not_ignored(
-            line, ignore_parse_errors, persistent_states
+            filename, lineno, line, ignore_parse_errors, persistent_states
```
Spacetown, line 201:
```suggestion
            line, ignore_parse_errors, persistent_states
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -41,27 +41,31 @@ def __init__(self, line: str) -> None:
 
     @staticmethod
     def raise_if_not_ignored(
-        line: str, ignore_parse_errors: set[str], persistent_states: dict[str, Any]
+        filename: str,
+        lineno: int,
+        line: str,
+        ignore_parse_errors: set[str],
+        persistent_states: dict[str, Any],
```
Spacetown, line 48:
```suggestion
        line: str, ignore_parse_errors: set[str], persistent_states: dict[str, Any],
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -41,27 +41,31 @@ def __init__(self, line: str) -> None:
...
     ) -> None:
         """
         Raise exception if not ignored by options
         >>> state = dict()
-        >>> NegativeHits.raise_if_not_ignored("code", None, state)
+        >>> NegativeHits.raise_if_not_ignored("file", 1, "code", None, state)
```
Spacetown, line 53:
```suggestion
        >>> NegativeHits.raise_if_not_ignored("code", None, state)
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -41,27 +41,31 @@ def __init__(self, line: str) -> None:
...
         Traceback (most recent call last):
             ...
         gcovr.formats.gcov.parser.common.NegativeHits: Got negative hit value in gcov line 'code' caused by a
         bug in gcov tool, see
         https://gcc.gnu.org/bugzilla/show_bug.cgi?id=68080. Use option
         --gcov-ignore-parse-errors with a value of negative_hits.warn,
         or negative_hits.warn_once_per_file.
-        >>> NegativeHits.raise_if_not_ignored("code", {"all"}, state)
+        >>> NegativeHits.raise_if_not_ignored("file", 1, "code", {"all"}, state)
```
Spacetown, line 61:
```suggestion
        >>> NegativeHits.raise_if_not_ignored("code", {"all"}, state)
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -41,27 +41,31 @@ def __init__(self, line: str) -> None:
...
         >>> state.get("negative_hits.warn_once_per_file")
-        >>> NegativeHits.raise_if_not_ignored("code", {"negative_hits.warn"}, state)
+        >>> NegativeHits.raise_if_not_ignored("file", 1, "code", {"negative_hits.warn"}, state)
         >>> state.get("negative_hits.warn_once_per_file")
-        >>> NegativeHits.raise_if_not_ignored("code", {"negative_hits.warn_once_per_file"}, state)
+        >>> NegativeHits.raise_if_not_ignored("file", 1, "code", {"negative_hits.warn_once_per_file"}, state)
         >>> state.get("negative_hits.warn_once_per_file")
         1
-        >>> NegativeHits.raise_if_not_ignored("code", {"negative_hits.warn_once_per_file"}, state)
+        >>> NegativeHits.raise_if_not_ignored("file", 1, "code", {"negative_hits.warn_once_per_file"}, state)
```
Spacetown, line 68:
```suggestion
        >>> NegativeHits.raise_if_not_ignored("code", {"negative_hits.warn_once_per_file"}, state)
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -41,27 +41,31 @@ def __init__(self, line: str) -> None:
...
         --gcov-ignore-parse-errors with a value of negative_hits.warn,
         or negative_hits.warn_once_per_file.
-        >>> NegativeHits.raise_if_not_ignored("code", {"all"}, state)
+        >>> NegativeHits.raise_if_not_ignored("file", 1, "code", {"all"}, state)
         >>> state.get("negative_hits.warn_once_per_file")
-        >>> NegativeHits.raise_if_not_ignored("code", {"negative_hits.warn"}, state)
+        >>> NegativeHits.raise_if_not_ignored("file", 1, "code", {"negative_hits.warn"}, state)
         >>> state.get("negative_hits.warn_once_per_file")
-        >>> NegativeHits.raise_if_not_ignored("code", {"negative_hits.warn_once_per_file"}, state)
+        >>> NegativeHits.raise_if_not_ignored("file", 1, "code", {"negative_hits.warn_once_per_file"}, state)
```
Spacetown, line 65:
```suggestion
        >>> NegativeHits.raise_if_not_ignored("code", {"negative_hits.warn_once_per_file"}, state)
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -41,27 +41,31 @@ def __init__(self, line: str) -> None:
...
         gcovr.formats.gcov.parser.common.NegativeHits: Got negative hit value in gcov line 'code' caused by a
         bug in gcov tool, see
         https://gcc.gnu.org/bugzilla/show_bug.cgi?id=68080. Use option
         --gcov-ignore-parse-errors with a value of negative_hits.warn,
         or negative_hits.warn_once_per_file.
-        >>> NegativeHits.raise_if_not_ignored("code", {"all"}, state)
+        >>> NegativeHits.raise_if_not_ignored("file", 1, "code", {"all"}, state)
         >>> state.get("negative_hits.warn_once_per_file")
-        >>> NegativeHits.raise_if_not_ignored("code", {"negative_hits.warn"}, state)
+        >>> NegativeHits.raise_if_not_ignored("file", 1, "code", {"negative_hits.warn"}, state)
```
Spacetown, line 63:
```suggestion
        >>> NegativeHits.raise_if_not_ignored("code", {"negative_hits.warn"}, state)
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -77,7 +81,9 @@ def raise_if_not_ignored(
             if "negative_hits.warn_once_per_file" in persistent_states:
                 persistent_states["negative_hits.warn_once_per_file"] += 1
             else:
-                LOGGER.warning(f"Ignoring negative hits in line {line!r}.")
+                LOGGER.warning(
+                    f"Ignoring negative hits in {filename}, line {lineno}: {line!r}."
```
Spacetown, line 85:
```suggestion
                    f"Ignoring negative hits in {':'.join(persistent_states['location'])}: {line!r}."
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -99,12 +105,16 @@ def __init__(self, line: str) -> None:
 
     @staticmethod
     def raise_if_not_ignored(
-        line: str, ignore_parse_errors: set[str], persistent_states: dict[str, Any]
+        filename: str,
+        lineno: int,
+        line: str,
+        ignore_parse_errors: set[str],
+        persistent_states: dict[str, Any],
```
Spacetown, line 112:
```suggestion
        line: str, ignore_parse_errors: set[str], persistent_states: dict[str, Any],
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -135,7 +145,9 @@ def raise_if_not_ignored(
             if "suspicious_hits.warn_once_per_file" in persistent_states:
                 persistent_states["suspicious_hits.warn_once_per_file"] += 1
             else:
-                LOGGER.warning(f"Ignoring suspicious hits in line {line!r}.")
+                LOGGER.warning(
+                    f"Ignoring suspicious hits in {filename}, line {lineno}: {line!r}."
```
Spacetown, line 149:
```suggestion
                    f"Ignoring suspicious hits in {{':'.join(persistent_states['location'])}}: {line!r}."
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -151,20 +163,42 @@ def check_hits(
...
+    gcovr.formats.gcov.parser.common.SuspiciousHits: Got suspicious hit value in gcov line 'code' caused by a
+    bug in gcov tool, see
+    https://gcc.gnu.org/bugzilla/show_bug.cgi?id=68080. Use option
+    --gcov-ignore-parse-errors with a value of suspicious_hits.warn,
+    or suspicious_hits.warn_once_per_file or change the threshold
+    for the detection with option --gcov-suspicious-hits-threshold.
+    >>> check_hits(-1, "code", {"all"}, 10, {"reading_line": ("file", 5)})
     0
-    >>> check_hits(1000, "", {"all"}, 10, {})
+    >>> check_hits(1000, "code", {"all"}, 10, {"reading_line": ("file", 5)})
```
Spacetown, line 187:
```suggestion
    >>> check_hits(1000, "code", {"all"}, 10, {"location": ("file", 5)})
```

Spacetown (commented):
Instead of `reading_line` I would use `location` and extract it from persistent_states where needed.

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -45,7 +45,7 @@ def raise_if_not_ignored(
     ) -> None:
         """
         Raise exception if not ignored by options
-        >>> state = dict()
+        >>> state = dict(location=("file", 5))
```
Spacetown, line 48: In this test the location isn't used. Shall the exception not also be extended by the line information?

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -103,7 +105,7 @@ def raise_if_not_ignored(
...
```
Spacetown, line 108: Here also extend the exception message with the location?

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -151,11 +155,28 @@ def check_hits(
 ) -> int:
     """
     Check if hits count is negative or suspicious, if the issue is ignored returns 0
-    >>> check_hits(1, "", {}, 10, {})
+    >>> check_hits(1, "code", {}, 10, {"location": ("file", 5)})
```
Spacetown, line 158: The location isn't used.
```suggestion
    >>> check_hits(1, "code", {}, 10, {})
```

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -151,11 +155,28 @@ def check_hits(
...
     1
-    >>> check_hits(-1, "", {"all"}, 10, {})
+    >>> check_hits(-1, "code", {}, 10, {"location": ("file", 5)})
```
Spacetown, line 160: Here also extend the exception message with the location?

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -151,11 +155,28 @@ def check_hits(
...
+    gcovr.formats.gcov.parser.common.SuspiciousHits: Got suspicious hit value in gcov line 'code' caused by a
+    bug in gcov tool, see
+    https://gcc.gnu.org/bugzilla/show_bug.cgi?id=68080. Use option
+    --gcov-ignore-parse-errors with a value of suspicious_hits.warn,
+    or suspicious_hits.warn_once_per_file or change the threshold
+    for the detection with option --gcov-suspicious-hits-threshold.
+    >>> check_hits(-1, "code", {"all"}, 10, {"location": ("file", 5)})
     0
-    >>> check_hits(1000, "", {"all"}, 10, {})
+    >>> check_hits(1000, "code", {"all"}, 10, {"location": ("file", 5)})
```
Spacetown, line 179: Not used.
```suggestion
    >>> check_hits(1000, "code", {"all"}, 10, {})
```

Spacetown (changes requested):
The dockets needs to be updated and I think it would be good to have the location also in the exception message and not only in the log entry if the error is ignored.

src/gcovr/formats/gcov/parser/common.py
kenchung285: I don't fully understand what you means. So I suggest that you are asking me about why adding the location into the state here but no related tests in this doctest.  

First, the `location` key is needed because it will be parsed at Line 80 - 82. The function need a key inside the `persistent_state`.  
Second, I think the logger we used here send message to `stderr` which can't be checked in a doctest.

src/gcovr/formats/gcov/parser/common.py
kenchung285: Similar to [this comment](https://github.com/gcovr/gcovr/pull/1075#discussion_r2009019595).  
The key is parsed and required in `NegativeHits.raise_if_not_ignored` and `SuspiciousHits.raise_if_not_ignored`

Spacetown: But here the function isn't called because we have a hit of `1` and therefore the data isn't used.

src/gcovr/formats/gcov/parser/common.py
kenchung285: Got it. I'll rewrite the message.

src/gcovr/formats/gcov/parser/common.py
kenchung285: I see. Let me remove these data

src/gcovr/formats/gcov/parser/common.py
```diff
@@ -109,8 +109,8 @@ def raise_if_not_ignored(
         >>> SuspiciousHits.raise_if_not_ignored("code", None, state)
         Traceback (most recent call last):
             ...
-        gcovr.formats.gcov.parser.common.SuspiciousHits: Got suspicious hit value in gcov line 'code' caused by a
-        bug in gcov tool, see
+        gcovr.formats.gcov.parser.common.SuspiciousHits: Got suspicious hit value in file:5: code
+        This is a bug in gcov tool, see
```
Spacetown, line 113:
```suggestion
        This is caused by a bug in gcov tool, see
```

Spacetown (approved):
Thanks for implementing this.

PR #1078, opened 2025-03-16
Refactor generation and reading of JSON report

Spacetown:
The report generation is changed to a serialisation of the data classes and the reading to a deserialisation instead of doing it in the report generation module.

Update the known bugs of the existing releases until 6.0 and add links to the releases which fixed the bug. Also update the automation for new version and new release iteration to update the version references.


src/gcovr/formats/gcov/parser/json.py
```diff
@@ -197,9 +197,9 @@ def _parse_file_node(
             )
         for index, condition in enumerate(line.get("conditions", [])):
             linecov.insert_condition_coverage(
-                index,
                 ConditionCoverage(
                     str(data_fname),
+                    conditionno=index,
```
dlegaultbbry, line 202: How does this new parameter affect merging of conditions from the previous problem that was encountered?

dlegaultbbry (commented):
I haven't found anything in reading the rest of the changes.

src/gcovr/formats/gcov/parser/json.py
Spacetown: For now it's only moved from the function `insert_condition_coverage` to the `ConditionCoverage` class that all classes track their id. This is a compatible change but for your problem I want to create a test in a follow up PR. If I write to if conditions in one line I should be able to force the error you received because the current implementation wants to merge them.

dlegaultbbry (approved):

Spacetown:
No. I increased it in this PR but it wasn't needed because the reference is the same. Thinking mor about, it should be increased because I added the excluded flag or this change should be done in a separate PR because it' independent from the refactoring.

dlegaultbbry:
Really up to you if you want to move the addition of a field to a separate PR. It would definitely be cleaner.

PR #1093, opened 2025-04-11
Documentation formatting fixes

Pesa:
Most docutils `include` directives have been replaced with sphinx `literalinclude` directives. For reasons unknown to me, the former does not get rendered correctly into HTML. I haven't tested other output formats. I also squeezed in a couple of minor typo fixes.


latk (approved):
Thank you for all these documentation fixes! This weird formatting was always bugging me, and I'm glad that you figured out how to solve this.

Consider adding a changelog entry to note these documentation improvements.

doc/source/faq.rst
```diff
@@ -82,15 +82,15 @@ because GCC's gcov tool does not make the necessary information available,
 and because different projects are interested in different kinds of branches.
 However, gcovr has the following options to *reduce* unwanted branches:
 
-With the :option:`gcovr --exclude-unreachable-branches` option,
-gcovr parses the *source code* to see whether that line even contains any code.
+With the :option:`--exclude-unreachable-branches <gcovr --exclude-unreachable-branches>`
```
Spacetown, line 85: Is the ` <gcovr --exclude-unreachable-branches>` needed? I used it always without it and see no difference.

doc/source/output/json.rst
```diff
@@ -1,12 +1,11 @@
-
-.. program::  gcovr
-
 .. _json_output:
 
 JSON Output
 ===========
 
-The ``gcovr`` command can also generate a JSON output using
+.. program:: gcovr
```
Spacetown, line 6: Why moving this in between the text?

Spacetown (commented):
Two questions only.

doc/source/faq.rst
Pesa: `:option:` works like other sphinx roles for cross-referencing: the `<foo>` part is the target, it is necessary if the text you want to display is different from the target. You can omit it if you don't need to override the displayed text. In this case I wanted to omit the "gcovr" part from the displayed text because it seemed redundant, and also for consistency with the rest of the documentation where the vast majority of options are referenced without "gcovr".

Or are you saying that `:option:--exclude-unreachable-branches` should work by itself, without "gcovr" at all?

doc/source/output/json.rst
Pesa: Just for consistency with other files, e.g., `clover.rst`

Pesa:
> Consider adding a changelog entry to note these documentation improvements.

Done.

doc/source/output/json.rst
Spacetown: Tier it is between heading and text which looks better in your opinion.

doc/source/faq.rst
Spacetown: Yes, this is how it's done in other parts of the documentation. But this needs an explicit `.. program:: gcovr` in the file.

Pesa: Ok I see, I can do it that way if you prefer.

Spacetown: It's less to write and I think easier to read in the source. Would be good if it's aligned in the whole documentation.

Spacetown: It's less to write and I think easier to read in the source. Would be good if it's aligned in the whole documentation.

Pesa: agreed.

Spacetown:
@pesa After a review is done please do not rewrite the history. GH looses track to the changes to my last review and I need to visit all files again.

PR #1092, opened 2025-04-09
Fix merging off different reported data across GCOV data files

Spacetown:
This PR merge the coverage data of elements with same properties. If e.g. the branches for a line have different branch numbers, source block ids or destination block ids they are kept separate instead of raising an error if the block numbers differ for the same branch number.

If a branch or condition number is available in the input file it is added to the JSON report.

If a report requires the branch numbers they are generated by sorting the branches by their properties, starting with index 0. The branch numbers in the reports can be changed.

Related #1009 
Closes #1081

PR #1094, opened 2025-04-14
Improve debug of data errors

Spacetown:
This PR adds back references into the data model to get the location information if a data error is raised. Also we are able to read in JSON data files without having access to the original sources to analyse user errors.

### DIFF
--- a/src/gcovr/formats/gcov/parser/json.py
+++ b/src/gcovr/formats/gcov/parser/json.py
@@ -89,14 +89,19 @@ def parse_coverage(
                 LOGGER.warning(
                     f"File {max_line_number} has {lines} line(s) but coverage data has {max_line_number} line(s)."
                 )
+                # GCOV itself adds the /*EOF*/ in the text report if there is no data and we used the same.
                 source_lines += [b"/*EOF*/"] * (max_line_number - lines)
         except OSError as e:
             if file["file"].endswith("<stdin>"):
-                message = f"Got sourcefile {file['file']}, using empty lines."
+                message = f"Got unreadable source file '{file['file']}', replacing with empty lines."
                 LOGGER.info(message)
             else:
+                # The exception contains the source file name,
+                # e.g. [Errno 2] No such file or directory: 'xy.txt'
                 message = f"Can't read file, using empty lines: {e}"
                 LOGGER.warning(message)
+            # If we can't read the file we use as first line the error
+            # and use empty lines for the rest of the lines.
             source_lines = [b""] * max_line_number
             source_lines[0] = f"/* {message} */".encode()
 
--- a/src/gcovr/formats/gcov/parser/json.py
+++ b/src/gcovr/formats/gcov/parser/json.py
@@ -38,14 +38,7 @@ from typing import Any, Optional
 
 from gcovr.utils import get_md5_hexdigest
 
-from ....data_model.coverage import (
-    BranchCoverage,
-    CallCoverage,
-    ConditionCoverage,
-    FileCoverage,
-    FunctionCoverage,
-    LineCoverage,
-)
+from ....data_model.coverage import FileCoverage
 from ....data_model.merging import FUNCTION_MAX_LINE_MERGE_OPTIONS, MergeOptions
 from ....filter import Filter, is_file_excluded
 from .common import (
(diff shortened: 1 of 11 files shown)


src/gcovr/formats/gcov/parser/json.py
```diff
@@ -87,26 +87,25 @@ def parse_coverage(
...
+                    f"File {max_line_number} has {lines} line(s) but coverage data has {max_line_number} line(s)."
                 )
-                # Python ranges are exclusive. We want to iterate over all lines, including
-                # that last line. Thus, we have to add a +1 to include that line.
-                for _ in range(lines, max_line_from_cdata):
-                    source_lines.append(b"/*EOF*/")
+                source_lines += [b"/*EOF*/"] * (max_line_number - lines)
+        except OSError as e:
+            if file["file"].endswith("<stdin>"):
+                message = f"Got sourcefile {file['file']}, using empty lines."
```
dlegaultbbry, line 95: `Got unreadable source file '{file['file']}', replacing with empty lines.`

src/gcovr/formats/gcov/parser/json.py
```diff
@@ -87,26 +87,25 @@ def parse_coverage(
...
+                LOGGER.info(message)
+            else:
+                message = f"Can't read file, using empty lines: {e}"
```
dlegaultbbry, line 98: `Can't read source file '{file['file']}', replacing with empty lines.`

src/gcovr/formats/gcov/parser/json.py
```diff
@@ -87,26 +87,25 @@ def parse_coverage(
...
+                LOGGER.warning(message)
+            source_lines = [b""] * max_line_number
```
dlegaultbbry, line 100: Why is above using EOF lines and this isn't?

dlegaultbbry:
So if I understand the changes, you are adding back where in the original source file the invalid data was generated from in case of an error like the conditions in !1092 ?

src/gcovr/formats/gcov/parser/json.py
Spacetown: No replacement in f-string?

dlegaultbbry: yes, you can f-string it, I was just suggesting a clearer message

src/gcovr/formats/gcov/parser/json.py
Spacetown: GCOV itself adds the `/*EOF*/` in the text report if there is no data and we used the same. It's only used for creating the MD5 sum of the line and we stub the entire file therefore I thought that `""` is better here.

Spacetown:
> So if I understand the changes, you are adding back where in the original source file the invalid data was generated from in case of an error like the conditions in #1092 ?

You can try it by running `gcovr --gcov-use-existing --gcov-keep ...` giving the directory of the two files you provided me in your issue.

src/gcovr/formats/gcov/parser/json.py
dlegaultbbry: Might be good to have a comment to explain that information. I would have never guessed.

dlegaultbbry:
> > So if I understand the changes, you are adding back where in the original source file the invalid data was generated from in case of an error like the conditions in #1092 ?
> 
> You can try it by running `gcovr --gcov-use-existing --gcov-keep ...` giving the directory of the two files you provided me in your issue.

I don't have the source anymore for the previous issue and haven't been able to see it happen again. Kind of counting on the other person which has a reproducible scenario to help figure it out with your new merge model.

src/gcovr/formats/gcov/parser/json.py
Spacetown: Ah, there is a Icon for suggestions if you add a comment. In that case I get a diff and only need to accept.

Spacetown: Ah, there is a Icon for suggestions if you add a comment. In that case I get a diff and only need to accept.

src/gcovr/formats/gcov/parser/json.py
Spacetown: The exception contains the source file name, e.g. `[Errno 2] No such file or directory: 'xy.txt'`

PR #1099, opened 2025-05-07
Fix key of calls when reading JSON format

Spacetown:
With #1090 the key in JSON was changed from `gcovr/calls` to `calls` but the deserialization still used the old key.

[no changelog]


doc/source/output/json.rst
```diff
@@ -149,16 +158,6 @@ gcovr/excluded: boolean
...
 
-gcovr/decision: object
-  The :ref:`decision <json_format_decision>` entry for this line, if any.
-  Absent if there is no decision to report.
-  Requires that :option:`--decisions` coverage analysis was enabled.
-
-gcovr/calls: object
-  The :ref:`call <json_format_call>` for this line, if any.
-  Absent if there is no call to report.
-  Requires that :option:`--calls` coverage analysis was enabled.
```
dlegaultbbry, line 160: You dropped this line when moving, I assume that was intentional?

Spacetown: No. That was the thing fixed in #1090 and they where not removed anymore when the option isn't used.

dlegaultbbry (approved):

Issue #1105, opened 2025-06-09
Error occurs in latest gcovr in gcovr.data_model


kenchung285:
# Describe the bug

An error occurs in my project when upgrading gcovr from released 8.3 to the latest main branch.
Error message:
```
(INFO) Reading coverage data...
(ERROR) Trackback (most recent call last):
  File "/home/ken/gcovr/src/gcovr/formats/gcov/workers.py", line 95, in worker
    work(*args, **kwargs)
  File "/home/ken/gcovr/src/gcovr/formats/gcov/read.py", line 467, in process_datafile
    done = run_gcov_and_process_files(
  File "/home/ken/gcovr/src/gcovr/formats/gcov/read.py", line 827, in run_gcov_and_process_files,
    process_gcov_json_data(gcov_filename, covdata, options)
  File "/home/ken/gcovr/src/gcovr/formats/gcov/read.py", line 205, in process_gcov_json_data
    covdata.insert_file_coverage(filecov, merge_options)
  File "/home/ken/gcovr/src/gcovr/formats/gcov/container.py", line 198, in insert_file_coverage
    self.data[key].merge(filecov, options)
  File "/home/ken/gcovr/src/gcovr/formats/gcov/coverage.py", line 1689, in merge
    self.lines.merge(other.lines, options)
  File "/home/ken/gcovr/src/gcovr/formats/gcov/coverage_dict.py", line 58, in merge
    self[key].merge(item, options)
  File "/home/ken/gcovr/src/gcovr/formats/gcov/coverage.py", line 1118, in merge
    self.calls.merge(other.calls, options)
  File "/home/ken/gcovr/src/gcovr/formats/gcov/coverage_dict.py", line 58, in merge
    self[key].merge(item, options)
  File "/home/ken/gcovr/src/gcovr/formats/gcov/coverage.py", line 897, in merge
    self.source_block_id = self._merge_property(
  File "/home/ken/gcovr/src/gcovr/formats/gcov/coverage.py", line 179, in _merge_property
    self.raise_merge_error(
  File "/home/ken/gcovr/src/gcovr/formats/gcov/coverage.py", line 140, in raise_merge_error
    raise GcovrMergeAssertionError(
gcovr.data_model.coverage.GcovrMergeAssertionError: /path/to/my/project/header/file:[line-number] (call 4) Source block ID must be equal, got 3 and 2.
GCOV data file of merge source is:
[...]

PR #1106, opened 2025-06-09
[Feature] Clean up temporary files generated by gcov tool when exception occurs.

### LINKED ISSUE #1096: [Feature] temporary files are not removed when error occurs.

Current `gcovr` generates some temporary files like `example.cpp##6ea9ce195z...`. It seems that these files are generated in my `home` directory and my project root directory.  

These files are not been deleted automatically if gcovr gets some error and stop, for example, a KeyboardInterrupt.  Users need to find these temporary files one by one and remove them manually. 

Remove these temporary files when catching interrupt signal if it's simple to implement.  

Or we can create a temporary folder to store these temporary file; in this way, it's easier for users to remove these temporary files and remove them manually without the need to find out where are these temporary files.

kenchung285:
# Related Issue

Please see #1096 for more information

# Tests

Due to #1105 errors in the latest main branch, this commits are manually tested on gcovr `8.3`. The manual test includes multi-thread situations.

# Close Issue

Closes #1096

### DIFF
--- a/src/gcovr/formats/gcov/read.py
+++ b/src/gcovr/formats/gcov/read.py
@@ -861,9 +861,12 @@ def run_gcov_and_process_files(
         if options.gcov_keep and done:
             # Keep the files with unique names
             basename = os.path.basename(abs_filename)
-            for filepath in active_gcov_files:
-                directory, filename = os.path.split(filepath)
-                os.replace(filepath, os.path.join(directory, f"{basename}.{filename}"))
+            for gcov_filename in active_gcov_files:
+                if os.path.exists(gcov_filename):
+                    directory, filename = os.path.split(gcov_filename)
+                    os.replace(
+                        gcov_filename, os.path.join(directory, f"{basename}.{filename}")
+                    )
         else:
             # Remove the used files
             remove_existing_files(list(active_gcov_files))


Spacetown:
Thanks for working on this but I think there is an easier way to I need to check it after my vacation on my computer.

src/gcovr/formats/gcov/read.py
```diff
@@ -82,14 +82,20 @@ def read_report(options: Options) -> CoverageContainer:
...
         for filename in sorted(datafiles):
-            pool.add(process_file, filename)
-        contexts = pool.wait()
+            pool.add(process_file, filename, temporary_files=temporary_files)
+        try:
+            contexts = pool.wait()
+        except:
+            for file_path in temporary_files:
+                if os.path.exists(file_path):
+                    os.remove(file_path)
```
Spacetown, line 98: I think the `temporary_files` return value is not needed here.

src/gcovr/formats/gcov/read.py
```diff
@@ -844,6 +853,8 @@ def run_gcov_and_process_files(
             ):
                 if os.path.exists(filepath):
                     os.remove(filepath)
+                    if filepath in temporary_files:
+                        temporary_files.remove(filepath)
```
Spacetown, line 857: This for loop should be spliced. In line 794 after calling `select_gcov_files_from_stdout` we should delete the files `list(all_gcov_files - active_gcov_files)` and at the end in a finally block we should delete all files in `active_gcov_files` unless `options.gcov_keep` is set or we are not `done`.
In that case we do not need the `temporary_files`. Correct me if I'm wrong here.

Spacetown (changes requested):
I think we do not need the `temporary_files` isn't needed and we should better handle it in the worker thread.
Please update the changeling and add yourself to the list of contributors.

Spacetown:
@kenchung285 Can you reply to the comments?

kenchung285:
Hi @Spacetown, I have replied to those comments and ping you for review.
But now I find those comments marked as `Pending` by github.
Do you know how to make these comments ready or just not labeled as `Pending`?

src/gcovr/formats/gcov/read.py
kenchung285: In this way, I think we are going to delete `active_gcov_files` in each thread when exception occurs. 

(The following part is partially wrong, please refer to the latest comment)
~~However, according to my understanding, an exception will only be captured once in a multi-thread process, that is to say, only one thread that capture the exception will go to the `finally` block and delete those `active_gcov_files` generated by the thread and keep those `active_gcov_files` generated by other threads remain.~~

Therefore, in my implementation, I don't handle the exception in each worker thread, and only capture it in the main thread in line 95. In this way, we can ensure only the main thread will capture the exception and delete those `active_gcov_files` through `temporary_files`

kenchung285: Hi, here's some update about the exception in a python multi-threading process after my survey and manual test with some python script

A signal ( KeyboardInterrupt is SIGINT, one of the signal to process ) will only be sent to the main thread in a python multi-threading process. See [python official document on signal vs thread](https://python.readthedocs.io/fr/latest/library/signal.html#signals-and-threads) for more information.

Therefore, if we handle these exceptions in worker threads, those `finally` block won't be triggered if an keyboardInterrupt occurs.

Spacetown:
> Hi @Spacetown, I have replied to those comments and ping you for review. But now I find those comments marked as `Pending` by github. Do you know how to make these comments ready or just not labeled as `Pending`?

You pressed the "Start review" button and in that case the comments are pending until you finish the review.

src/gcovr/formats/gcov/read.py
Spacetown: On keyboard interrupt it's fine for me to left the file because the user wants to abort (same as sending SIG_TERM or SIG_KILL).
We should delete the files as soon and as local as possible. Also if someone wants to keep the files why shall they be deleted if an interrupt occurs?

kenchung285: I see. But is there a possible situation that someone wants to keep the files when they interrupt on their own? Those files are automatically removed by worker threads in just few seconds even if user not interrupt the files.

kenchung285: Or if we need to support keeping these `active_gcov_file`, we may need a new option for user to take control of cleaning up the files or not.

Spacetown: As I said, the not used files can be removed directly after `select_gcov_files_from_stdout`:
And the rest of the files can be removed at the end:
```python
    filename = None
    out = None
    err = None
    active_gcov_files = []
    try:
        ...
        with locked_directory(chdir):
            ...
            # find the files that gcov created
            active_gcov_files, all_gcov_files = select_gcov_files_from_stdout(
                out,
                gcov_filter=options.gcov_filter,
                gcov_exclude=options.gcov_exclude,
                chdir=chdir,
            )
            # Remove the not used files immediately
            for filepath in list(all_gcov_files - active_gcov_files)
                    if os.path.exists(filepath):
                        os.remove(filepath)
            ...
    except RuntimeError as exc:
        # If we got an merge assertion error we must end the processing
        done = False
        error(
            f"Trouble processing {abs_filename!r} with working directory {chdir!r}.\n"
            f"Stdout of gcov was >>{out}<< End of stdout\n"
            f"Stderr of gcov was >>{err}<< End of stderr\n"
            f"Exception was >>{str(exc)}<< End of stderr\n"
            f"Current processed gcov file was {filename!r}.\n"
            "Use option --verbose to get extended information."
        )
    finally:
        if options.gcov_keep and done:
            # Keep the files with unique names
            basename = os.path.basename(abs_filename)
            for file in active_gcov_files:
                directory, filename = os.path.split(file)
                os.replace(file, os.path.join(directory, f"{basename}.{filename}"))
        else:
            # Remove the files
            for filepath in active_gcov_files
                if os.path.exists(filepath):
                    os.remove(filepath)
```

kenchung285: > Also if someone wants to keep the files why shall they be deleted if an interrupt occurs

I want to know more about this consideration. If all files are removed at the end of the process, users are not easily to know what files will be kept when an interrupt occurs, then in what situation, users may want to keep the files?

Spacetown: If the user sets `options.gcov_keep` he wants to keep the files. If we remove all not used files direct after creation the user can be sure that the present files are used.

kenchung285: I see. Then if the use doesn't set `options.gcov_keep` and the user abort gcovr, should we remove `active_gcov_file` ? 

Take myself, as a user of `gcovr`, for example, I sometimes aborted `gcovr` and before my next run, I need to remove temporary files manually. So I recommend to remove these `active_gcov_file` if user doestn't set `options.gcov_keep`

Spacetown: I pushed on your branch. Now on keyboard interrupt the pool is drained and all the files are removed depending on the option `--gcov-keep`. Please can you check this on your site?

kenchung285:
> Please add yourself to the list of contributors and update the CHANGELOG.

Sorry for late reply. I've update the CHANGELOG and I've been added to the list of authors in https://github.com/gcovr/gcovr/pull/1075

Spacetown (approved):
LGFM

mariustache:
Hi @kenchung285, @Spacetown. I was trying to use latest development to test the `--sonarqube-metric` feature recently added (thanks for that) and I ran into issues. After doing a git bisect, it seems this commit causes it.

My env:
```
LLM version 20.1.0
macOS Sequoia version 15.5
```

What I am using to install `gcovr`:
```
pip install git+https://github.com/gcovr/gcovr.git@<commit_id>
```

and my simplified command for generating the coverage reports:
```
gcovr -j32 --gcov-executable 'llvm-cov gcov' --sonarqube gcovr.xml
```

Initially, I thought it might be a compatibility issue, but it seems not, since it works with different `gcovr` commits.

The error looks like:
```
(ERROR) - MainThread - Traceback (most recent call last):
  File "/Users/me/venvs/conan/lib/python3.9/site-packages/gcovr/formats/gcov/workers.py", line 95, in worker
    work(*args, **kwargs)
  File "/Users/me/venvs/conan/lib/python3.9/site-packages/gcovr/formats/gcov/read.py", line 475, in process_datafile
    done = run_gcov_and_process_files(
  File "/Users/me/venvs/conan/lib/python3.9/site-packages/gcovr/formats/gcov/read.py", line 837, in run_gcov_and_process_files
    raise AssertionError(
AssertionError: Sanity check failed, output file /Users/me/git/myproject/build/Debug/lib/notify_monitor/CMakeFiles/lib_notify_monitor.dir/./vector##19f8b70c96fd5538770f5498829618df.gcov doesn't exist but no error from GCOV detected.

(ERROR) - MainThread - Error occurred while reading reports:
  File "/Users/me/venvs/conan/lib/python3.9/site-packages/gcovr/__main__.py", line 384, in main
    covdata = gcovr_formats.read_reports(options)
  File "/Users/me/venvs/conan/lib/python3.9/site-packages/gcovr/formats/__init__.py", line 98, in read_reports
    covdata = GcovHandler(options).read_report()
  File "/Users/me/venvs/conan/lib/python3.9/site-packages/gcovr/formats/gcov/__init__.py", line 231, in read_report
    return read_report(self.options)
  File "/Users/me/venvs/conan/lib/python3.9/site-packages/gcovr/formats/gcov/read.py", line 93, in read_report
    contexts = pool.wait()
  File "/Users/me/venvs/conan/lib/python3.9/site-packages/gcovr/formats/gcov/workers.py", line 181, in wait
    raise RuntimeError(
RuntimeError: Worker thread raised exception, workers canceled.
```

Not sure what is happening. Do you have some thoughts/ideas on why this happens?

Spacetown:
@mariustache I think I've found the problem. We remove the files in the finally but the directory lock is already released. I'll create a fix for this and would be happy if you can test it.
Can you try to run without `-j32`?

PR #1114, opened 2025-07-05
Add trusted publishing

### LINKED ISSUE #1111: [CI/CD] Use Trusted Publishing when uploading to PyPI
Here's the guide: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

Among benefits, you can stop keeping the long-lived API token in GH secrets. Additionally, this mechanism will produce and upload PEP 740 digital attestations for you: https://blog.pypi.org/posts/2024-11-14-pypi-now-supports-digital-attestations/ / https://blog.trailofbits.com/2024/11/14/attestations-a-new-generation-of-signatures-on-pypi/ / https://peps.python.org/pep-0740/.

The ecosystem is slowly moving towards publishing this sort of provenance information: https://trailofbits.github.io/are-we-pep740-yet/.

P.S. Implementing this will require splitting a CI job for uploading from the job for building the dists and having privilege separation. But nothing too complicated beyond that.

Spacetown:
Add trusted publishing as suggested in #1111.

Closes #1111.


noxfile.py
webknjaz, line 1: I recommend keeping this for emergencies.

src/gcovr/formats/gcov/read.py
webknjaz, line 1: This be in this PR?

.github/workflows/CI.yml
```diff
@@ -196,8 +196,15 @@ jobs:
     runs-on: ubuntu-22.04
     permissions:
       contents: write
+      id-token: write  # IMPORTANT: mandatory for trusted publishing
+
+    environment:  # Environment for trusted publishing
+      name: test.pypi
+      url: https://test.pypi.org/p/gcovr
```
webknjaz, line 203: These privileges should only be given to a separate publishing job which should only contain downloading the dists and publishing them. The transitive build deps and any other software shouldn't have access to OIDC due to future risks of impersonation / privilege escalation. Additionally, things like checking if the dists are buildable should be running continuously, on PRs, in contexts where releasing isn't happening.

noxfile.py
Spacetown: If the pipeline is working we can trigger it manual.

Spacetown:
@webknjaz [This](https://github.com/gcovr/gcovr/actions/runs/16126694419) is a manual triggered run but the upload on test.pypy.org is failing because of a development version.

webknjaz:
> @webknjaz [This](https://github.com/gcovr/gcovr/actions/runs/16126694419) is a manual triggered run but the upload on test.pypy.org is failing because of a development version.

@Spacetown no, it's not because of `.dev`. It's because of the `+gXXXXXX` part, which is what PEP 440 calls a local version identifier and explicitly forbids public indices from accepting. So (Test)PyPI will reject any dists that have that.

The error message from PyPI is misleading and I filed an issue about that @ https://github.com/pypi/warehouse/issues/18348.

webknjaz:
> If in the future shared workflows are working for trusted publishing the can be changed to input variables for the shared workflow..

@Spacetown I'm going to propose such a reusable workflow to be hosted under PyPA. There's an accidental corner case, letting people use reusable workflows sometimes. But that's actually a bug in PyPI. We brainstormed how to fix it seamlessly at PyCon US Sprints: https://github.com/pypi/warehouse/issues/11096#issuecomment-2895081700.

.github/actions/setup-runner/action.yml
```diff
@@ -4,46 +4,58 @@ name: Set up Runner
...
-      run: |
-        choco install ninja
 
-    - name: Install ninja and libxml2 (MacOs)
-      if: ${{ startsWith(inputs.os,'macos-') }}
+    # ------------------------------ Python, all OS ------------------------------
+    - uses: actions/setup-python@v5
+      with:
+        python-version: ${{ inputs.python-version }}
+    - id: pip-cache
```
webknjaz, line 48: FYI, I made an action that is a bit more clever about caching and is ABI-aware: https://github.com/marketplace/actions/cache-python-deps.

Spacetown: For now I would keep it as it is.

.github/workflows/gcovr-ci-job.yml
```diff
@@ -87,28 +92,6 @@ jobs:
           # Set the CC environment
           echo "CC=${{ inputs.gcc }}" >> $GITHUB_ENV
 
-          if [ -z "$(sed -n '/Next Release/ p' CHANGELOG.rst)" ] ; then
```
webknjaz, line 90: @Spacetown if this ever comes up, here's how I make `setuptools-scm` produce a version string without the local part in contexts where this is needed (merging to the default branch in my case): https://github.com/cherrypy/cheroot/blob/fdaa24d/.github/reusables/tox-dev/workflow/reusable-tox/hooks/prepare-for-tox-run/action.yml#L82-L125 + https://github.com/cherrypy/cheroot/blob/fdaa24d/.github/workflows/ci-cd.yml#L245-L265.

.github/workflows/CI.yml
```diff
@@ -147,150 +321,89 @@ jobs:
...
       matrix:
         gcc: [gcc-5, gcc-6, gcc-8, gcc-9, gcc-10, gcc-11, gcc-12, gcc-13, gcc-14, clang-10, clang-13, clang-14, clang-15, clang-16]
 
-  validate-reports:
-    needs: [changelog-check, spell-check, upload-coverage]
-
+  # -----------------------------------------------------------------------------------------------
+  publish:
+  # -----------------------------------------------------------------------------------------------
+    needs: [build-package, documentation, host, docker]
```
webknjaz, line 344: @Spacetown consider integrating https://github.com/marketplace/actions/alls-green#why as a CI gate that is possible to add to a branch protection ruleset as a single check with no need to manually enumerate all the jobs in the test matrix.

.github/workflows/CI.yml
```diff
@@ -147,150 +321,89 @@ jobs:
...
-        if: ${{ (github.repository == 'gcovr/gcovr') && (github.event.ref == 'refs/heads/main') && env.GCOVR_VERSION != '' }}
+        if: ${{ needs.build-package.outputs.publish_push_tag == 'true' }}
         run: |
-          git push origin "refs/tags/$GCOVR_VERSION"
+          git push origin "refs/tags/${{ needs.build-package.outputs.tag }}"
 
       - name: Create release and upload build artifacts
-        if: ${{ (github.repository == 'gcovr/gcovr') && (github.event.ref == 'refs/heads/main') && (env.GCOVR_VERSION != '') }}
+        if: ${{ needs.build-package.outputs.publish_push_tag == 'true' }}
         uses: softprops/action-gh-release@c062e08bd532815e2082a85e87e3ef29c3e6d191
```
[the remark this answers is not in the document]
Spacetown: I want to have one job which pushes all at once. Accessing the PyPI environment is protected, it's only possible on main branch and it needs a reviewer to approve the workflow.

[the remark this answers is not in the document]
Spacetown: But adding your proposed action in https://github.com/gcovr/gcovr/pull/1114#discussion_r2195059248 is ok? Where to start end where to end, whom can I trust and whom not?

PR #1126, opened 2025-07-28
Fix handling of lines without a function

Spacetown:
There are several constellations where data was added to the wrong function:
- With gcc-5, gcc-6 and clang compiler functions with specializations
  (template functions) use one block in the output which starts with all
  names. All functions are added to the data model but only the last one
  has coverage data (of all functions).
  --> This can't be changed on our side because of missing information.
- For gcc-8 and newer (unless GCOV JSON is used) the output of functions
  with specialization starts with a block with overall coverage followed
  by a block for each specialized function. If a normal function was in
  front of this block the overall counts where added to this function.
- A forced inline function does not contain a function name in the
  output. If the function is at the begin of the output it
  is ignored, in the middle of the file it is added to the previous
  function as before.

The known bugs are now in a separate file and linked from the CHANGELOG.


src/gcovr/formats/gcov/parser/text.py
```diff
@@ -396,54 +396,97 @@ def _gather_coverage_from_line(
...
 
-            filecov.insert_function_coverage(
-                filecov.data_sources,
-                MergeOptions(func_opts=FUNCTION_MAX_LINE_MERGE_OPTIONS),
-                mangled_name=name,
-                demangled_name=None,
-                lineno=lineno,
-                count=count,
-                blocks=blocks,
+            # If a specialized function is following a normal one the overall lines where added
```
dlegaultbbry, line 414: where -> were

src/gcovr/formats/gcov/parser/text.py
```diff
@@ -396,54 +396,97 @@ def _gather_coverage_from_line(
...
 
-        return _ParserState(
-            function_name=state.function_name,
-            linecov=state.linecov if linecov is None else linecov,
-            line_contents=line.source_code,
-            block_id=state.block_id,
-        )
+        is_noncode = extra_info & _ExtraInfo.NONCODE
+        if not is_noncode:
+            # This occur if a file starts with a function with specializations.
```
dlegaultbbry, line 439: This occurs ...

dlegaultbbry:
Can you provide an example output of what you mean with specializations and how the code miscounts coverage for functions?

src/gcovr/data_model/coverage.py
```diff
@@ -2009,6 +2009,15 @@ def insert_line_coverage(
 
         return self.lines[key]
 
+    def remove_line_coverage(self, linecov_list: list[LineCoverage]) -> None:
+        """Remove line coverage objects from File.
+
+        Objects can be added by the text parser if a normal function is followed by a template.
+        This specialization is detected to late and we need to remove the lines again.
```
dlegaultbbry, line 2016: ... detected too late ...

dlegaultbbry (approved):
Yes I did look at it from a functional point of view after looking into the known_bugs example to understand what it was about.

Issue #1042, opened 2024-12-21
Option to select decisions as branches for coverage reports


chandryan:
# Is your feature request related to a problem?

We use gcovr to get the branch coverage of new code within PRs. We're often struggling with False Negatives and user code using some libraries like Boost or gRPC seems to be especially affected by this. The coverage metrics from `--decisions` look like an attractive alternative, but we can't make use of that because they are not considered as input for the coverage report (I'm using `--sonarqube`).

# Describe the solution you'd like

I would like to have an option to use decision coverage as the branch coverage for the XML report that is output with the `--sonarqube` option. I'm not sure about what exact content the other types of coverage reports supported by gcovr contain, I suppose they also don't use the decisions and one option could be generically used for this?

My solution proposal is to introduce an additional option like the following:

```
    --decisions-as-branches
Use the decision coverage as branch coverage for coverage reports, which cannot show both.
Implies --decisions. Used by x, y, and SonarQube reports.
```

# Describe alternatives you've considered

I thought about some variations of the option's impact, though the proposed solution is my favorite one.

*Alternative 1: An option only for the SonarQube report*

I suppose the same principle can be applied to other reports as well.

*Alternative 2: Use decisions as branches for all reports*

For reports like the HTML reports this is not necessary, as they can show both metrics. It would also be unnecessarily confusing to actually show the decisions column as a branches column. However, it might be worth a thought colouring the lines in HTML reports according to the decisions instead of the branches.

# Additional context

Happy to contribute and create a PR in case you agree with the idea 😄

Spacetown:
The option could be reused but I'm not sure if I like the idea of changing the content of a report file by adding the "wrong" data to the elements. 
Also I wouldn't change the coloring of the html report.

@latk what do you think about this idea?

Spacetown:
> However, I'm really not a fan of the `--decisions` feature. Unless GCC 14+ is used (where we can take condition-coverage data directly from the gcov JSON), the decision analyzer is super fragile, makes unreasonably strong assumptions about the formatting of the source code, and only works for .gcov file inputs. 

I suggest to add a warning if this feature is used without having the JSON files of gcc >= 14.

[...]

PR #1132, opened 2025-08-02
Add option `--sonarqube-metric` to define which metrics to report

chandryan:
Introduce an option ``--sonarqube-metric`` that lets the user define which metrics are printed in the SonarQube XML report.

Related to https://github.com/gcovr/gcovr/issues/1042

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,7 @@ New features and notable changes:
 - Add branch information to ``Coveralls`` report. (:issue:`1121`)
 - Add support to define :option:`--exclude-lines-by-pattern` and :option:`--exclude-branches-by-pattern`
   more than once. (:issue:`1129`)
+- Add support to define :option:`--sonarqube-metric` (:issue:`1042`)
 
 Bug fixes and small improvements:
 
--- a/src/gcovr/formats/sonarqube/__init__.py
+++ b/src/gcovr/formats/sonarqube/__init__.py
@@ -30,6 +30,8 @@ class SonarqubeHandler(BaseHandler):
     @classmethod
     def get_options(cls) -> list[Union[GcovrConfigOption, str]]:
         return [
+            # Global options needed for report
+            "show_decision",
             GcovrConfigOption(
                 "sonarqube",
                 ["--sonarqube"],
@@ -43,6 +45,15 @@ class SonarqubeHandler(BaseHandler):
                 type=OutputOrDefault,
                 default=None,
                 const=OutputOrDefault(None),
+            ),
+            GcovrConfigOption(
+                "sonarqube_metric",
+                ["--sonarqube-metric"],
+                config="sonarqube-metric",
+                group="output_options",
+                help=("The metric type to report."),
+                choices=("line", "branch", "decision"),
+                default="branch",
             ),
         ]
 
--- a/src/gcovr/formats/sonarqube/write.py
+++ b/src/gcovr/formats/sonarqube/write.py
@@ -17,9 +17,11 @@
 #
 # ****************************************************************************
 
+from typing import Optional, Union
 from lxml import etree  # nosec # We only write XML files
 
 from ...data_model.container import CoverageContainer
+from ...data_model.stats import CoverageStat, DecisionCoverageStat
 from ...options import Options
 from ...utils import write_xml_output
 
@@ -28,6 +30,11 @@ def write_report(
     covdata: CoverageContainer, output_file: str, options: Options
 ) -> None:
     """produce an XML report in the SonarQube generic coverage format"""
+
+    if options.sonarqube_metric == "decision" and not options.show_decision:
+        raise AssertionError(
+            "--sonarqube-metric=decision needs the option --decisions."
+        )
 
     root_elem = etree.Element("coverage")
     root_elem.set("version", "1")
@@ -44,8 +51,12 @@ def write_report(
                 line_node.set("lineNumber", str(linecov.lineno))
                 line_node.set("covered", "true" if linecov.is_covered else "false")
 
-                if linecov.branches:
+                stat: Optional[Union[CoverageStat, DecisionCoverageStat]] = None
+                if options.sonarqube_metric == "branch" and linecov.branches:
                     stat = linecov.branch_coverage()
+                elif options.sonarqube_metric == "decision" and linecov.decision:
+                    stat = linecov.decision_coverage()
+                if stat:
                     line_node.set("branchesToCover", str(stat.total))
                     line_node.set("coveredBranches", str(stat.covered))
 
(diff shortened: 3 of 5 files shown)


src/gcovr/formats/sonarqube/write.py
```diff
@@ -44,10 +51,18 @@ def write_report(
...
-                    line_node.set("branchesToCover", str(stat.total))
-                    line_node.set("coveredBranches", str(stat.covered))
+                if options.sonarqube_metric != "line":
+                    stat: Optional[Union[CoverageStat, DecisionCoverageStat]] = None
+                    if options.sonarqube_metric == "branch" and linecov.branches:
+                        stat = linecov.branch_coverage()
+                    elif options.sonarqube_metric == "decision" and linecov.decision:
+                        stat = linecov.decision_coverage()
+                    else:
+                        raise RuntimeError("Unknown value {options.sonarqube_metric} for --sonarqube-metric")
```
chandryan: I had to remove this code, because this check does not work with the current conditionals: It also triggers in scenarios where a line does not have branch or decision coverage.

Do you see a benefit in having it?
The values are already checked by the arg parser, which means this code is not reachable.

doc/source/faq.rst
```diff
@@ -157,3 +157,4 @@ The following options are only used if available:
 - ``--json-format``: Use JSON intermediate format. Can only be used if no calls are needed in the report.
 - ``--demangled-names``: Not available for LLVM based ``gcov``.
 - ``--hash-filenames``: Available since GCC 7, as fallback the option ``--preserve-paths`` is used.
+- ``--condition``: Available since GCC 14, additionally requires the compiler to be invoked with ``-fcondition-coverage``.
```
chandryan, line 160: Double-checking, it seems this option is not correct? Do you have a test case which confirms this setting is valid?

I took it from `src/gcovr/formats/gcov/read.py`, but according to the [gcov documentation](https://gcc.gnu.org/onlinedocs/gcc/Invoking-Gcov.html) the option should be ``--conditions`` instead of ``--condition``.

src/gcovr/formats/sonarqube/__init__.py
```diff
@@ -44,8 +46,26 @@ def get_options(cls) -> list[Union[GcovrConfigOption, str]]:
...
+                ["--sonarqube-metric"],
+                config="sonarqube-metric",
+                group="output_options",
+                help=("The metric type to report. Default is '{default!s}'."),
+                choices=("line", "branch", "condition", "decision"),
+                default="branch",
+            ),
         ]
 
+    def validate_options(self) -> None:
```
chandryan, line 60: For the option `--sonarqube-metric=condition`, can we do some validation?

I suppose it's not possible to check whether the file has been compiled with `-fcondition-coverage`, but would it be possible to check if we were able to enable `--condition(s)` in gcov?

chandryan (commented):
Posted comments to discuss some details about gcc's condition coverage and the suggested validation in `write_report`

src/gcovr/formats/sonarqube/write.py
```diff
@@ -29,6 +31,11 @@ def write_report(
 ) -> None:
     """produce an XML report in the SonarQube generic coverage format"""
 
+    if options.sonarqube_metric == "decision" and not options.show_decision:
+        raise AssertionError(
+            "--sonarqube-metric=decision needs the option --decisions."
+        )
+
```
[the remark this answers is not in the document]
chandryan: Moved code to `validate_options` and changed `AssertionError` to `RuntimeError` like in the other handlers.

chandryan:
TODO: I still need to add tests

src/gcovr/formats/sonarqube/__init__.py
```diff
@@ -44,6 +46,15 @@ def get_options(cls) -> list[Union[GcovrConfigOption, str]]:
                 default=None,
                 const=OutputOrDefault(None),
             ),
+            GcovrConfigOption(
+                "sonarqube_metric",
+                ["--sonarqube-metric"],
+                config="sonarqube-metric",
+                group="output_options",
+                help=("The metric type to report."),
+                choices=("line", "branch", "decision"),
```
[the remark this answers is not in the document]
chandryan: Added the option ``--sonarqube-metric=condition``

doc/source/faq.rst
chandryan: It seems gcov accepts both. Please give feedback whether you prefer to leave it as is or update to be aligned with the documentation.

Spacetown: Please can you update it to be aligned with GCOV documentation.

src/gcovr/formats/sonarqube/__init__.py
Spacetown: The available options of GCOV are not available here. What we can do is to print a warning if we do not have any conditions:
```python
if not any(
        filter(lambda filecov: filecov.condition_coverage().total > 0, covdata.values())  # type: ignore [arg-type]
    ):
    LOGGER.warning(...)
```

src/gcovr/formats/sonarqube/write.py
```diff
@@ -44,10 +46,18 @@ def write_report(
                 line_node.set("lineNumber", str(linecov.lineno))
                 line_node.set("covered", "true" if linecov.is_covered else "false")
 
-                if linecov.branches:
-                    stat = linecov.branch_coverage()
-                    line_node.set("branchesToCover", str(stat.total))
-                    line_node.set("coveredBranches", str(stat.covered))
+                if options.sonarqube_metric != "line":
```
Spacetown, line 49: I'm unsure with the option to only have line coverage. Do you think that this is required?

CHANGELOG.rst
```diff
@@ -33,6 +33,7 @@ New features and notable changes:
 - Add branch information to ``Coveralls`` report. (:issue:`1121`)
 - Add support to define :option:`--exclude-lines-by-pattern` and :option:`--exclude-branches-by-pattern`
   more than once. (:issue:`1129`)
+- Add :option:`--sonarqube-metric` to define which metric is used as  branch metric in report. (:issue:`1042`)
```
Spacetown, line 36:
```suggestion
- Add :option:`--sonarqube-metric` to define which metric is used as branch metric in report. (:issue:`1042`)
```

doc/source/faq.rst
chandryan: Updated to `--conditions` as per GCOV documentation.

src/gcovr/formats/sonarqube/write.py
```diff
@@ -25,12 +26,19 @@
...
+
 
 def write_report(
     covdata: CoverageContainer, output_file: str, options: Options
 ) -> None:
     """produce an XML report in the SonarQube generic coverage format"""
 
+    if not any(
+        filter(lambda filecov: filecov.condition_coverage().total > 0, covdata.values())  # type: ignore [arg-type]
+    ):
```
Spacetown, line 39:
```suggestion
    ) and (options.sonarqube_metric == "condition"):
```

Spacetown (changes requested):
One minor change needed. The warning should only be printed if conditions are requested.

Spacetown (approved):

PR #1153, opened 2025-09-24
Add excluded metrics

### LINKED ISSUE #988: Show metrics for excluded items in reports

Spacetown:
Add excluded metrics to the HTML report summary of directory and files:
image

Closes #988


dlegaultbbry:
Trying to make sense of the stats and perhaps that should also be explained in the docs.

For example, Lines of 15/17/10. I presume the total is without the excluded lines (meaning that there's a coverable total of 27 lines).

dlegaultbbry:
> @dlegaultbbry We've no description of the report and how the statistics are interpreted. Would it be more clear if we use the same format in the summary at the top? image
> 
> Or is it more intuitiv if we change the stats to have the excluded counted to the total but ignored for the coverage %? E.g.:
> 
> ```python
> stat = {
>     covered: 10,
>     excluded: 5,
>     total: 25,
> }
> 
> percent = (stat.covered * 100.0) / (stat.total - stat.excluded)) # -> 50%
> ```

I like option #2 better with the same columns as before in the top portion. Regardless, I think it needs to be mentioned somewhere else confusion will inevitably happen =]

PR #1155, opened 2025-09-25
Make html report's <title> more descriptive

trillllian:
Currently, the `<title>` of each page in the HTML report is identical (all of them are just `--html-title`). This makes it quite annoying to navigate between multiple files in the same report, opened in different tabs.

This PR adds the filename to the individual source file pages, and the title "functions" to the function list page. I left the directory listing page as-is (it serves as the index page).

I didn't make an issue for this, because this fix seems uncontroversial enough to me :)


Spacetown:
Thanks for implementing this. First of all, please add yourself to the list of Contributors.

I think the filename and the `Functions` should be added after the title and not in front of it and also all the reference data need to be updated as described in the [documentation](https://gcovr.com/en/latest/contributing.html#run-tests-with-docker).

Issue #1173, opened 2025-10-22
Change exclude line colour


dlegaultbbry:
# Is your feature request related to a problem?

The colour of the excluded lines in the default themes is a very faint 'pink' shade which makes it hard to find them quickly in a large source file on a white background.

# Describe the solution you'd like

Change the default excluded line colour to be more eye popping

# Describe alternatives you've considered

I tried the following and contrast is a net improvement and mixes well with other theme colours as well as leaving the code text readable:
- darkkhaki
- grey
- darkgrey

# Additional context

I can make the change if there is agreement on a particular colour.

dlegaultbbry:
> [@dlegaultbbry](https://github.com/dlegaultbbry) The tests are updated and you can start. Together with the change of the color I suggest to put following lines to a separate `color.css` and include them in the `styles.css`:
> 
> [gcovr/src/gcovr/formats/html/default/style.css](https://github.com/gcovr/gcovr/blob/55d85d4852eb0905d971a7e67d0db1bee06d4287/src/gcovr/formats/html/default/style.css#L6-L37)
> 
> Lines 6 to 37 in [55d85d4](/gcovr/gcovr/commit/55d85d4852eb0905d971a7e67d0db1bee06d4287)
>  .theme-green, .theme-blue { 
>      --unknown_color: lightgray; 
>      --low_color: #FF6666; 
>      --medium_color: #F9FD63; 
>      --partial_covered_color: var(--medium_color); 
>      --uncovered_color: #FF8C8C; 
>      --warning_color: orangered; 
>      --notTakenBranch_color: red; 
>      --notTakenCondition_color: red; 
>      --uncheckedDecision_color: darkorange; 
>      --notTakenDecision_color: red; 
>      --notInvokedCall_color: red; 
>      --excluded_color: rgb(255, 241, 229); 
>  } 
>   
>  .theme-green { 
>      --high_color: #85E485; 
>      --covered_color: #85E485; 
>      --takenBranch_color: green; 
>      --takenCondition_color: green; 
>      --takenDecision_color: green; 
>      --invokedCall_color: green; 
>  } 
>   
>  .theme-blue { 
>      --high_color: #66B4FF; 
>      --covered_color: #66B4Ff; 
>      --takenBranch_color: blue; 
>      --takenCondition_color: blue; 
>      --takenDecision_color: blue; 
>      --invokedCall_color: blue; 
>  }

[...]

PR #1179, opened 2025-10-28
split theme colours into separate CSS file (#1173)

dlegaultbbry:
- Update the default HTML theme excluded line color to darkkhaki for better visibility

First commit: making sure that the CSS color split introduces no differences in the test data
Second commit: update the default HTML theme color + update reference data


CHANGELOG.rst
```diff
@@ -17,6 +17,9 @@ Bug fixes and small improvements:
 - Fix coveralls report if there are several coverage elements per line. (:issue:`1165`)
 - Fix not excluded conditions if branches are excluded by comments. (:issue:`1171`)
 - Fix reported excluded branches in ``LCOV`` report. (:issue:`1171`)
+- Change default HTML theme excluded line color. (:issue:`1173`)
+- Move theme colors to a separate CSS file so they can be overridden with a simple
+  `--html-template-directory` directory with only the style.colors.css file (:issue:`1173`)
```
Spacetown: Here we reference the PR.
```suggestion
- Change default HTML theme excluded line color. (:issue:`1179`)
- Move theme colors to a separate CSS file so they can be overridden with a simple
  `--html-template-directory` directory with only the style.colors.css file (:issue:`1179`)
```

Spacetown (changes requested):
Split looks fine.

Spacetown (commented):
LGFM.
Please do not force push because GH "Show changes since last review" won't work anymore and the already added comments can't be shown correct.

dlegaultbbry:
Think I need some help in deciphering that failure and if it's relevant to the changes.

Spacetown (approved):
LGFM

PR #1199, opened 2025-11-18
Detect missing function lines

### LINKED ISSUE #1166: Missing coverage results after updating to 8.4
# Describe the bug

The last week or so our code coverage suddenly dropped to 0%.

Our project is built on CMake and runs coverage in two steps:
1. Uses the built-in coverage capabilities of `ctest` to call `gcov` to accumulate the results of all the tests in our project.
2. Runs `gcovr` to produce the output in the required formats and exclude certain source files (e.g. tests, generated code) for the final result.

When viewing the `--verbose` output, every single line has the following error:
> Ignoring line coverage because of missing function name

The results at the end show effectively no code, even though the debug output shows it's aware of the coverage and source:

```
lines: 0.0% (0 out of 0)
functions: 0.0% (0 out of 0)
branches: 0.0% (0 out of 0)
```

I am not 100% sure whether this is due to the accumulation of coverage data or expected command-line arguments for `gcov` differing, though I suspect it's more likely the latter. (more details in context)

# To Reproduce

1. Create a CMake project with multiple tests through `ctest`
2. Add the following compiler flags to enable coverage: `-O0 --coverage`
3. Use the following file for `DartConfiguration.tcl.in` for the base configuration for CTest:
   ```
   # Minimal Dart configuration for CMake to run coverage tests.
   SourceDirectory: @CMAKE_SOURCE_DIR@
   BuildDirectory: @CMAKE_BINARY_DIR@
   Compiler: @CMAKE_CXX_COMPILER@
   CompilerVersion: @CMAKE_CXX_COMPILER_VERSION@
   CoverageCommand: @COVERAGE_COMMAND@
   CoverageExtraFlags: @COVERAGE_EXTRA_FLAGS@
   ```
4. Add the following to hook up `gcov` to CTest, assuming `DartConfiguration.tcl.in` is at the project root:
   ```
   find_program(GCOV gcov REQUIRED)
   set(COVERAGE_COMMAND ${GCOV})
   set(COVERAGE_EXTRA_FLAGS "--long-file-names --hash-filenames")
   configure_file(${CMAKE_SOURCE_DIR}/DartConfiguration.tcl.in
    ${CMAKE_BINARY_DIR}/DartConfiguration.tcl @ONLY)
   ```
5. Run `ctest -T Test -T Coverage` to generate the accumulated coverage for all the tests.
6. Run `gcovr` with `--use-gcov-files` and `--object-directory` to the output location. (`<build>/Testing/Coverage` for our configuration)

# Expected behavior

Coverage is generated.

# Screenshots

Image

- OS: Linux (both Rocky Linux 9 and 10)
- GCC version: 13.3, 14.2
- GCOVR version: 8.4
- Project directory layout:
  - Code and CMakeLists.txt at the project root.
  - Code built to `build` subdirectory.
  - .gcov files added to `build/Testing/CoverageInfo`

# Additional context

Since we utilize CTest to accumulate the coverage results before running `gcovr`, there are limitations based on the command-line arguments. CTest is used because it allows us to centrally register and manage tests through CMake and re-use that existing system both for running our tests normally as well as part of the code coverage pipeline.

Some things I tried include:
* Enabling `--all-blocks` when calling `gcov`. (note that this is currently missing from the [FAQ for gcov arguments](https://gcovr.com/en/8.4/faq.html#which-options-are-used-for-calling-gcov))
* Enabling `--json-format` when testing with Rocky Linux 10 and GCC 14.2.

In both cases I was unable to get to the point of invoking `gcovr` because the coverage support for CTest were unable to work with the output changes and errored out. As a result I haven't been able to verify whether these arguments (especially `--all-blocks`) avoid the regression.

Spacetown:
Add a warning if function lines are missing in external GCOV files. This can occur if the data is generated outside GCOVR and option `--branch-probabilities` is missing.
To not ignore the lines of the file they are assigned to a placeholder function.

Closes #1166


src/gcovr/formats/gcov/parser/text.py
```diff
@@ -305,6 +305,19 @@ def parse_coverage(
...
             lines_with_errors.append((raw_line, ex))
 
+    if not any(isinstance(line, _FunctionLine) for line, _ in tokenized_lines):
+        file_for_message = (
+            "\n  ".join(" -> ".join(file_tuple) for file_tuple in sorted(data_filename))
+            if isinstance(data_filename, set)
+            else data_filename
+        )
+        raise RuntimeError(
+            f"No function line found in gcov file:\n   {file_for_message}\n"
```
here-abarany: Note that this has three spaces after the newline before printing `file_for_merge`, while `file_for_merge` uses two spaces after the newline when there are multiple files.

Spacetown:
@here-abarany Our test suite detected one problem. for clang compiler the decision test is failing:
```
                No function line found in gcov file:
                   main.gcda -> /gcovr/tests/decisions/output/decisions/switch_test.h##5b00ab17e6f7c638d27e0ef4ae668e29.gcov
                This may indicate that the file was generated without the proper gcov options (especially --branch-probabilities)?
                See <https://gcovr.com/en/stable/faq.html#which-options-are-used-for-calling-gcov>.
```

With active trace we can see that all hits are 0 here:
```
(TRACE) Parsed line: _MetadataLine(key='Source', value='./switch_test.h')
(TRACE) Parsed line: _MetadataLine(key='Graph', value='./main.gcno')
(TRACE) Parsed line: _MetadataLine(key='Data', value='./main.gcda')
(TRACE) Parsed line: _MetadataLine(key='Runs', value='1')
(TRACE) Parsed line: _MetadataLine(key='Programs', value='1')
(TRACE) Parsed line: _SourceLine(hits=0, lineno=1, source_code='#pragma once', extra_info=NONCODE)
(TRACE) Parsed line: _SourceLine(hits=0, lineno=2, source_code='', extra_info=NONCODE)
(TRACE) Parsed line: _SourceLine(hits=0, lineno=3, source_code='#include <cstdint>', extra_info=NONCODE)
(TRACE) Parsed line: _SourceLine(hits=0, lineno=4, source_code='#include <string>', extra_info=NONCODE)
...
```

What do you think?
- Only check if we use external files. (preferred)
- Only warning instead of error.
- ...

here-abarany:
So does this mean that clang doesn't provide the necessary function entries? Does that mean that without those entries, the coverage results will be empty for the full project?

If there are some legitimate cases where the required function entries are missing for a given file, but still present for others, then I think an "all of the above" approach may be warranted, where:
1. If `gcov` is internally managed and guaranteed to have the required arguments, it's not useful to either have an error or warn the user as it is largely outside of their control.
2. If there are legitimate cases where the function entries are missing, then a warning can still inform the user of potential misuse while still preventing false positives from aborting the tool.

However, if some toolsets simply don't output the required function entries, it might be worth re-thinking that requirement and falling back to the simple line coverage without extra branch or function information.

Spacetown:
What about canceling this approach here and starting from scratch. If there are no function lines at all we can assume a dummy function (with a warning) to which all the lines are added to and do not add this special function to the list of functions? In that case all the lines are reported at the end.

Spacetown:
@here-abarany Only the function is assigned to a dumb name, no function element is created. But you remind me to update the HTML generation. The function list and the link to it is now only created if there is at least one function.

PR #1208, opened 2025-11-29
Detect `[`, `]`, `[]` and `[]{` as noncode

### LINKED ISSUE #1164: Issues upgrading to gcvor 8.4
Hello,
I recently noticed, that version 8.4 just came out and all my coverage pipelines are now failing (because they all default to the latest version).
As it seems, gcovr runs just fine but, as I always upload all *gcov files, my upload-artifacts step failed, because no *gcov file could be found.
So, my guess is, that the `--keep` option does not work anymore?
Have a look at these two runs, where the latter uses a pinned version 8.3 and the former the 8.4.
In 8.4 run there are no *gcov files (have a look at the "Find gcov-files" steop)
https://github.com/DNKpp/mimicpp/actions/runs/18229882737/job/51910413516
https://github.com/DNKpp/mimicpp/actions/runs/18230017853/job/51910885744?pr=137

Secondly, gcovr's runtime has increased by 3-4x! Before it was at ~2min.

Lastly, I wanted to strengthen my usage of gcovr and follow the general advise to call gcovr from the actual build-directory instead of just the repo root. This was quite challenging as (to my very surprise) I had to adjust my relative filter, even though I made sure that I properly adjusted `--root` to were it pointed before. Reading the docs it states it's the *current working directory* but also mentions root. So, I think that wording should be further clarified but to be honest, this is quite unexpected that gcovr prefers the working directory over the explicit root option, because this means I should use absolute paths just to make sure the correct pattern is applied.

I also tested out the `--include` option, for which the docs says
> Files are searched recursive from the --root directory

Looking at the log, this doesn't seem to be true. This is how I call gcovr:
```
cd build
gcovr --root .. \
    --verbose \
    --include "include/mimic++(_ext)?" \
    --filter "${GITHUB_WORKSPACE}/include/mimic++(_ext)?" \
    ...
```
This is what the logs says:
> (DEBUG) File system is case sensitive.
> (DEBUG) Filters for --root: (1)
> (DEBUG)  - re.compile('^/__w/mimicpp/mimicpp/')
> (DEBUG) Filters for --filter: (1)
> (DEBUG)  - AbsoluteFilter(/__w/mimicpp/mimicpp/include/mimic++(_ext)?)
> (DEBUG) Filters for --exclude: (0)
> (DEBUG) Filters for --include: (1)
> (DEBUG)  - RelativeFilter(include/mimic++(_ext)? root=/__w/mimicpp/mimicpp/build)

As you can see, the `root=/__w/mimicpp/mimicpp/build` does not point to the directory set to `--root`, even the documentation says it should.

Spacetown:
Ignore lines containing only square brackets or only the start of a lambda operator.

Fixed #1164

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -48,6 +48,7 @@ Bug fixes and small improvements:
 - Extend support for zipped reports. If last suffix is ``.xz`` then LZMA is used to compress the report. (:issue:`1201`)
 - Fix function exclusion in report generation. (:issue:`1206`)
 - Do not ignore lines without functions, e.g. from inlined code. (:issue:`1204`)
+- Extend patterns for detection of noncode lines with ``[``, ``]``, ``[]`` and ``[]{``. (:issue:`1208`)
 
 Documentation:
 
--- a/src/gcovr/exclusions/noncode.py
+++ b/src/gcovr/exclusions/noncode.py
@@ -29,6 +29,7 @@ from ..logging import LOGGER
 
 _C_STYLE_COMMENT_PATTERN = re.compile(r"/\*.*?\*/")
 _CPP_STYLE_COMMENT_PATTERN = re.compile(r"//.*?$")
+_WHITESPACE_PATTERN = re.compile(r"\s+")
 
 
 def remove_unreachable_branches(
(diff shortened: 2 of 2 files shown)


Spacetown:
According to https://github.com/gcovr/gcovr/issues/1164#issuecomment-3592456565 this will not be merged because it doesn't solve all cases.

PR #1211, opened 2025-11-30
Add test with ccache

Spacetown:
Add test with ccache with with make and by prefixing the compilation command.


dlegaultbbry (approved):
LGTM

Did you compare to an identical test without ccache to see if the coverage is reported the same with and without for the same code?

Issue #1190, opened 2025-11-06
lcov output errors


abrown-RV:
# Describe the bug
We are now combining multiple coverage artifacts into one report, so the flow has become: multiple lcov files -> lcov merge -> lcov genhtml. I want to stress none of this is blocking, just figured could pass along for improved quality in the lcov output from gcovr.

Getting warnings and errors when feeding the gcovr lcov output into actual lcov for further processing. This output here is after following lcov's advice to convert `inconsistant` errors into warnings, so the `inconsistant` lines are actual errors that stop processing. I've just got the warnings here for more examples.
```
lcov: WARNING: (format) invalid characters removed from testname in tracefile coverage/gcovr.info: 'GCOVR report'->'GCOVR_report'
        (use "lcov --ignore-errors format,format ..." to suppress this warning)
lcov: WARNING: (unsupported) Function begin/end line exclusions not supported with this version of GCC/gcov; require gcc/9 or newer: attempting to derive function end lines - see lcovrc man entry for 'derive_function_end_line'.
        (use "lcov --ignore-errors unsupported,unsupported ..." to suppress this warning)
lcov: WARNING: (inconsistent) "/home/abrown/Catena/sdks/cpp/common/include/StructInfo.h":238: line is not hit but at least one branch on line has been evaluated.
        To skip consistency checks, see the 'check_data_consistency' section in man lcovrc(5).
        (use "lcov --ignore-errors inconsistent,inconsistent ..." to suppress this warning)
lcov: WARNING: (inconsistent) "/home/abrown/Catena/sdks/cpp/common/include/ParamWithValue.h":114: line is hit but no branches on line have been evaluated.
lcov: WARNING: (inconsistent) "/home/abrown/Catena/sdks/cpp/common/include/ParamWithValue.h":84: line is hit but no branches on line have been evaluated.
lcov: WARNING: (inconsistent) "/home/abrown/Catena/sdks/cpp/connections/REST/src/ServiceImpl.cpp":175:  function catena::REST::ServiceImpl::is_port_in_use_() const found on line but no corresponding 'line' coverage data point.  Cannot derive function end line.  See lcovrc man entry for 'derive_function_end_line'.
[...]

PR #1212, opened 2025-12-02
Ensure coverage for function line in LCOV report

Spacetown:
Temporary merge the data for the LCOV report and ensure that there is always a line data for a function.

Related #1190


Spacetown:
@abrown-RV, @dlegaultbbry I think this can be merged because the open point seems to be related to the command call (https://github.com/gcovr/gcovr/issues/1190#issuecomment-3691700486).

dlegaultbbry (approved):

Issue #1169, opened 2025-10-15
Improve HTML formatting in nested mode


moorlily:
In the existing HTML report, filenames include the full path name and are cut off when nested.  The full path is required for the --html-detail option.  It would be nice to only show the file or subdirectory name in --html-nested mode with the current directory stated at the top.  This would greatly help navigating deeply nested code trees.

Spacetown:
Can you elaborate a little bit more? Maybe with a screenshot of current and expected (by modifying the file).

moorlily:
Unfortunately I cannot get a screenshot.  I did some more fussing about with my work and found that the issue lies with the difference between themes: `github.dark-green` and `green` .  The light green will spread the long filenames out to be readable.  The dark green sets up its columns and data differently, such that the filename column is the same size as the function or branch column

moorlily:
The styles between github.dark-green and the default style green are very different.  However since I cannot get a screenshot outside of my work, I will withdraw the request.

Spacetown:
@latk, @dlegaultbbry Im working on improving the HTML report and have my problems with the old theme. The GitHub theme is easier to change and I would like to remove the old theme and make the GitHub theme to the default one. What do you think about this?

Spacetown:
> But TBH I'm not sure why the “github” theme would be easier to modify than the “default” templates, they appear structurally similar to me?

The main page is a table in the default theme and a flex grid in the GitHub theme. If we want to implement the sorting capabilities in the overview or function lists we need to do it twice.
We can also change the default theme to use a grid layout in the overview. I'll give it a try.

> Careful though: removing the default theme will likely break everyone who's using the gcovr --html-template-dir option. We have explicitly reserved the freedom to change templates however we wish, but this might be worth calling out prominently in the next changelog.

You're right, if they only overwrite some files it can break. But this always possible if we change the content of a template.

Spacetown:
@moorlily I think I reproduced what you tried to explain.

Default theme:
<img width="1532" height="61" alt="Image" src="https://github.com/user-attachments/assets/106bbdad-178d-458d-b649-ae3f5227acc3" />

Github theme (full name is shown as hover):
<img width="1282" height="61" alt="Image" src="https://github.com/user-attachments/assets/e4078dcb-0ea9-43b3-a6e4-96f515cc9ab6" />

dlegaultbbry:
Then I'm fine if the path is cutoff in this case. It's on the end user to adjust things if the data is many levels deep with nothing in between.

PR #1216, opened 2025-12-10
Change overview and function tables of HTML report

Spacetown:
Replace the tables with a grid layout to have the same look and feel.
With this change we also need to implement java script
improvements like sorting only once.

Old default theme:
image

New default theme:
image

Function list in both themes:
image

Related #1169

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -258,6 +258,14 @@
                 "transy"
                 // cspell:enable
             ]
+        },
+        {
+            "languageId": "html",
+            "ignoreWords": [
+                // cspell:disable
+                "functionslist"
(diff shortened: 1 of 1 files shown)


doc/images/screenshot-html-default-blue-src.jpeg
dlegaultbbry: We lost the branch coverage showing red in most of the new images. Was that intentional?

doc/images/screenshot-html.jpeg
dlegaultbbry: Is the CSS of this new image table wrong in any way?

PR #1219, opened 2025-12-26
Add sorting to HTML report

### LINKED ISSUE #1156: Add list of functions table sorting based on call count.

My project has some files with many lines, which can reach the hundreds. I would like to see with a quick glance how many times a function was called, so I can focus optimization time on those functions that are hot code.

However, gcovr does not allow to order the table of "List of functions" depending on call count, so checking which functions are most called with a quick glance is not possible.

Add a way to reorder the list of functions table. Ideally from the html itself, alternately with a command argument based on call count.

An alternate solution could be exporting to csv and ordering the rows manually, but this is a bit unconvinient.

# Additional context

Not needed

Spacetown:
Add sorting feature to the grid tables in the HTML report by clicking on the column header.
The value of `--sort` is taken for the report generation to configure the initial sorting on startup. If the value is `uncovered-number` the number is taken to sort the coverage values, for all other cases the percent value is used. If there are lines with same value the first column (file or function name) is used to sort them always in ascending order.

For directory overview page:
image
And for function list:
image

Closes #1156

### DIFF
--- a/tests/template-function/reference/template-function/clang-11/coverage.html
+++ b/tests/template-function/reference/template-function/clang-11/coverage.html
@@ -996,28 +996,28 @@ gcovr.sortGridTable = function (rowHeaderColumn) {
         th.classList.remove('sorted-ascending', 'sorted-descending');
     });
     rowHeaderColumn.classList.add(isAscending ? 'sorted-descending' : 'sorted-ascending');
-}"
+}
 
 document.addEventListener("DOMContentLoaded", () => {
   gcovr.fileLoaded();
 });
     </script>
   </head>
-"
+
   <body class="js-disabled">
 
     <header>
       <h1>GCC Code Coverage Report</h1>
       <hr/>
 <div id="summary" class="summary">
-<div>"
+<div>
   <table class="legend">
     <tr>
       <th scope="row">Directory:</th>
       <td>.</td>
     </tr>
     <tr>
-      <th scope="row">Coverage:</th>"
+      <th scope="row">Coverage:</th>
       <td class="legend">
         <span class="coverage-low">low: &ge; 0%</span>
         <span class="coverage-medium">medium: &ge; 75.0%</span>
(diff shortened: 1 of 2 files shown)


Spacetown:
@latk With #139 the option `--html-encoding` was introduced. With CSS I add a `↑` or `↓` to the column header as sorting indication which isn't rendered correct if the report e.g. uses `cp1252`:
image

Do you have an idea how to fix this? Is the option `--html-encoding` still needed today or can we force `uff-8`? There is no context why it was introduced.

doc/examples/example_html.html
```diff
@@ -858,6 +868,55 @@
...
+          : b.children[columnIndex].textContent.trim().toLowerCase();
+
+        let comparison = 0;
+        if (!isNaN(parseFloat(cellA)) && !isNaN(parseFloat(cellB))) {
+            comparison = parseFloat(cellA) - parseFloat(cellB);
+        } else {
+            comparison = cellA.localeCompare(cellB);
+        }
+
+        if( comparison == 0) {
```
dlegaultbbry:
```suggestion
        if (comparison == 0) {
```

src/gcovr/formats/html/common/sorting.js
```diff
@@ -0,0 +1,48 @@
...
+          : b.children[columnIndex].textContent.trim().toLowerCase();
+
+        let comparison = 0;
+        if (!isNaN(parseFloat(cellA)) && !isNaN(parseFloat(cellB))) {
+            comparison = parseFloat(cellA) - parseFloat(cellB);
+        } else {
+            comparison = cellA.localeCompare(cellB);
+        }
+
+        if( comparison == 0) {
```
dlegaultbbry:
```suggestion
        if (comparison == 0) {
```

Spacetown:
> minor comments

I'll change this when merging main brand h with toggle of line background.

dlegaultbbry (approved):
LGTM. Will be a helpful addition on top of all the other HTML tweaks.

dlegaultbbry:
Looking forward to seeing a new official release with all the enhancements!

PR #1236, opened 2026-01-16
Improve default html theme

Spacetown:
Improve alignment and wrapping data columns.
- All percent are aligned on the `%`.
- All counters are right aligned if the elements aren't wrapped. If the line gets wrapped the text is left aligned.

image


dlegaultbbry:
Should the example images be re-generated with the changes?

dlegaultbbry (approved):
Looks clean to me and a definite improvement. Thanks for making it happen!

PR #1240, opened 2026-01-29
Add compare mode for JSON, text and HTML report.

### LINKED ISSUE #886: Include multiple coverage reports in html with different coloring
I'm trying to identify code areas which are executed by INPUT_A but not by INPUT_B. I can generate two distinct coverage reports and then create two different html reports, but its not easy to see then where the difference in coverage is.

Ideally I'd like to be able to point gcovr to multiple folders containing .gcda files generated by different inputs. In the html view I then can see different colored lines for each input/run. For example: Green - covered by both inputs, Red - not covered by any input, blue - only covered by INPUT_A, yellow - only covered by INPUT_B 

So far I create two distinct reports and compare them manually

**Additional context**
N/A

Thanks!

Regards,

Robert

Spacetown:
If option `--json-compare` is given exact two trace files must be given with `--json-add-tracefile`.

In the JSON report the following properties are added:
```json
                    "diff": "Changed",
                    "diff_details": {
                        "execution_count": "Changed",
                        "blocks_percent": "Changed",
                        "excluded": "Strictly equal"
                    }
``` 
Following results are possible:

- `Removed`: The data is only available in first report.
- `Added`: The data is only available in second report.
- `Changed`: The data is covered in one report but uncovered in the other report.
- `Approximately equal`: The data is covered in both reports but with different hit counts.
- `Strict equal`: The data has the exact same coverage in both reports.

In text report the not equal lines are listed:
```
                     GCC Code Coverage Report (Diff mode)
Directory: .
File                                                   State   Lines
subdir/A/C/D/File6.cpp                               Changed   1,3,6
subdir/A/C/file5.cpp                     Approximately equal   1,3,6
subdir/A/file1.cpp                                     Added
subdir/A/File2.cpp                       Approximately equal   1,3-5
subdir/A/file3.cpp                                   Removed
subdir/A/File4.cpp                       Approximately equal   1,3-4
subdir/A/file7.cpp                            Strictly equal
subdir/B/main.cpp                                    Changed   19
                                         Approximately equal   13-18,21
```

The test summary report:
```
Removed file coverage:
  subdir/A/file3.cpp
Added file coverage:
  subdir/A/file1.cpp
Changed file coverage:
  subdir/A/C/D/File6.cpp
  subdir/B/main.cpp
Approximately equal file coverage:
  subdir/A/C/file5.cpp
  subdir/A/File2.cpp
  subdir/A/File4.cpp
Strictly equal file coverage:
  subdir/A/file7.cpp

```

The HTML overview:
image

In the HTML report a border at the end of the line is added which can be toggled with the corresponding button:
image

Closes #886

### DIFF
--- a/.github/workflows/gcovr-ci-job.yml
+++ b/.github/workflows/gcovr-ci-job.yml
@@ -49,7 +49,7 @@ jobs:
       NOX_CONTAINER_ARGUMENTS: >-
         ${{ inputs.container && format(' --session docker_run_compiler({0}) --', inputs.gcc) || '' }}
       GENERATE_DOCUMENTATION: >-
-        ${{ inputs.container && contains(fromJSON('["gcc-5", "gcc-6", "gcc-8", "gcc-9", "clang-10"]'), inputs.gcc) }}
+        ${{ inputs.container && contains(fromJSON('["gcc-5", "gcc-6", "gcc-7", "gcc-8", "gcc-9", "clang-10"]'), inputs.gcc) }}
     timeout-minutes: 25
     steps:
       - uses: actions/checkout@v4
--- a/doc/source/output/json.rst
+++ b/doc/source/output/json.rst
@@ -588,8 +588,8 @@ With the :option:`--merge-mode-functions` you can change this:
 JSON Format compare
 -------------------
 
-You can merge coverage data from two runs by adding them with
-:option:`--json-add-tracefile` and :option:`--json-compare`.
+You can compare coverage data from two runs by using :option:`--json-compare`
+and adding the two files with :option:`--json-add-tracefile`.
 
 For each run, generate :ref:`JSON output <json_output>`:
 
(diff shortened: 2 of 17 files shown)


Spacetown:
The docs will be extended. I think I APPROXIMATELY_EQUAL should be changed to APPROX_EQUAL.

To check the output I use the option `--skip-clean` when running the test and inspect the files in the output directory.

Spacetown:
> I also noticed that in some test output the css and js files are not emitted like compare/strictly-equal/gcc-14. That would make it hard to be able to load these files in a browser to inspect the output.

I always check the output of gcc-5 or clang-10. If I need the full output of gcc-14 I run the test with the option `--skip-clean` and open the files in the output directory.

Spacetown:
> As a future enhancement to reduce review diffs and avoid needing the output to be generated, perhaps just dump the CSS for each theme once and then just point to that single file for each of the test output files. Same with the JS part.
The CSS and the JS depend on the used CLI options. Both are templates. 
What I can do is to dump it per test and link it there.

Issue #1243, opened 2026-02-20
Unable to use gcovr with source based code coverage


xTachyon:
I tried to isolate what I'm seeing and I published the test [here](https://github.com/xTachyon/cov_test).
Run `./run.sh` and change `ENV version=..` to reproduce.

Using a somewhat not trivial project (the one on the repo), I'm getting the following errors:

## clang 18
```
#18 0.221 (INFO) Reading coverage data...
#18 8.576 (WARNING) /x/default.profraw: Found 'mcdc_records' in exported JSON report. This is ignored by GCOVR.
#18 10.12 (ERROR) Error occurred while reading reports:
#18 10.12 Traceback (most recent call last):
#18 10.12   File "/usr/local/lib/python3.12/dist-packages/gcovr/__main__.py", line 417, in main
#18 10.12     covdata = gcovr_formats.read_reports(options)
#18 10.12               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
#18 10.12   File "/usr/local/lib/python3.12/dist-packages/gcovr/formats/__init__.py", line 101, in read_reports
#18 10.12     covdata = LlvmHandler(options).read_report()
#18 10.12               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
#18 10.12   File "/usr/local/lib/python3.12/dist-packages/gcovr/formats/llvm/__init__.py", line 98, in read_report
#18 10.12     return read_report(self.options)
#18 10.12            ^^^^^^^^^^^^^^^^^^^^^^^^^
#18 10.12   File "/usr/local/lib/python3.12/dist-packages/gcovr/formats/llvm/read.py", line 105, in read_report
#18 10.12     read_json(
#18 10.12   File "/usr/local/lib/python3.12/dist-packages/gcovr/formats/llvm/read.py", line 269, in read_json
#18 10.12     read_json_files(data_source, data, options, merge_options, covdata, version)
#18 10.12   File "/usr/local/lib/python3.12/dist-packages/gcovr/formats/llvm/read.py", line 343, in read_json_files
#18 10.12     filecov.raise_data_error(
#18 10.12   File "/usr/local/lib/python3.12/dist-packages/gcovr/data_model/coverage.py", line 293, in raise_data_error
#18 10.12     raise GcovrDataAssertionError(
#18 10.12 gcovr.exceptions.GcovrDataAssertionError: /x/sqlite3.c Missing line coverage for line 189072 required by branch record.
[...]

PR #1245, opened 2026-02-23
Workaround missing linecov in profdata

Spacetown:
If LLVM profdata contains branch coverage data but the line coverage is
missing a line with 0 hits is added. This is the same behavior as
llvm-cov, see https://github.com/gcovr/gcovr/issues/1243#issuecomment-3941891536.


dlegaultbbry (approved):
LGTM

PR #1248, opened 2026-02-24
New html templates "boost"

sdarwin:
Add new template "boost" which supports dark/light theme in a single page.
image
image

Also a flat view is available:
image


Spacetown:
@sdarwin Thanks. I'll try to find some time this weekend.

Spacetown:
> At least superficially, I don't see where `count` was renamed to `execution_count` in any of the files of this pull request. How does it affect things? Currently the templates and scripts are working locally, without a problem, and the files are in sync, so they can be copied back and forth. In the future, if `count` is renamed to `execution_count` such that it breaks the templates, could there be some conditional logic added in the templates which is backwards compatible, so that can be a temporary workaround until the code is more stable, and continue to keep files the same in both repos... However I don't currently see where this is happening with `count`.

This I haven't done yet on your branch. You can check the template changes in last commit on main branch.

sdarwin:
Yes, maybe it will have to be:

```
pip install git+https://github.com/gcovr/gcovr.git@954a1b45f59cdb4ef83b7fb656093402c035bc66
```

Spacetown:
> > The file list is not scrollable in Safari.
> 
> Could you provide more details? Which file list? (there is the sidebar tree, and then other windows). How does non-scrolling exhibit itself? Could you host a copy at a certain URL, and I will visit that with safari. Or... I just published at https://develop.capy.cpp.al/gcovr/ . it seems to work. Which page is broken

If you open the provided link, click on src folder and reduce the height of the window you can't scroll to the entries at the bottom.
BTW: It's the same behaviour in Chrome.

> > The function list is generated but not not working or linked anywhere. Was this intended?
> 
> I don't know what this means. do you recommend a specific code change?

We always create a overall list of functions (e.g. tests/compare/reference/equal/clang-12/coverage.github.functions.html) This is linked on the main page and contains a sortable list of all functions with direct links to the definition. Also in the two existing themes each source file contains a hidden list of functions the top which can be used to jump directly to the definition in that file.

Spacetown:
Commit `8dc1762283bbc30044f12e998b78e7f762e0f849` on main branch introduces the Jinja2 rendering.

swaldhoer:
Ok, I see, then I'll guess my comment changes to: Add a SPDX identifier to all files in the repo anyway, so that SBOM creation can be simplified ;)

Spacetown:
@sdarwin I added the structure at the end of the `js` file. The normalisation is removed because the structure contains already the normalised data.

I've my doubts with `collapseSingleChildDirs`. Here you see how it looks for a very deep structure with one file:
image
In the tree on the left the structure is not visible and at the top it's too long.

If I select the file the report is:
image
Here the structure at the top doesn't show the folders and I can't go back to the parent folder.

sdarwin:
>  a very deep structure with one file

checking.

sdarwin:
> I've my doubts with collapseSingleChildDirs

Fixed.

src/gcovr/formats/html/boost/gcovr.js
```diff
@@ -398,28 +414,13 @@
...
-      }
-      collapseSingleChildDirs(node.children);
-    }
+    // Tree data is produced already-normalized and already-sorted by the
+    // upstream tooling (Python's gcovr_build_tree.py or gcovr itself), so
+    // no normalize/sort pass is needed here. We still dedupe + join chains.
+      deduplicateTree(window.GCOVR_TREE_DATA);
+      joinSingleChildDirs(window.GCOVR_TREE_DATA);
+      sortTree(window.GCOVR_TREE_DATA);
+      renderTree(treeContainer, window.GCOVR_TREE_DATA);
```
Spacetown: Is this needed in Javascript or is this better moved to the generation of the tree?

src/gcovr/formats/html/boost/gcovr.js
```diff
@@ -2097,4 +2158,4 @@
 
 })();
 
-window.GCOVR_TREE_DATA = {{ GCOVR_TREE_DATA | tojson(2) | safe }};
+window.GCOVR_TREE_DATA = {{ GCOVR_TREE_DATA | default([]) | tojson(2) | safe }};
```
Spacetown, line 2074: Do we need the `default` here or is this only needed for your template processing?

Spacetown (commented):
Only 2 comments. Maybe we should finish this first step and optimise in a follow up PR.

Edit: A deep structure move the buttons out of the view port but this is also valid for the existing themes and can be fixed in a separate PR.

src/gcovr/formats/html/boost/gcovr.js
sdarwin: From Claude:
Re: location of deduplicateTree / joinSingleChildDirs / sortTree / renderTree
Our preference is to keep all four in JS. Quick rationale for each:

renderTree — must be JS; manipulates the DOM. Not a question.

joinSingleChildDirs — this is a presentation transform, not a data transform. A "joined" entry like boost/url is a UI artifact (one compact sidebar row, one click to drill down to the deepest dir). The same tree data is also consumed by the breadcrumb, the file search, the "Flat" view toggle, and initNavOverride — none of which want the joined shape. Keeping tree.json as the canonical, unjoined structure preserves that flexibility. Moving the join into the producer would either bake one view into the data file or require an extra "view" parameter in the producer, which feels like the wrong layer.

sortTree — small (~15 lines), idempotent, and defensive. With our recent changes the Python builder already sorts each level, so this is essentially a no-op for our pipeline. We left it in JS as a safety net for any future producer (or hand-edited tree.json) that doesn't sort. Cheap insurance.

deduplicateTree — also small and idempotent. It handles the case where a directory page lists an entry whose path contains the parent dir name, producing a parent > parent > ... shape. We can't easily prove this never happens across all gcovr versions/flags/projects (a quick scan of one sample input shows no triggering paths, but that's anecdotal). At ~20 lines it's cheap enough to keep as a guard. If you have evidence it's actually unreachable in current gcovr output, we're happy to drop it.

Overall principle we're trying to hold to: Python (the producer) owns data shape — single-segment names, sorted, sibling-merged via normalize_tree. JS (the consumer) owns view transforms — dedupe-leftovers, join-for-compactness, render. The current split keeps the JSON contract simple and one-view-agnostic.

src/gcovr/formats/html/boost/gcovr.js
sdarwin: what if GCOVR_TREE_DATA is undefined? it doesn't hurt anything.  In our pipeline there may be moments when this variable is not set.

Spacetown: Whith this PR the variable is always defined. If the theme is `boost` it contains the tree, else a empty list.

sdarwin:
> move the buttons out of the view port 

I don't know what this means.  However the sound of "moving buttons" ... ?  Our designer thoughtfully planned out the appearance of the pages, and buttons.

src/gcovr/formats/html/boost/gcovr.js
Spacetown: Thanks for the explanation. I think the deduplicateTree must be removed. If the name is the same you can't remove one level. If gcovr reports a directory with the same name as the parent one there are two directories.
I know projects in my company where this is a used layout, don't ask.

Spacetown:
> > move the buttons out of the view port
> 
> I don't know what this means. However the sound of "moving buttons" ... ? Our designer thoughtfully planned out the appearance of the pages, and buttons.

If the window width is too small it looks like this:
image

sdarwin:
That is odd. In the screenshot you posted, the nested/flat buttons are separate from the long directory.
What I am getting, they are connected or joined.  See attached screenshot.
Either way...  if there is a problem when displaying very long directories, then, the problem is displaying very long directories, and it's not absolutely necessarily that buttons need to be re-arranged.   Although, still possible. 

Perhaps open an issue, with the screenshots, explanation, and we can investigate further.    And maybe in a separate pr.

Screenshot from 2026-05-19 23-04-39

sdarwin:
- deduplicateTree() removed
- default([]) removed

sdarwin:
returning default([]) 

In our pipeline, these events happen:
1. first run of gcovr to generate raw coverage.json
2. process coverage.json, produce coverage-fixed.json
3. second run of gcovr. input is coverage-fixed.json. Templates from a custom folder.
4. inject tree data into {output_dir}/index.js"  (i.e. gcovr.js)

Step 3 crashes without default([]). 
This seems small and safe: "if GCOVR_TREE_DATA is undefined, default to []".

Spacetown:
> returning default([])
> 
> In our pipeline, these events happen:
> 
> 1. first run of gcovr to generate raw coverage.json
> 2. process coverage.json, produce coverage-fixed.json
> 3. second run of gcovr. input is coverage-fixed.json. Templates from a custom folder.
> 4. inject tree data into {output_dir}/index.js"  (i.e. gcovr.js)
> 
> Step 3 crashes without default([]). This seems small and safe: "if GCOVR_TREE_DATA is undefined, default to []".

As mentioned, this is because of missing value in main branch but not in this PR. After merge it is always set. For your pipeline it's needed until you use the merge commit of gcovr in your environment.

sdarwin:
I may not understand the mechanism of how it is set.

We are currently using custom templates in a separate directory, and point to those templates. Even though our custom templates are identical to this PR... they are not the "boost" theme, they are just a set of custom templates.
1. If we continued with that arrangement, will GCOVR_TREE_DATA always be set?
2. If another company enjoys the boost theme, but wants to customize and modify it and develop it, they could copy the files to their own custom directory, and point to them using the "--html-template-dir" flag.  Again, it won't be the "boost" theme, it will just be "--html-template-dir".  Will GCOVR_TREE_DATA always be set?

Maybe the answer is just "yes", that GCOVR_TREE_DATA will be like an "environment variable" which is exposed to all templates, all the time, all gcovr templates of any type, whether they are custom templates or standard, it doesn't matter, everyone gets a copy of GCOVR_TREE_DATA, is that correct?

In the meantime until the PR is merged, we are still actively using the main branch of gcovr/gcovr,  so during this overlap period it's really helpful to keep the default(), it's pretty safe and harmless.  Then, after this is merged, and we can verify your assertion in production, and go back again and remove default().  But even then, is it harming anything? it makes the code robust to crashes.

Spacetown:
> I may not understand the mechanism of how it is set.
> 
> We are currently using custom templates in a separate directory, and point to those templates. Even though our custom templates are identical to this PR... they are not the "boost" theme, they are just a set of custom templates.
> 
> 1. If we continued with that arrangement, will GCOVR_TREE_DATA always be set?
> 2. If another company enjoys the boost theme, but wants to customize and modify it and develop it, they could copy the files to their own custom directory, and point to them using the "--html-template-dir" flag.  Again, it won't be the "boost" theme, it will just be "--html-template-dir".  Will GCOVR_TREE_DATA always be set?

With the last commit on your branch the answer is yes. Previously it was only together with boost theme.

> In the meantime until the PR is merged, we are still actively using the main branch of gcovr/gcovr, so during this overlap period it's really helpful to keep the default(), it's pretty safe and harmless. Then, after this is merged, and we can verify your assertion in production, and go back again and remove default(). 

Ok.

> But even then, is it harming anything? it makes the code robust to crashes.

With this I've my doubts because I want to crash if the required data is missing because it points out that there was a problem in gcovr logic.

sdarwin:
Latest modification:  if the browser window is made thinner and narrower when viewing a file with a long directory name, the buttons in the upper right corner will somewhat take precedence over the breadcrumbs, meaning the buttons will continue to be visible. As the screen narrows, it flips to flat mode, and the buttons convert into a gear icon which is smaller.  All of this seems to be working, and hopefully hasn't broken anything.  If you still have a concern about long directory names, let me know, or maybe its browser specific, I am testing on chrome,firefox,ubuntu. It seems alright.

Spacetown:
@sdarwin Thanks for implementing the theme.

sdarwin:
I will test this next week hopefully.

You removed function joinSingleChildDirs(nodes) ?

Did you understand what that was doing? Has it been implemented another way? This is more in the presentation layer than the data layer, and so it seems to belong in the template.

---

Edit: I should withhold my comments until I test it.  :-)    Maybe its ok.

Spacetown:
From the function comment and the code I thought it's not needed because this is already done by gcovr.
The only thing I missed was the sorting.

sdarwin:
@Spacetown we were collaborating on this pull request, right?  We should have a conversation about changes.  Ask for reviews.  At the last moment, you removed joinSingleChildDirs, and merged it, but did not ask me to take a look.   Notice these differences.

Earlier boost templates:

Screenshot from 2026-06-15 08-48-12

 
 
Screenshot from 2026-06-15 08-41-58

 
Latest:

Screenshot from 2026-06-15 08-34-43

Codecov (an inspiration for parts of the design):

Screenshot from 2026-06-15 08-35-09

Are you seeing these same results on your end?  

What I mean, codecov would not show a particular file with a name such as  "boost/corosio/io_context.hpp".  What it does is collapse the directories.     After directories have been collapsed, and you are browsing a directory, the file names are ordinary file names like "io_context.hpp".

Spacetown:
@sdarwin I fixed one issue when removing directories with one child and added by mistake this issue. I'm working on this but this part should be done in the gcovr tree structure generation and not in JS because sorting should always be directory first and then alphabetical.

Spacetown:
@sdarwin I've solved the issue with the sorting of the directories and found another issue in the template. If a directory tree is collapsed the last part is visible in the overview instead of the first part. For me this looks now as expected:
image

- The directories are sorted first.
- For the collapsed directory `subdir_0` is shown instead of `subdir_9`.
- If there is a common root directory it's visible in the tree.

If you agree, I'll raise a PR which you can test.

sdarwin:
Just from your last message, I do not precisely see what the issue is.

Here is a copy of the boost templates, prior to the recent modification of replacing joinSingleChildDirs.

https://master.corosio.cpp.al/gcovr/

It mostly looks good. Does this have the problem you are referring to? What is a URL there which exhibits the issue?

Here is a copy of the boost templates, after the recent modification of replacing joinSingleChildDirs.

https://develop.corosio.cpp.al/gcovr/

It's jumbled, we need to fix this one, and then see how it looks.  

 I would like to see the issue replicated. could you fix the collapse_directories_with_single_child issue, and then we can test and see what it's doing?

sdarwin:
Perhaps I am seeing the issue also.   Is it true, that it's working fine with very small directory structures, but something weird happens on deep directories?  Here was a test environment from a month ago:
https://2.url.prtest2.cppalliance.org/gcovr/
Drill down into the first level of include/boost/url.    There are subdir_a and subdir1, however they only appear in the sidebar tree, and not in the main window.

Spacetown:
> Just from your last message, I do not precisely see what the issue is.
>
> Here is a copy of the boost templates, prior to the recent modification of replacing joinSingleChildDirs.
> 
> https://master.corosio.cpp.al/gcovr/
>
> It mostly looks good. Does this have the problem you are referring to? What is a URL there which exhibits the issue?

I always use the test data in the repository and no other tests. In your link above the other error is visible. In the file list the folder `details` should be shown instead of the file `except.cpp`.

image

> Here is a copy of the boost templates, after the recent modification of replacing joinSingleChildDirs.
> 
> https://develop.corosio.cpp.al/gcovr/
> 
> It's jumbled, we need to fix this one, and then see how it looks.

As I said I'm working on this and it's looks very well.

> I would like to see the issue replicated. could you fix the collapse_directories_with_single_child issue, and then we can test and see what it's doing?

This is what I've done in the test data from where the screenshot was taken.

> Perhaps I am seeing the issue also. Is it true, that it's working fine with very small directory structures, but something weird happens on deep directories? Here was a test environment from a month ago: https://2.url.prtest2.cppalliance.org/gcovr/ Drill down into the first level of include/boost/url. There are subdir_a and subdir1, however they only appear in the sidebar tree, and not in the main window.

In the main window we have three times `param.hpp` but the second and third are sorted in by their directory name `subdir...`. Here the first directory should be visible instead of the file name.

Spacetown:
> > Just from your last message, I do not precisely see what the issue is.
> > Here is a copy of the boost templates, prior to the recent modification of replacing joinSingleChildDirs.
> > https://master.corosio.cpp.al/gcovr/
> > It mostly looks good. Does this have the problem you are referring to? What is a URL there which exhibits the issue?
> 
> I always use the test data in the repository and no other tests. In your link above the other error is visible. In the file list the folder `details` should be shown instead of the file `except.cpp`.
> 
> image
> > Here is a copy of the boost templates, after the recent modification of replacing joinSingleChildDirs.
> > https://develop.corosio.cpp.al/gcovr/
> > It's jumbled, we need to fix this one, and then see how it looks.
> 
> As I said I'm working on this and it's looks very well.
> 
> > I would like to see the issue replicated. could you fix the collapse_directories_with_single_child issue, and then we can test and see what it's doing?
> 
> This is what I've done in the test data from where the screenshot was taken.
> 
> > Perhaps I am seeing the issue also. Is it true, that it's working fine with very small directory structures, but something weird happens on deep directories? Here was a test environment from a month ago: https://2.url.prtest2.cppalliance.org/gcovr/ Drill down into the first level of include/boost/url. There are subdir_a and subdir1, however they only appear in the sidebar tree, and not in the main window.
> 
> In the main window we have three times `param.hpp` but the second and third are sorted in by their directory name `subdir...`. Here the first directory should be visible instead of the file name.

PR #1270, opened 2026-06-14
Fix runtime issue with many functions in a single file

### LINKED ISSUE #1268: Significant performance regression with version 8.6
# Description

There is a significant performance regression in the time to generate the HTML report between version 8.3 and 8.6. For a heavily templated header file (this one - https://github.com/trilinos/Trilinos/blob/master/packages/stk/stk_topology/stk_topology/topology_detail/meta_functions.hpp) the time went from under a minute to process that file to 4-5 hours (just for that one file, the full report is even more painful).

I have run this through a number of profiling tests and narrowed it down to almost all the time being spent in this block of code in `formats/html/write.py`

```python
for functioncov in filecov.functioncov(key=lambda functioncov: functioncov.key):
	for lineno in functioncov.linenos:
		f_data = dict[str, Any]()
		f_data["name"] = functioncov.name
		f_data["filename"] = filecov.properties["filtered_name"]
		f_data["html_filename"] = os.path.basename(filecov.properties["sourcefile"])
		f_data["line"] = lineno
		f_data["execution_count"] = functioncov.execution_count[lineno]
		f_data["blocks_percent"] = functioncov.blocks_percent[lineno]
		f_data["excluded"] = functioncov.excluded[lineno]
		function_stats = filecov.filter_for_function(functioncov).stats
		f_data["line_coverage"] = function_stats.line.percent_or(100.0)
		f_data["branch_coverage"] = function_stats.branch.percent_or("-")
		f_data["condition_coverage"] = function_stats.condition.percent_or("-")

		file_data["function_list"].append(f_data)
		functions[
			(
				functioncov.key,
				str(f_data["filename"]),
				int(f_data["line"]),
			)
		] = f_data
```

specifically, almost all the cost is in this line

```python
function_stats = filecov.filter_for_function(functioncov).stats
```

which appears to create a `Nfuncs X Nlines` cost in that inner loop. My problem case has about 180k functions and 600k lines because of all the template expansions (only about 41 actual unique code lines). Using the `--merge-lines` option does not help here, since the full line list is still being operated on at that point. 

If I manually pre-process the JSON coverage file with a python script to merge the duplicated lines first, this goes back to taking less than a minute and produces the expected coverage numbers.

Some ideas for possible solutions:

1. Apply the line-merging earlier when using `--merge-lines` so they arrive merged at this point already.
2. Find a way to make `filecov.filter_for_function` more efficient (maybe by pre-applying the filters)
3. Add an option to skip including the "function list" part of the report/HTML page and skip this loop entirely

I can share a reproducer JSON coverage file for this case if needed/helpful too.

# Files to reproduce

[html_perf.tar.gz](https://github.com/user-attachments/files/28311679/html_perf.tar.gz)

Spacetown:
With #1194 a dictionary with the lines of a function was introduced. This dictionary wasn't used in filter_for_function.

Fixes #1268

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,8 @@ New features and notable changes:
 
 Bug fixes and small improvements:
 
+- Fix runtime problem introduced with 8.4. (:issue:`1270`)
+
 Documentation:
 
 - Add section about ``gcov`` options which should not be used to FAQ. (:issue:`1237`)
@@ -40,6 +42,10 @@ Internal changes:
 
 8.6 (13 January 2026)
 ---------------------
+
+Known bugs:
+
+- :ref:`fix_1270`
 
 Breaking changes:
 
@@ -65,6 +71,7 @@ Known bugs:
 
 - :ref:`fix_1229`
 - :ref:`fix_1231`
+- :ref:`fix_1270`
 
 Breaking changes:
 
@@ -169,6 +176,7 @@ Known bugs:
 - :ref:`fix_1197`
 - :ref:`fix_1206_1`
 - :ref:`fix_1206_2`
+- :ref:`fix_1270`
 
 Breaking changes:
 
--- a/doc/source/known_bugs.rst
+++ b/doc/source/known_bugs.rst
@@ -9,6 +9,25 @@ Known bugs
 This list contains bugs for version 6.0 and newer, always check the latest
 version of this file available `here <https://gcovr.com/en/latest/known_bugs.html>`_.
 
+.. _fix_1270:
+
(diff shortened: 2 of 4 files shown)


Spacetown:
@tvoskui Any chance to test this?

tvoskui:
> @tvoskui Any chance to test this?

Not yet, hopefully can get to it soon though

tvoskui:
@Spacetown I ran both 8.6 and your PR on the test case. The PR took about 30 seconds while 8.6 took 4+ hours. HTML output looks equivalent, so I think this fixes it.

PR #1271, opened 2026-06-16
Fix issues in boost theme

Spacetown:
- The subdirectories were sorted in between the files instead of the top.
- For collapsed sub directories always the last part of the path was visible in the list.

Fixes #1248


Spacetown:
@sdarwin Please can you check the branch with your data?

sdarwin:
There is progress. The recent issues seem to be improved.

Compare:
https://develop.corosio.cpp.al/gcovr/  New
https://master.corosio.cpp.al/gcovr/    A few weeks ago

What I am seeing now is an extra top-level container, which I think should not be there.

Screenshot from 2026-06-16 07-21-49

It appears, not just in the sidebar tree, but also in the breadcrumbs, when you drill down to files.

Spacetown:
This is only the case if there is a common root directory, see `tests/html/reference/nested-theme-boost-green/gcc-5/coverage_nested.html` in the test data.
If there is no common root directory there is no root element, see `tests/html/reference/nested-theme-boost-blue/gcc-5/coverage_nested.html`.

This is also the case in the existing themes, see the summary at the top:
image

If you use `corosio_lcov_develop` as root directory for gcovr the root element should be removed.

sdarwin:
This is a bizarre issue.

Compare to the default templates:

gcovr 8.3

https://958.url.prtest2.cppalliance.org/gcovr/index.html

Screenshot from 2026-06-16 15-17-23

Next,

gcovr 8.6

https://develop.corosio.cpp.al/gcovr/

Screenshot from 2026-06-16 15-18-08

Notice the differences, above.  

8.3 shows the directory as "./" and the 8.6 shows the directory as "../corosio_lcov_develop".   ( these are slightly different projects on our side.  "corosio" compared to "url". )

 Was this a known difference between 8.3 and 8.6?

I think a neutral non-descript "./" is preferable to displaying the name of the top level folder and writing "../corosio_lcov_develop".

However, either way, the presentation of the main page, at the top level, shows:

- include
- src

During months of working on the boost theme such an issue never appeared.  Even as recently as yesterday it did not happen. Right?   This is very new.  It has just occurred.

Even if the root folder truly is named "corosio_lcov_develop", the sidetree doesn't need to know that. Or to display it.

Again refer to codecov as a reference.  The top-level listing are 

- include...
- src...

We should follow the presentation method of codecov.

Source data is coming from a json file.

 https://develop.corosio.cpp.al/gcovr/coverage-fixed.json 
 
 The json file doesn't mention "corosio_lcov_develop" anywhere.   
 
 Even if the root folder on the server happens to be "../something", you could instead think of it as a very neutral "./"  , and then avoid showing it.  This is an artistic choice.

Consider how collapsing multi-level directories required some amount of work.  It was not automatic.   We wrote code that collapses empty directories.    Figuratively speaking, the same concept might apply here, which is to say, this might require a bit more code to achieve.    Although, as recently as yesterday, it was working fine, so I don't know why this is happening.

Spacetown:
> Was this a known difference between 8.3 and 8.6?

No. Please can you check also 8.4 and 8.5? I checked the code and the reference data and the report directory was always the same. 

> I think a neutral non-descript "./" is preferable to displaying the name of the top level folder and writing "../corosio_lcov_develop".
> 
> However, either way, the presentation of the main page, at the top level, shows:
> 
> * include
> * src
> 
> During months of working on the boost theme such an issue never appeared. Even as recently as yesterday it did not happen. Right? This is very new. It has just occurred.
> 
> Even if the root folder truly is named "corosio_lcov_develop", the sidetree doesn't need to know that. Or to display it.

If you create a report for `/a/b` and all files are in `/a/b/corosio_lcov_develop` the `corosio_lcov_develop` must be shown.

> Again refer to codecov as a reference. The top-level listing are
> 
> * include...
> * src...
> 
> We should follow the presentation method of codecov.

The presentation depends on the root directory.
 
> Source data is coming from a json file.
> 
> https://develop.corosio.cpp.al/gcovr/coverage-fixed.json
> 
> The json file doesn't mention "corosio_lcov_develop" anywhere.
> 
> Even if the root folder on the server happens to be "../something", you could instead think of it as a very neutral "./" , and then avoid showing it. This is an artistic choice.

Do you have the code and directory structure for generating the report to check where this folder is added?

> Consider how collapsing multi-level directories required some amount of work. It was not automatic. We wrote code that collapses empty directories. Figuratively speaking, the same concept might apply here, which is to say, this might require a bit more code to achieve. Although, as recently as yesterday, it was working fine, so I don't know why this is happening.

The collapsing should be done once when generating the structure and not every time the page is loaded in the browser. We already collapsed the directories and this logic was changed in the JS file.

sdarwin:
OK. What a relief to discover what was happening. One of the steps in our existing build jobs is using symlinks, which never caused difficulties before. Until now. The latest code changes in this pull request interact via "os.path.realpath" which resolves symlinks, and change the apparent file path.   This is an open question: if symlinks must be prohibited, or can be supported.  Anyway.   I will create another pull request with a slight change to show the fix.

Spacetown:
> OK. What a relief to discover what was happening. One of the steps in our existing build jobs is using symlinks, which never caused difficulties before. Until now. The latest code changes in this pull request interact via "os.path.realpath" which resolves symlinks, and change the apparent file path. This is an open question: if symlinks must be prohibited, or can be supported. Anyway. I will create another pull request with a slight change to show the fix.

Symlinks are explicit supported because filters are applied to the realpath after resolving all symlinks. Is this branch now working also in your environment as expected?

sdarwin:
The gcovr process involves many layers. It's impossible to make a general statement for all software at all times, either
1. "supporting symlinks" == resolving symlinks
or
2. "supporting symlinks" == leaving them unmodified

Each one might be true in that instance and then the reverse would cause an issue.  

> Is this branch now working also in your environment as expected?

Yes. I think so.   https://develop.corosio.cpp.al/gcovr/index.html

Issue #1275, opened 2026-06-24
Boost Theme does not show excluded metrics


dlegaultbbry:
# Describe the bug

If you compare the 2 screenshots in #1273 , one can see the boost theme omits the excluded metrics compared to the other themes

# To Reproduce

N/A

# Expected behavior

The excluded metrics show be shown in all themes to be consistent

# Screenshots

See #1273 

# Desktop (please complete the following information)

- OS: Ubuntu 22.04
- GCC version: 12.2.0
- GCOVR version: latest main https://github.com/gcovr/gcovr/commit/9a0e25ba4829f2ae70597380a99cd01c4faf6b4d

# Additional context

N/A

dlegaultbbry:
Changes are held in this branch and are layered on top of the changes in #1277: https://github.com/dlegaultbbry/gcovr/tree/boost/excluded

PR #1274, opened 2026-06-24
fix boost theme HTML output for standalone and details formats

dlegaultbbry:
- Output needs to show full path of each item in these views
  - Fixed the logic to display the file/folder icon properly
- Add propagation of the missing html options so they can be used
  in templates
- Add propagation of options to CSS template
- In standalone or details view the file navigation bar is always
  empty so hide it
  - Adjusted the sidebar spacing based on if it is present or not
    using CSS variables
  - Remove flat/nested toggle

Fixes https://github.com/gcovr/gcovr/issues/1273

PR #1277, opened 2026-06-24
boost theme: fix issues related to optional data

### LINKED ISSUE #1276: Boost Theme conditions column issue and missing summary
# Describe the bug

2 Issues with the boost theme with conditions:
- Conditions column wraps and it shouldn't. This is regardless of the width of the screen in my browser window.
- There's no global Conditions section displayed at the top like Lines, Functions, Branches

It's also missing the calls and decisions summary card which can be added in the same fix

# To Reproduce

N/A

# Expected behavior

- Conditions column should show in appropriate location
- Conditions summary should be shown at the top

# Screenshots

Image

- OS: Ubuntu 22.04
- GCC version: 12.2.0
- GCOVR version: latest main https://github.com/gcovr/gcovr/commit/9a0e25ba4829f2ae70597380a99cd01c4faf6b4d

# Additional context

N/A

dlegaultbbry:
- Change the HTML boost theme to add the following:
  - Conditions Summary Card
  - Decisions Summary Card
  - Calls Summary Card
  - Change the source page to add the same missing summaries
as the top level cards
  - Change the source page to add the toggle buttons with the
matching columns
  - Change the functions page to add the same missing summaries
as the top level cards
  - Adjust the CSS and other templates for the missing columns
in the various pages
- There's no decision 'excluded' field; fix template which
  wanted to access this field
- Use the SHOW_X checks everywhere to be consistent. If there's
  no data for that type of optional item, it will show no
  coverage but at least all summaries/columns will align
  as a set

Fixes #1276

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,7 +27,7 @@ Bug fixes and small improvements:
 
 - Fix runtime problem introduced with 8.4. (:issue:`1270`)
 - Fix boost HTML details and simple output. (:issue:`1274`)
-- Fix boost HTML optional data (conditions, decisions, calls) output support. (:issue:`1277`)
+- Fix boost HTML support for Conditions, Decisions and Calls optional stats. (:issue:`1277`)
 
 Documentation:
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,7 @@ Bug fixes and small improvements:
 
 - Fix runtime problem introduced with 8.4. (:issue:`1270`)
 - Fix boost HTML details and simple output. (:issue:`1274`)
+- Fix boost HTML optional data (conditions, decisions, calls) output support. (:issue:`1277`)
 
 Documentation:
 
--- a/src/gcovr/formats/html/boost/directory_page.content.html
+++ b/src/gcovr/formats/html/boost/directory_page.content.html
@@ -2,10 +2,7 @@
 {# Check if we have any function or branch data #}
 {% set has_functions = entries | selectattr('functions.coverage', 'ne', '-') | list | length > 0 %}
 {% set has_branches = entries | selectattr('branches.coverage', 'ne', '-') | list | length > 0 %}
-{% set has_conditions = SHOW_CONDITION_COVERAGE and (entries | selectattr('conditions.coverage', 'ne', '-') | list | length > 0) %}
-{% set has_decisions = SHOW_DECISION and (entries | selectattr('decisions.coverage', 'ne', '-') | list | length > 0) %}
-{% set has_calls = SHOW_CALLS and (entries | selectattr('calls.coverage', 'ne', '-') | list | length > 0) %}
-<div class="file-list-container{% if not has_functions %} no-functions{% endif %}{% if not has_branches %} no-branches{% endif %}{% if not has_conditions %} no-conditions{% endif %}{% if not has_decisions %} no-decisions{% endif %}{% if not has_calls %} no-calls{% endif %}">
+<div class="file-list-container{% if not has_functions %} no-functions{% endif %}{% if not has_branches %} no-branches{% endif %}{% if not SHOW_CONDITION_COVERAGE %} no-conditions{% endif %}{% if not SHOW_DECISION %} no-decisions{% endif %}{% if not SHOW_CALLS %} no-calls{% endif %}">
   <div class="file-list-header">
     <div class="col-name sortable{% if info.sort_by == 'filename' %} {{info.sorted}}{% endif %}" data-sort="filename">
       Name
@@ -26,17 +23,17 @@
       Branches
     </div>
     {% endif %}
-    {% if has_conditions %}
+    {% if SHOW_CONDITION_COVERAGE %}
     <div class="col-conditions sortable" data-sort="conditions">
       Conditions
     </div>
     {% endif %}
-    {% if has_decisions %}
+    {% if SHOW_DECISION %}
     <div class="col-decisions sortable" data-sort="decisions">
       Decisions
     </div>
     {% endif %}
-    {% if has_calls %}
+    {% if SHOW_CALLS %}
     <div class="col-calls sortable" data-sort="calls">
       Calls
     </div>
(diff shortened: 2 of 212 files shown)


dlegaultbbry:
More work is needed here, when you flip between flat/nested, the optional columns do not display the data for each file.

src/gcovr/formats/html/boost/directory_page.summary.html
```diff
@@ -70,7 +70,88 @@ <h3>Branches</h3>
         </div>
         <div class="stat-row">
           <span class="stat-label">Total</span>
-          <span class="stat-value">{{info.branches.total - info.branches.excluded}}</span>
+          <span class="stat-value">{{info.branches.total - info.branches.excluded if info.branches.excluded != '-' else 0}}</span>
```
Spacetown, line 73: I don't like a negation if a else is used.

```suggestion
          <span class="stat-value">{{info.branches.total - (0 if info.branches.excluded == '-' else info.branches.excluded)}}</span>
```

src/gcovr/formats/html/write.py
```diff
@@ -148,10 +148,10 @@ def __load_css_template(options: Options) -> Template:
         return templates(options).get_template("style.css")
 
     @staticmethod
-    def render(options: Options, root_info: "RootInfo") -> str:
+    def render(options: Options, **data) -> str:
```
Spacetown: 🤔 Why was this working. I think we should fail if a template requests an undefined value.

Spacetown (changes requested):
Only a minor issue.

Spacetown:
@dlegaultbbry I would finish this PR before continuing with #1274.

src/gcovr/formats/html/boost/directory_page.summary.html
dlegaultbbry: Sure. This negation will disappear anyway once I get to doing #1275. I just didn't want to pack all changes in one PR making it harder to review.

dlegaultbbry:
It's easier to finish the other one as I've added changes and found other issues that need to be fixed generally which will be useful in this change.

src/gcovr/formats/html/write.py
dlegaultbbry: Yes, you can set `Environment(...,undefined=StrictUndefined)` and it will fail. I did that and I found errors in the templates referencing unknown fields. We might want to fix those in a separate PR though.

CHANGELOG.rst
```diff
@@ -26,6 +26,8 @@ New features and notable changes:
 Bug fixes and small improvements:
 
 - Fix runtime problem introduced with 8.4. (:issue:`1270`)
+- Fix boost HTML details output. (:issue:`1274`)
```
```suggestion
```

dlegaultbbry:
I have many other changes to add to this PR. I found other issues in testing so I'd really like to get the other boost PR finished so I can rebase on those changes and include the remaining items I fixed + all the test reference updates.

dlegaultbbry:
Latest includes additional changes on top of the last part which also fixes optional columns on the functions listing page. Once we agree on the changes, I can run the full test suite to update the reference data.

Note: I'll be on vacation July 24th-Aug 3rd so you'll know if I do not reply in that period.

CHANGELOG.rst
```diff
@@ -27,6 +27,7 @@ Bug fixes and small improvements:
 
 - Fix runtime problem introduced with 8.4. (:issue:`1270`)
 - Fix boost HTML details and simple output. (:issue:`1274`)
+- Fix boost HTML optional data (conditions, decisions, calls) output support. (:issue:`1277`)
```
Spacetown: Can this be rephrased? I don't get it.

src/gcovr/formats/html/boost/directory_page.content.html
```diff
@@ -2,10 +2,7 @@
 {# Check if we have any function or branch data #}
 {% set has_functions = entries | selectattr('functions.coverage', 'ne', '-') | list | length > 0 %}
```
Spacetown, line 3: I'm not sure about this two checks if the data is present. The page layout is changing depending on the viewed file.

Spacetown (commented):
Two notes for discussion.

CHANGELOG.rst
dlegaultbbry: I'll figure out a better sentence.

src/gcovr/formats/html/boost/directory_page.content.html
dlegaultbbry: I'll trial the changes on a project where there are header files shown in the output and where they have no coverable content.

CHANGELOG.rst
dlegaultbbry: How about `Fix boost HTML support for Conditions, Decisions and Calls optional stats in all pages where they can be displayed.`

src/gcovr/formats/html/boost/directory_page.content.html
Spacetown: If all files in a folder do not have branches the columns are removed. I think that's something which is not expected.

CHANGELOG.rst
Spacetown: I think `Fix boost HTML support for Conditions, Decisions and Calls optional stats.` would be enough.

src/gcovr/formats/html/boost/directory_page.content.html
dlegaultbbry: Do we want to remove the conditional on showing the columns even if they are empty? I do not want to destroy the original intent of the theme you and @sdarwin put together.

Spacetown: On the top level there would be the items because we've branches and functions available but for some folders the items magical disappear if there is no data. The more I think about this I don't like this if it's not evaluated on root level and applied on whole report like it's done for `SHOW_CONDITION_COVERAGE`.
Let's change this in a separate PR applying it to all themes.

Spacetown:
@dlegaultbbry Can you update the reference data and switch the PR to ready?

Spacetown (approved):
LGFM

PR #1280, opened 2026-07-07
boost theme: add display of excluded metrics

### LINKED ISSUE #1275: Boost Theme does not show excluded metrics

dlegaultbbry:
- To keep things aligned with other themes, display
  the excluded metrics
- Decisions has no excluded metrics so display a '-'
  as default for this stats type
- Reduce padding for card stats lines so that we waste
  less space from adding the excluded field

Fixes #1275


sdarwin:
The idea of the boost theme was not "To keep things aligned with other themes" but to redesign them in an aesthetic way.  I am concerned about "add display of excluded metrics" which could often fill the screen with a thousand extra zeros everywhere. is this really needed? or could it be opt-in?

dlegaultbbry:
I'm not opposed to having an option to display the excluded metrics (off by default) to exclude them for people who do not care about that metric. 

I don't understand why this theme is more special than others with regard to the data it is showing. Every user of this tool needs something different out of it for their purpose.

sdarwin:
It would be easier to discuss if you post screenshots.  Here is an idea (which sounds alright at least conceptually in the abstract.) Conditionally display 'excluded' (if you really think it's informational and important), however not when 0.   So, if a file is "5 / 10 / 20" then show "5 / 10 / 20".  However, when 0, just show "5 / 20".  The 0 is completely omitted.   Not "5 / 0 /  20".  Likewise for icons on a page. If excluded==0, wrap that in a conditional. Simply don't show the button at all.   For us, our coverage, it would just be pages and pages of zeros, which causes things to look cluttered and busy.   With this method, it would be hidden.  Or, as suggested before, a cli option, still not impossible, but better to reduce cli flags when possible.

dlegaultbbry:
> It would be easier to discuss if you post screenshots. Here is an idea (which sounds alright at least conceptually in the abstract.) Conditionally display 'excluded' (if you really think it's informational and important), however not when 0. So, if a file is "5 / 10 / 20" then show "5 / 10 / 20". However, when 0, just show "5 / 20". The 0 is completely omitted. Not " "5 / 0 / 20". Likewise for icons on a page. If excluded==0, wrap that in a conditional. Simply don't show the button at all. For us, our coverage, it would just be pages and pages of zeros, which causes things to look cluttered and busy. With this method, it would be hidden. Or, as suggested before, a cli option, still not impossible, but better to reduce cli flags when possible.

Checkout the repo/branch and navigate to the tests folder and go look at the various boost theme reference data in `tests/html/reference/`. These represent the updated visuals from these changes and are included in the second commit.

I work in safety critical software and it is required hence why I have added this in the original templates and in the boost one which I plan to use because it is very nice.

I'll defer to @Spacetown on the go/nogo decision of adding a cli generation option to enable excluded metrics in the visuals which I already said I'm 100% ok with. This is similar to your 'no-branches' request.

Spacetown:
> I work in safety critical software and it is required hence why I have added this in the original templates and in the boost one which I plan to use because it is very nice.

I also no this requirement from my work.
 
> I'll defer to @Spacetown on the go/nogo decision of adding a cli generation option to enable excluded metrics in the visuals which I already said I'm 100% ok with. This is similar to your 'no-branches' request.

For me all HTML reports should show the same data in a different way.

What I can imagine is to turn the data of if nothing is excluded but this would be a next step.

dlegaultbbry:
I created https://github.com/gcovr/gcovr/issues/1281 so we can debate the merits of potential solutions and the needs of everyone.

sdarwin:
You wrote "(only one file needs to be > 0 and they would be displayed)", that would only happen if an entire column is devoted to 'excluded' however if the data is just included together as mere text "1/2/3" then it's easy to edit or modify the text in-place.

dlegaultbbry:
Let me know where we're at with regard to this change. I think it's ready to merge and then you can adjust things based on your excluded auto-detection.

Spacetown:
For me it looks good but I'm waiting for a response on https://github.com/gcovr/gcovr/issues/1281#issuecomment-5055866564. In my opinion we can merge this PR and #1287.

PR #1283, opened 2026-07-07
Add missing changes for #1282

Spacetown:
- Add `StrictUndefined` to all Jinja2 environments.
- Refactor HTML generation to have better reuse.
- Use always alphanumerical sort order for next/previous links. Do not
  use directories in links.


dlegaultbbry:
Markdown also wants the strict setting as it uses jinja templates.

PR #1287, opened 2026-07-24
Add excluded numbers only if there are any excludes

Spacetown:
- Align the way how the values are shown in the summary of the default and the GitHub theme.
- Hide the excluded if the value is `0`.


Spacetown:
@dlegaultbbry, @sdarwin, @Pesa Please can you check this PR?
- The summary in the default and GitHub theme are aligned and use spans.
- Change the background colour to the foreground colour in the default theme
- For all themes the coverage bar in the overview and the line coverage values are put to the same column.
- Excluded values are only shown if not 0.
- The total value is without the excluded.

dlegaultbbry:
Checked out your branch to look at the doc examples and test output to see what it looks like:
* If you look at tests/exclusion/reference/exclude-branch/gcc-14/coverage.html the Coverage low/medium/high text in top left does not fit in the red/brown/green round boxes (highlight the whole line to see it cleanly)
* I like the regrouping of the line coverage numbers beside the progress bar instead of a separate column
* I think it was clearer with '/' instead of 'of'. This is how stats are generally displayed everywhere for ratios.
* I'd just us 'excl' instead of 'excluded' as it makes the naming of excluded lines consistent everywhere with reduced number of characters
  * tests/exclusion/reference/exclude-line-branch/gcc-14/coverage.default.html 
  * tests/exclusion/reference/exclude-line-branch/gcc-14/coverage.github.html
  * tests/exclusion/reference/exclude-line-branch/gcc-14/coverage.boost.html
* The fact that the metrics are now long, they often wrap on separate lines within the cell which now leads to less data being displayed on the screen (more scrolling). Perhaps the line/cell padding could be reduced to make things slightly more compact to try and keep things on a single line (unless the screen is so tiny that everything wraps).
  * tests/exclusion/reference/exclude-line-branch/gcc-14/coverage.boost.html Branches and Conditions of foo.cpp are wrapped for me even though my screen is very wide (see other comment on large numbers below which is related)
* In the boost theme, why is the spacing of the 'functions' column different than all other ones? I think it's because of the table padding which could probably be removed for the right edge, that would release some unused space for the rest. See tests/exclusion/reference/exclude-line-branch/gcc-14/coverage.boost.html
* For all these themes, we likely should add examples with fake input data of a file with 5000+ lines to it to see how large numbers affects the output when you get something like 3780/4012 lines covered with a few hundred branches.
* The text for the github and default theme could do like the boost theme and adjust the font size dynamically

Some of the suggestions above are debatable and should be done separately to keep this PR size down.

Spacetown:
> * If you look at tests/exclusion/reference/exclude-branch/gcc-14/coverage.html the Coverage low/medium/high text in top left does not fit in the red/brown/green round boxes (highlight the whole line to see it cleanly)
> * I think it was clearer with '/' instead of 'of'. This is how stats are generally displayed everywhere for ratios.
> * I'd just us 'excl' instead of 'excluded' as it makes the naming of excluded lines consistent everywhere with reduced number of characters
>   
>   * tests/exclusion/reference/exclude-line-branch/gcc-14/coverage.default.html
>   * tests/exclusion/reference/exclude-line-branch/gcc-14/coverage.github.html
>   * tests/exclusion/reference/exclude-line-branch/gcc-14/coverage.boost.html

changed

> * The fact that the metrics are now long, they often wrap on separate lines within the cell which now leads to less data being displayed on the screen (more scrolling). Perhaps the line/cell padding could be reduced to make things slightly more compact to try and keep things on a single line (unless the screen is so tiny that everything wraps).
>   
>   * tests/exclusion/reference/exclude-line-branch/gcc-14/coverage.boost.html Branches and Conditions of foo.cpp are wrapped for me even though my screen is very wide (see other comment on large numbers below which is related)
> * In the boost theme, why is the spacing of the 'functions' column different than all other ones? I think it's because of the table padding which could probably be removed for the right edge, that would release some unused space for the rest. See tests/exclusion/reference/exclude-line-branch/gcc-14/coverage.boost.html

Changed but still the columns are wrapped.

> * For all these themes, we likely should add examples with fake input data of a file with 5000+ lines to it to see how large numbers affects the output when you get something like 3780/4012 lines covered with a few hundred branches.
> * The text for the github and default theme could do like the boost theme and adjust the font size dynamically

Should be done with separate PRs.

dlegaultbbry (approved):
LGTM - Re-checked the pages I commented on last time.

PR #1288, opened 2026-08-06
Keep Fortran module procedures when excluding internal functions

Eljees:
Fixes #1253.

gfortran names a procedure defined inside a module `__<module>_MOD_<name>` (GCC manual, "Naming conventions"). Because the symbol starts with `__`, `_function_can_be_excluded()` classifies it as compiler-internal, and since 8.4 `remove_function_coverage()` drops the function *together with all of its lines* — so the body of every module procedure disappears from the report.

@Spacetown asked in the issue how a Fortran file should be detected. This uses two signals together, so the exception cannot leak into other languages:

- the source file suffix (`.f`, `.for`, `.ftn`, `.fpp`, `.f90`, `.f95`, `.f03`, `.f08`, `.f18`, case-insensitive), and
- the symbol matching the gfortran module pattern `__<module>_MOD_<name>`.

Everything gcovr means to exclude is untouched: `_GLOBAL__sub_I_*` keeps its own branch, and any other `__`-prefixed symbol is still excluded — including a `_MOD_`-shaped name in a C++ file.

Tests (`tests/test_exclusions.py`, new file — it drives `remove_internal_functions()` on a `FileCoverage` directly, so it needs neither gfortran nor a CI matrix change):

- `test_keep_fortran_module_function` — in `module_help.F90`, `__module_help_MOD_help_convert` survives while `__compiler_internal` is still removed;
- `test_remove_fortran_shaped_function_from_cpp_file` — the identically shaped symbol in `module_help.cpp` is still removed, i.e. the exemption does not weaken filtering elsewhere.

`CHANGELOG.rst` and `AUTHORS.txt` updated as `CONTRIBUTING.rst` asks; commit is signed off.

AI-assisted (LLM used for drafting and for running the checks); the analysis and the runs are mine.


CHANGELOG.rst
```suggestion
- Fix runtime problem introduced with 8.4. (:issue:`1270`)
- Fix boost HTML details and simple output. (:issue:`1274`)
- Fix boost HTML support for Conditions, Decisions and Calls optional stats. (:issue:`1277`)
- Add excluded metrics to boost HTML template (:issue:`1280`)
- Fix alignment in summary header of HTML single page report. (:issue:`1284`)
- Fix link to lines in HTML single page report. (:issue:`1285`)
- Preserve coverage for procedures declared in Fortran modules. (:issue:`1288`)
```

Spacetown (changes requested):
LGFM. Only CHANGELOG needs to be fixed.

Eljees:
Applied your suggestion — thanks for writing out the exact block. Nothing else in the PR changed.

CHANGELOG.rst
Eljees: Applied the suggested ordering and PR-number reference in 1fe02e0. Thanks!

Spacetown (approved):

PR #1289, opened 2026-08-10
fix: keep Fortran module procedures in coverage

Eljees:


Eljees:
Closing this as an accidental duplicate of #1288. The signed and reviewed work continues there.

PR #1291, opened 2026-08-12
fix: keep Fortran module procedures in coverage

Eljees:


Spacetown:
@Eljees Why are you creating several PRs for same branch?

Eljees:
Sorry - that one is on me. #1291 was a second attempt at the same Fortran module fix that #1288 already covered, and it should not have been opened while #1288 was still in review. Nothing in it differed from what you merged.

Thanks for closing it, and for the CHANGELOG catch on #1288.

PR #1293, opened 2026-08-17
Feature/add fail under condition

### LINKED ISSUE #1292: Reassess usefulness of --fail-under-condition
We should reassess if `--fail-under-condition` is actually useful. 

I've provided an example below, where the condition and branch coverage actually differ. I know this is an academic case, but for automotive development we are required to test for this metric. So we want our CI to warn us if we ever fail to uphold this metric.

Our current solution is to have an extra script that parses the output of gcovr to the assert that the condition coverage is upheld... it would be way nicer to have it directly within the tool so we don't need to parse html outputs manually.

> The following function results in a 100% branch coverage but only 83.3% condition coverage so It's not guaranteed to be identical in all cases
> 
> Function under test:
> ```c
> #include <stdbool.h>
> #include <stdint.h>
> 
> #define MCDC_RET_IF   (0x0001)
> #define MCDC_RET_ELSE (0x0002)
> 
> uint32_t mcdc_check(bool a, bool b, bool c) {
>     uint32_t retval;
> 
>     if ((a || b) && c) {
>         retval = MCDC_RET_IF;
>     } else {
>         retval = MCDC_RET_ELSE;
>     }
> 
>     return retval;
> }
> ```
> 
> Test cases:
> ```c
> void test_mcdc_check_true(void) {
>     bool param_a = true;
>     bool param_b = true;
>     bool param_c = true;
> 
>     /* Call function under test. */
>     TEST_ASSERT_EQUAL(MCDC_RET_IF, mcdc_check(param_a, param_b, param_c));
> }
> 
> void test_mcdc_check_false(void) {
>     bool param_a = false;
>     bool param_b = false;
>     bool param_c = false;
> 
>     /* Call function under test. */
>     TEST_ASSERT_EQUAL(MCDC_RET_ELSE, mcdc_check(param_a, param_b, param_c));
> }
> 
> void test_mcdc_check_c_false(void) {
>     bool param_a = false;
>     bool param_b = true;
>     bool param_c = false;
> 
>     /* Call function under test. */
>     TEST_ASSERT_EQUAL(MCDC_RET_ELSE, mcdc_check(param_a, param_b, param_c));
> }
> ```
> 
> Image
> 
> So in general i think having an additional `--fail-under-condition` is useful 

 _Originally posted by @mtnmbr in [#1163](https://github.com/gcovr/gcovr/issues/1163#issuecomment-5178027283)_

mtnmbr:
Resolves https://github.com/gcovr/gcovr/issues/1292

Regarding the comment on https://github.com/gcovr/gcovr/issues/1292#issuecomment-5309694684 i could adapt the PR to use the same exit-code as decision coverage if this is an issue


src/gcovr/__main__.py
```diff
@@ -59,6 +57,7 @@
 EXIT_BRANCH_NOK = 4
 EXIT_DECISION_NOK = 8
 EXIT_FUNCTION_NOK = 16
+EXIT_CONDITION_NOK = 32
```
Spacetown: As written in #1292 I wouldn't use a separate exit-code for conditions and instead rename `EXIT_DECISION_NOK` to `EXIT_DECISION_OR_CONDITION_NOK` because they are doing the same thing. The first one is done by parsing the source code and the second one is done by gcc itself.

src/gcovr/__main__.py
Spacetown, line 1: Why changing the include order?

CHANGELOG.rst
```diff
@@ -26,6 +26,7 @@ New features and notable changes:
 - Add compare mode for JSON, text and HTML reports. (:issue:`1240`, :issue:`1266`)
 - Stub missing line coverage for branches in LLVM source based code coverage format. (:issue:`1245`)
 - Add support for TOML files with :option:`--config` (:issue:`1258`)
+- Add :option:`--fail-under-condition` which will error under a given minimum condition coverage. (:issue:`1291`)
```
```suggestion
- Add :option:`--fail-under-condition` which will error under a given minimum condition coverage. (:issue:`1293`)
```

Spacetown (changes requested):
The option was still under discussion in #1292.

src/gcovr/__main__.py
mtnmbr: I think this is from my global ruff formatting config sorting the includes alphabetically 

- first std imports with `import`
- then std imports with `from`
- then external imports

I think the ruff behavior stems from the original `isort` implementation.

I can revert that if you don't like the ordering or make it consistent by adding the `'I'` rufle to ruffs lint selection.

This would also fix some inconsistencies in the alphabetic ordering across the other files.

mtnmbr: I've added it as a separate commit (https://github.com/gcovr/gcovr/pull/1293/changes/931702ef49710f196fac69c492e3d6b432b47562) so you can judge better if you want this included or not.

Spacetown: Please undo this change. Such things must be done in a separate PR fixing all files.

src/gcovr/__main__.py
mtnmbr: Done!

tests/condition/test_condition.py
Spacetown: Would rename to `test_conditions` and the folder to `conditions`.

Spacetown (approved):

tests/conditions/test_conditions.py
Spacetown, line 1: Would rename to test_conditions and the folder to conditions.

tests/test_args.py
```diff
@@ -351,6 +351,7 @@ def test_html_single_page_without_html_details_or_html_nested(
         "--fail-under-branch",
         "--fail-under-decision",
         "--fail-under-function",
+        "--fail-under-condition",
```
Spacetown: I started to use the order line, branch, condition, decision, functions, calls.
```suggestion
       "--fail-under-condition",
       "--fail-under-decision",
       "--fail-under-function",
```

tests/test_args.py
```diff
@@ -369,6 +370,7 @@ def test_failed_under_threshold_nan(
...
```
Spacetown:
```suggestion
        "--fail-under-condition",
        "--fail-under-decision",
        "--fail-under-function",
```

tests/test_args.py
```diff
@@ -387,6 +389,7 @@ def test_failed_under_threshold_negative(
...
```
Spacetown:
```suggestion
        "--fail-under-condition",
        "--fail-under-decision",
        "--fail-under-function",
```

CHANGELOG.rst
```diff
@@ -26,6 +26,8 @@ New features and notable changes:
 - Add compare mode for JSON, text and HTML reports. (:issue:`1240`, :issue:`1266`)
 - Stub missing line coverage for branches in LLVM source based code coverage format. (:issue:`1245`)
 - Add support for TOML files with :option:`--config` (:issue:`1258`)
+- Add :option:`--fail-under-condition` which will error under a given minimum condition coverage,
+  it shares the exit status ``8`` with :option:`--fail-under-decision`. (:issue:`1293`)
```
```suggestion
- Add :option:`--fail-under-condition` as a synonym for :option:`--fail-under-decision`. (:issue:`1293`)
```

tests/conditions/test_conditions.py
```diff
@@ -0,0 +1,106 @@
...
+        "--fail-under-condition=100",
+        use_main=True,
+    )
+    assert process.returncode == 8
+    messages = caplog.record_tuples
+    assert len(messages) == 1
+    assert messages[0][1] == logging.ERROR
+    assert messages[0][2].startswith("Failed minimum condition coverage ")
+    caplog.clear()
+
```
Spacetown, line 99:
```suggestion

    process = gcovr_test_exec.gcovr(
        "--json-add-tracefile=coverage.json",
        "--decisions",
        "--fail-under-condition=100",
        use_main=True,
    )
    assert process.returncode == 8
    messages = caplog.record_tuples
    assert len(messages) == 2
    assert messages[0][1] == logging.ERROR
    assert messages[0][2].startswith("Failed minimum condition coverage ")
    assert messages[1][1] == logging.ERROR
    assert messages[1][2].startswith("Failed minimum decision coverage ")
    caplog.clear()
```

src/gcovr/configuration.py
```diff
@@ -565,7 +565,8 @@ class UseSortUncoveredPercentAction(GcovrDeprecatedConfigOptionAction):
             "Exit with a status of 2 "
             "if the total line coverage is less than MIN. "
             "Can be ORed with exit status of '--fail-under-branch', "
-            "'--fail-under-decision', and '--fail-under-function' option."
+            "'--fail-under-decision', '--fail-under-function', and "
+            "'--fail-under-condition' option."
```
Spacetown:
```suggestion
            "'--fail-under-condition-or-decision' and '--fail-under-function'."
```

src/gcovr/configuration.py
```diff
@@ -578,7 +579,8 @@ class UseSortUncoveredPercentAction(GcovrDeprecatedConfigOptionAction):
             "Exit with a status of 4 "
             "if the total branch coverage is less than MIN. "
             "Can be ORed with exit status of '--fail-under-line', "
-            "'--fail-under-decision', and '--fail-under-function' option."
+            "'--fail-under-decision', '--fail-under-function', and "
+            "'--fail-under-condition' option."
```
Spacetown:
```suggestion
            "'--fail-under-condition-or-decision' and '--fail-under-function'."
```

src/gcovr/configuration.py
Spacetown:
```suggestion
        ["--fail-under-condition", "--fail-under-decision"],
```

src/gcovr/configuration.py
```diff
@@ -590,6 +592,21 @@ class UseSortUncoveredPercentAction(GcovrDeprecatedConfigOptionAction):
         help=(
             "Exit with a status of 8 "
             "if the total decision coverage is less than MIN. "
+            "The status is shared with the '--fail-under-condition' option. "
```
Spacetown:
```suggestion
```

src/gcovr/configuration.py
```diff
@@ -590,6 +592,21 @@ class UseSortUncoveredPercentAction(GcovrDeprecatedConfigOptionAction):
...
+    ),
+    GcovrConfigOption(
+        "fail_under_condition",
+        ["--fail-under-condition"],
+        type=check_percentage,
+        metavar="MIN",
+        help=(
+            "Exit with a status of 8 "
+            "if the total condition coverage is less than MIN. "
+            "The status is shared with the '--fail-under-decision' option. "
```
Spacetown:
```suggestion
```

src/gcovr/__main__.py
```diff
@@ -430,6 +444,7 @@ def _setup_pattern(option: str, patterns: list[str]) -> tuple[re.Pattern[str], .
         covdata,
         options.fail_under_line,
         options.fail_under_branch,
+        options.fail_under_condition,
         options.fail_under_decision,
```
Spacetown:
```suggestion
        options.fail_under_condition_or_decision,
```

src/gcovr/__main__.py
```diff
@@ -70,6 +70,7 @@ def get_exit_code(
     covdata: CoverageContainer,
     threshold_line: float,
     threshold_branch: float,
+    threshold_condition: float,
     threshold_decision: float,
```
Spacetown:
```suggestion
    threshold_condition_or_decision: float,
```

src/gcovr/__main__.py
```diff
@@ -79,6 +80,7 @@ def get_exit_code(
     if (
         threshold_line > 0.0
         or threshold_branch > 0.0
+        or threshold_condition > 0.0
         or threshold_decision > 0.0
```
Spacetown:
```suggestion
        or threshold_condition_or_decision > 0.0
```

src/gcovr/__main__.py
Spacetown:
```suggestion
        condition_or_decision_nok = False
        if threshold_condition_or_decision > 0.0:
            # Allow data with no conditions.
            percent_condition = stats.condition.percent_or(100.0)
            if percent_condition < threshold_condition_or_decision:
                condition_nok = True
                LOGGER.error(
                    "Failed minimum condition coverage (got %s%%, minimum %s%%)",
                    percent_condition,
                    threshold_condition,
                )

            # Allow data with no decisions.
            percent_decision = stats.decision.percent_or(100.0)
            if percent_decision < threshold_condition_or_decision:
                condition_or_decision_nok = True
                LOGGER.error(
                    "Failed minimum decision coverage (got %s%%, minimum %s%%)",
                    percent_decision,
                    threshold_condition_or_decision,
                )
```

src/gcovr/configuration.py
Spacetown:
```suggestion
        "fail_under_condition_or_decision",
        ["--fail-under-condition-or-decision", "--fail-under-condition", "--fail-under-decision"],
```

CHANGELOG.rst
```suggestion
- Add :option:`--fail-under-condition-or-decision` and :option:`--fail-under-condition` as a synonym for
    :option:`--fail-under-decision`, change the key for the configuration file. (:issue:`1293`)
```

Spacetown (commented):
I suggest to also use the same threshold value. since conditions are the compiler internal data and decisions our own parsing result they should be interchangable.

Spacetown:
You can apply suggestions in batch instead of doing one commit per suggestion.

mtnmbr:
Sorry i didn't know that. I am currently squashing changes locally.

Spacetown:
You also need to add yourself to the authors.

tests/test_args.py
mtnmbr: should these use the old synonyms or just the new `--fail-under-condition-or-decision` instead?

mtnmbr:
One thing i noted. The option `--decisions` is sometimes referred to as `--decisions` and sometimes as `--decision` what should the goal be here? should i fix this within this PR?

mtnmbr:
So merging the flag into one now also requires `--fail-under-condition-and-decision` to force the `--decisions` flag even if only condition should be covered. Is this intended? i hope i implemented it correctly.

src/gcovr/__main__.py
```diff
@@ -395,8 +405,8 @@ def _setup_pattern(option: str, patterns: list[str]) -> tuple[re.Pattern[str], .
         LOGGER.error("%s", e)
         return EXIT_CMDLINE_ERROR
 
-    if options.fail_under_decision > 0.0 and not options.show_decision:
-        LOGGER.error("--fail-under-decision need also option --decision.")
+    if options.fail_under_condition_or_decision > 0.0 and not options.show_decision:
+        LOGGER.error("--fail-under-condition-or-decision need also option --decisions.")
         return EXIT_CMDLINE_ERROR
```
Spacetown: This block must be removed because it's not valid anymore.
```suggestion
```

Spacetown:
@mtnmbr Please do not rewrite history because I always need to check all changes.

mtnmbr:
> @mtnmbr Please do not rewrite history because I always need to check all changes.

Alright, sorry. I am used to gitlab which lets' you show the diff between changes after rebasing. Thus i tend to keep the commit history clean.

Spacetown:
> > @mtnmbr Please do not rewrite history because I always need to check all changes.
> 
> Alright, sorry. I am used to gitlab which lets' you show the diff between changes after rebasing. Thus i tend to keep the commit history clean.

I never saw this in GitLab. The commits are unrelated and therefor the system can't track the comments and changes.

mtnmbr:
> > > @mtnmbr Please do not rewrite history because I always need to check all changes.
> > 
> > 
> > Alright, sorry. I am used to gitlab which lets' you show the diff between changes after rebasing. Thus i tend to keep the commit history clean.
> 
> I never saw this in GitLab. The commits are unrelated and therefor the system can't track the comments and changes.

image
you do have the same option on github. it's just hidden behind the compare changes button. (similar to gitlab) so after a force push this basically is the new commit added. but i will just append commits for now and squash them in the end :)

Spacetown:
But that's different to:
image
Also the review comments get outdated and are not seen anymore in the file view because the commit, the comment belongs to, isn't part of the history anymore.

mtnmbr:
I don't understand the reason for the documentation fail, but besides that coverage now also seems to be okay (did you change any settings there? i was uncertain as to why it failed before)

EDIT:
ok seemed to be a flaky URL timeout.

Is there anything pending to be changed?

Spacetown:
It was a rate limit of the link check.

Spacetown (approved):
LGFM

mtnmbr:
Thank you :)

PR #1294, opened 2026-08-18
Add import sorting to ruff lints

mtnmbr:
Opening a separate PR to discuss the import sorting.

Regarding discussion in https://github.com/gcovr/gcovr/pull/1293#discussion_r3801689717 

Let me know if you find this valuable or not :)


Spacetown:
Would wait until #1287 is merged.

Spacetown:
@mtnmbr I don't get the CI error, locally (MacOS and in container) everything is fine.
After updating the pipeline to show the fix I get following:
```
--- a/tests/test_args.py
+++ b/tests/test_args.py
@@ -25,11 +25,11 @@ import sys
 from pathlib import Path
 
 import pytest
+from gcovr.version import __version__
 
 from gcovr.__main__ import main
 from gcovr.data_model.coverage import FileCoverage
 from gcovr.data_model.version import FORMAT_VERSION
-from gcovr.version import __version__
 from tests.conftest import GCOVR_ISOLATED_TEST
```

This change doesn't make sense for me. Any idea?

mtnmbr:
That's odd. so the `gcovr.version` import should be at the bottom. For me if i manually place the version just below pytest ruff also reverts it immediatly.

I've also tested different ruff versions. my current hatch setup has 0.13.x but also the latest 0.16.7  resulted in the same code. 

So in my windows/WSL setup here i did not get any other results as you did locally

Spacetown:
So the only option would be to turn it of in this file or do you have another idea?

EDIT:
I found the issue. the file _version.py_ is not part of the repository and not created yet therefore it is added to third-party section. I updated _pyproject.toml_ to consider `gcovr` always as first-party.

mtnmbr:
Perfect. i've just asked CLAUDE and it came up with the same solution. Seems to be a valid addition :)

PR #1297, opened 2026-09-09
Fix KeyError when merging tracefiles with mixed function positions

### LINKED ISSUE #1296: KeyError while writing the report when merging tracefiles where only some carry a function "pos"
# Describe the bug

Merging tracefiles where the same function is known on several lines and only some carry a `pos` aborts with a `KeyError` while writing the report. All `--merge-mode-functions` values are affected.

Tracefiles disagree about `pos` whenever they were captured with different gcov versions: gcovr only records positions when it reads gcov's JSON output, which it uses only if `gcov --version` prints `JSON format version: 2` (gcc 14+). With an older gcov the text parser is used and records none.

`FunctionCoverage.serialize` assumes every line in `execution_count` has a `start`/`end` entry, but neither merge branch keeps that true: separate mode extends `execution_count` with every line of the other object but only extends `start`/`end` `if other.start is not None`, and the single-line branch only re-keys them `if self.start is not None and other.start is not None`, leaving a stale key when one side has none.

# To Reproduce

1. `with_pos.json`:

```json
{"gcovr/format_version": "0.14",
 "files": [{"file": "foo.c",
            "functions": [{"name": "foo", "lineno": 10, "execution_count": 1,
                           "blocks_percent": 100.0, "pos": ["10:1", "13:1"]}],
            "lines": [{"line_number": 10, "count": 1, "branches": []}]}]}
```

2. `without_pos.json`: identical but `"lineno": 20`, `"line_number": 20`, no `pos`.

3. ```console
   $ gcovr -a with_pos.json -a without_pos.json --merge-mode-functions=separate --json merged.json
     File ".../gcovr/data_model/coverage.py", line 1978, in serialize
       ":".join([str(e) for e in self.start[lineno]]),
   KeyError: 20
   ```

Single-line modes fail too when the positioned side is not on the merged line: with `pos` on line 20 and none on line 10, `--merge-mode-functions=merge-use-line-min` gives `KeyError: 10`.

# Expected behavior

The merge succeeds and `pos` is written for the lines that have one. Giving both files a `pos` already works, which is what points at the asymmetry.

# Desktop

- OS: Ubuntu 24.04 (container), also reproduced on macOS
- GCC version: tracefiles from gcc 13.3.0 and gcc 14.3.0 merged in the same run
- GCOVR version: 8.6 (PyPI) and `main` at c543b771, both reproduce
- Project directory layout: not relevant, the reproducer is two hand-written tracefiles

# Additional context

Found in a CI setup that collects coverage on several platforms and merges the per-platform tracefiles: those built with a cross toolchain (gcc 14) carry positions on every function entry, those built with the host compiler (gcc 13) carry none. Merging them produces no report at all.

Treating `start`/`end` as sparse, like `blocks_percent` and `excluded` already are, fixes both paths. I have a branch ready if that looks right.

Nearby: `min(*self.start.values(), *other.start.values())` in the single-line branch degrades to `min(tuple)` and returns an `int` when only one value is present.

kartben:
gcovr records a function's position only when the gcov binary reports one, so tracefiles produced with gcov 14 or newer carry "pos" and older ones do not. Merging both raises a KeyError while writing the report as soon as the function is known on more than one line: the merged entry gains the line for the counters, but start and end keep only the lines they already had.

Serialize "pos" for the lines that have a position rather than assuming every line has one, and key the merged position by the line the counters were merged into so no stale entry is left behind when only one side of the merge carries positions.

Fixes #1296


src/gcovr/data_model/coverage.py
```diff
@@ -2368,16 +2373,26 @@ def merge(
...
-                    )
+                # or the minimum start, dropping the positions if neither side has one
+                starts = [
+                    *(() if self.start is None else self.start.values()),
+                    *(() if other.start is None else other.start.values()),
+                ]
+                self.start = (
+                    None
+                    if not starts
+                    else CoverageDict[int, tuple[int, int]]({lineno: min(starts)})
```
Spacetown: Do not use `if not ...` together with an `else`.

```suggestion
                    CoverageDict[int, tuple[int, int]]({lineno: min(starts)})
                    if starts
                    else None
```

src/gcovr/data_model/coverage.py
```diff
@@ -2368,16 +2373,26 @@ def merge(
...
+                )
+                # or the maximum end
+                ends = [
+                    *(() if self.end is None else self.end.values()),
+                    *(() if other.end is None else other.end.values()),
+                ]
+                self.end = (
+                    None
+                    if not ends
+                    else CoverageDict[int, tuple[int, int]]({lineno: max(ends)})
```
Spacetown:
```suggestion
                    CoverageDict[int, tuple[int, int]]({lineno: max(ends)})
                    if ends
                    else None
```

src/gcovr/data_model/coverage.py
Spacetown, line 2384:
```suggestion
                starts = []
                if self.start is not None:
                    starts += self.start.values()
                if other.start is not None:
                    starts += other.start.values()
                if starts:
                    self.start = CoverageDict[int, tuple[int, int]]({lineno: min(starts)})
```

src/gcovr/data_model/coverage.py
Spacetown, line 2394:
```suggestion
                ends = []
                if self.end is not None:
                    ends += self.end.values()
                if other.start is not None:
                    ends += self.end.values()
                if ends:
                    self.end = CoverageDict[int, tuple[int, int]]({lineno: max(ends)})
```

CHANGELOG.rst
```diff
@@ -37,6 +37,7 @@ Bug fixes and small improvements:
 - Fix link to lines in HTML single page report. (:issue:`1285`)
 - Preserve coverage for procedures declared in Fortran modules. (:issue:`1288`)
 - Fix missing branches and conditions for uncovered lines in HTML report. (:issue:`1290`)
+- Fix ``KeyError`` when merging tracefiles in which only some report a function position. (:issue:`1296`)
```
Spacetown, line 40:
```suggestion
- Fix ``KeyError`` when merging tracefiles in which only some report a function position. (:issue:`1297`)
```

Spacetown (commented):
You should also add your name to _AUTHORS.txt_ and add an entry to the _doc/source/known_bugs.rst_. The bug should be linked in the _CHANGELOG.rst_ from version `8.0` to `8.6`.

CHANGELOG.rst
kartben: mmh, why? This _is_ the issue? https://github.com/gcovr/gcovr/pull/1296

Spacetown: Because we always reference the PR. Not every PR has a related issue.

PR #1302, opened 2026-09-17
Add support for multiple text reports

### LINKED ISSUE #1286: Support multiple values for `--txt-metric` to display several coverage columns in text report

Currently, `--txt-metric` only accepts a single value (`line`, `branch`, or `decision`). When I want to see both line and branch coverage details in the text report, I have to run gcovr twice with different `--txt-metric` values and compare the outputs manually.

The `--print-summary` report does show all metrics at once, but only as percentages — it doesn't provide the per-file detail table with total/covered/missing columns that the text report offers.

Allow `--txt-metric` to accept multiple values, e.g.:
```gcovr --txt-metric line branch```

This would produce a text report with multiple column groups, one per metric. For example:

```
GCC Code Coverage Report
Directory: .
File                 Lines              Branches
                  Total  Exec  Cover   Total  Taken  Cover   Missing
src/foo.c            42    38    90%      12     8    66%   41-42,10,15
src/bar.c            30    30   100%       8     8   100%
TOTAL                72    68    94%      20    16    80%
```

The default behavior (`--txt-metric line`) would remain unchanged for backward compatibility.

- Running gcovr multiple times with different `--txt-metric` values and comparing outputs side by side — works but cumbersome.
- Using `--print-summary` to get all metrics — but it only shows percentages, not the per-file detail with missing lines/branches.
- Using a different output format (HTML, JSON) — these show all metrics but are not suitable for quick terminal inspection.

# Additional context

Related but different from #1281 which is about excluded metrics visibility. This request is about combining the existing metric types into a single text report.

### PR DESCRIPTION
**Pull request by Spacetown**
Extend option `--txt-metric` to support multiple reports in one run.

EDIT: After discussion the metrics are printed side by side and the column with the missing lines is removed in that case.

```
-------------------------------------------------------------------------------------------------
                                     GCC Code Coverage Report
Directory: .
-------------------------------------------------------------------------------------------------
File                                     Decisions    Taken  Cover   Conditions    Taken  Cover
-------------------------------------------------------------------------------------------------
main.cpp                                        69       35    50%           86       44    51%
switch_test.cpp                                  0        0    --%            0        0    --%
switch_test.h                                    3        1    33%            0        0    --%
-------------------------------------------------------------------------------------------------
TOTAL                                           72       36    50%           86       44    51%
-------------------------------------------------------------------------------------------------
```

Closes #1286

### CODE REVIEW

File: src/gcovr/formats/txt/__init__.py [hunk 1]
```diff
@@ -45,16 +45,19 @@ def get_options(cls) -> list[GcovrConfigOption | str]:
...
                 ["--txt-metric"],
-                config="txt-metric",
+                config="txt-metrics",
                 group="output_options",
-                help=("The metric type to report. Default is '{default!s}'."),
+                help=(
+                    "The metric type to report. If option is given multiple times the "
+                    "reports are printed in the given order. Default is 'line'."
+                ),
                 choices=("line", "branch", "decision"),
```
dlegaultbbry: What about conditions since decisions is there (option could simply be ignored if there's no condition data)

dlegaultbbry (commented):
Non blocking comment. Rest LGTM.

src/gcovr/formats/txt/__init__.py
Spacetown: Condition coverage is not available in text report yet. I'll add it also.

dlegaultbbry: Feel free to do this addon in a separate PR and go ahead and merge this one.

Spacetown: Already done. 😄

Spacetown:
@dlegaultbbry, @Hugo-Fabresse If we add an option `--txt-report-hide-missing-column` it would be possible to also generate a report where the different metrics can be printed side by side. The report would be more compact and even with only one metric better readable if there are many missing lines.
What's your opinion?

dlegaultbbry:
I think the idea has merrit for `--txt-metric-no-missing`.

Instead of adding this option, what about allowing to do `--txt-metric line,branch,condition` and if you do request more than one at a time, the missing stats are simply not published and you only get the detailed summary with each type in its own column similar to the top level HTML output. If you only request one stat type you get the 'Missing' part for it as current behaviour.

Spacetown:
That sounds even better. If someone needs the _Missing_ column he needs to generate a JSON and run the report generation for each metric separate. 

Since the `--txt-metric` has a choice you need to use the option several times. Supporting a comma separated list we need to implement the validation on our own. I didn't"t want to do this but I think it"s better to spend the time for this because it's more intuitive. What do you think?

dlegaultbbry:
I'm fine with multiple --txt-metric options to simplify the logic.

Spacetown:
Will continue tomorrow or on Sunday and switch the PR to ready.
Shall the different metrics be separated by a `|`?

dlegaultbbry:
I don't think it needs to, if everything is well aligned in columns with sufficient empty space between it should be readable.

This is based on if it looks somewhat like: https://github.com/gcovr/gcovr/issues/1286#issuecomment-5050608109

Spacetown:
The total column has 10 characters and the exec column only 9. This should be fairly enough.

